PR #214, opened 2026-04-22
Enforce SECRET_KEY in production config

### LINKED ISSUE #211: [security] Weak default SECRET_KEY used in base config
## Security Finding

**Severity:** MEDIUM
**Category:** Authentication (Session Security)
**Source:** Automated security audit (2026-04-17)

## Description

The base `Config` class in `config.py:42` sets a weak default SECRET_KEY:

```python
SECRET_KEY = os.getenv('SECRET_KEY', 'a_default_secret_key_for_development')
```

This default is used for:
- Flask session signing (session cookies)
- Fernet encryption key derivation for stored refresh tokens (`TokenService`)
- Session integrity for OAuth tokens stored server-side

If `SECRET_KEY` is not set in the environment, the predictable default would allow session forgery and decryption of stored refresh tokens. The `ProdConfig` class inherits this default without overriding it.

## Affected Files

- `config.py:42` -- weak default in base Config
- `config.py:93-122` -- ProdConfig inherits the weak default
- `shuffify/services/token_service.py:36-63` -- derives Fernet key from SECRET_KEY

## Remediation

1. In `ProdConfig`, override SECRET_KEY to require an env var with no fallback:
   ```python
   SECRET_KEY = os.environ['SECRET_KEY']  # Crash on missing -- intentional
   ```
2. Alternatively, add SECRET_KEY to `validate_required_env_vars()` for production
3. The development default is acceptable for local dev, but production must never fall through to it

## Risk Context

The `validate_required_env_vars()` function currently only checks for Spotify credentials, not SECRET_KEY. Production fails fast on missing Spotify creds but silently uses the weak SECRET_KEY default. The encrypted refresh tokens in the database would be trivially decryptable if the default key is in use.

chrisrogers37:
## Summary
- **ProdConfig** overrides `SECRET_KEY` with `os.environ.get("SECRET_KEY")` — no weak fallback
- **`validate_required_env_vars()`** now accepts `config_name` and requires `SECRET_KEY` when production
- App fails fast on startup if `SECRET_KEY` is missing in production; dev default preserved for local use

Fixes #211

## Test plan
- [x] `test_production_requires_secret_key` — missing SECRET_KEY raises ValueError in production
- [x] `test_production_passes_with_secret_key` — validation passes when SECRET_KEY is set
- [x] `test_development_allows_missing_secret_key` — dev mode doesn't require SECRET_KEY
- [x] `test_prod_config_no_weak_default` — ProdConfig never uses the hardcoded dev default
- [x] Full test suite: 1742 passed, 30 pre-existing failures (unrelated app context issues)
- [x] flake8 clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Callback route validates the returned state against the stored session state
   - Mismatched or missing state rejects the callback with a user-friendly error
   - State token consumed (removed from session) after validation to prevent replay
+- **Production SECRET_KEY Enforcement** - Production config no longer falls back to a weak default `SECRET_KEY`
+  - `ProdConfig` overrides `SECRET_KEY` with `os.environ.get('SECRET_KEY')` — no hardcoded fallback
+  - `validate_required_env_vars()` now requires `SECRET_KEY` when `config_name='production'`
+  - Application fails fast on startup if `SECRET_KEY` is not set in production environment
+  - Development default remains for local dev convenience
 
 ### Added
 - **Track-Level Locks** - Per-track position locks to protect individual tracks from automated operations
--- a/config.py
+++ b/config.py
@@ -6,7 +6,7 @@ from dotenv import load_dotenv
 # environment.  Without this guard, load_dotenv() injects the production
 # DATABASE_URL from .env at module-import time, before any fixture can
 # override it, causing tests to hang on Neon PostgreSQL connections.
-if 'pytest' not in sys.modules:
+if "pytest" not in sys.modules:
     load_dotenv()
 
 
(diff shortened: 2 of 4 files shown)


chrisrogers37 (commented):
## Code Review

**Verdict: Approved** — clean, well-scoped fix. Correct fail-fast behavior in production.

---

### What this PR does

Removes the weak hardcoded fallback for `SECRET_KEY` in production and wires fail-fast startup enforcement:

1. `ProdConfig.SECRET_KEY = os.environ.get("SECRET_KEY")` — no fallback
2. `validate_required_env_vars(config_name)` now requires `SECRET_KEY` when `config_name == "production"`
3. `create_app` passes `config_name` to the validator and re-raises `ValueError` in production → app won't start without a secret

Chain verified: bare `create_app()` with no FLASK_ENV defaults to `"production"` per the `os.getenv("FLASK_ENV", "production")` fallback — so enforcement kicks in by default. `load_dotenv()` runs at the top of `config.py` (before class definitions), so `ProdConfig.SECRET_KEY` gets the env value correctly at import time in production.

---

### Findings

**Suggestion** — `config.py:32`: `validate_required_env_vars` checks presence but not quality.

```python
missing = [var for var in required_vars if not os.getenv(var)]
```

`SECRET_KEY=x` or `SECRET_KEY=password` would pass. Consider adding a minimum entropy check for production:

```python
if config_name == "production":
    secret = os.getenv("SECRET_KEY", "")
    if len(secret) < 32:
        raise ValueError("SECRET_KEY must be at least 32 characters in production")
```

Not a blocker — presence enforcement is the primary win here — but worth a follow-up issue.

---

**Nit** — `tests/test_app_factory.py:275` (`test_prod_config_no_weak_default`):

```python
assert ProdConfig.SECRET_KEY != "a_default_secret_key_for_development"
```

This passes trivially when `SECRET_KEY` is not set (asserts `None != "..."`). The intent is good — catches regression if someone removes the override — but the assertion is slightly misleading. A comment clarifying the `None` case is expected here would help readers:

```python
# None is expected when SECRET_KEY is not set — the important thing is it's not the hardcoded dev default
assert ProdConfig.SECRET_KEY != "a_default_secret_key_for_development"
```

---

**Note — test isolation vs class-level config evaluation**: `ProdConfig.SECRET_KEY` is a class attribute evaluated once at module import. In tests where `config.py` is already cached, `patch.dict("os.environ", {"SECRET_KEY": "..."})" after the fact doesn't update the class attribute — `validate_required_env_vars` passes (it uses `os.getenv` at call time) but `app.config["SECRET_KEY"]` may stay `None`. This is masked by `_init_token_encryption`'s `try/except`. Non-issue in production (dotenv loads before class eval, so the class attribute is correct). Worth being aware of in test design.

---

**Note — CI**: Only GitGuardian visible in checks. PR body claims 1742/1742 local. No automated test runner showing on this PR — consistent with prior PR observations.

---

### Summary

| Finding | Severity |
|---------|----------|
| No SECRET_KEY strength/entropy check | Suggestion |
| `test_prod_config_no_weak_default` assertion could be clearer | Nit |
| Class-level eval timing in tests | Note (non-prod) |

Core fix is correct and complete. The fail-fast startup behavior, ProdConfig override, and test coverage (4 new focused tests) are all solid.

*Reviewed by code-reviewer-bot. Note: GitHub prevents self-approval — posting as comment with Approved verdict.*

PR #217, opened 2026-04-22
Fix db_app fixture flakiness in batch test runs

### LINKED ISSUE #202: Investigate db_app fixture flakiness in batch test runs
## Context

When running multiple test files together that use the \`db_app\` fixture, intermittent \`sqlite3.OperationalError: no such table\` errors occur during \`db.drop_all()\` teardown. Tests pass individually but fail in batch.

Errors seen:
- \`no such table: workshop_sessions\`
- \`no such table: users\`
- \`no such table: user_settings\`

### The \`db_app\` fixture

**\`tests/conftest.py:273-301\`**:
```python
@pytest.fixture
def db_app():
    app = create_app("development")
    app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///:memory:"
    app.config["TESTING"] = True
    app.config["SCHEDULER_ENABLED"] = False

    with app.app_context():
        db.drop_all()       # line 296 — setup cleanup
        db.create_all()     # line 297 — create fresh schema
        yield app
        db.session.remove() # line 299 — cleanup sessions
        db.drop_all()       # line 300 — teardown ← FAILS HERE
```

### Fixture dependency chain

```
db_app (conftest.py:273)
  ├── app_ctx (conftest.py:303) — wraps db_app in app context
  └── auth_client (conftest.py:310) — authenticated test client with pre-seeded User
```

**~36 test files** use these fixtures across routes, services, models, and schemas.

### Other app fixtures that may interfere

| Fixture | File:Line | DB Strategy | Notes |
|---------|-----------|-------------|-------|
| \`app\` | \`conftest.py:214\` | \`db.create_all()\` only, no \`drop_all()\` cleanup | Used by lighter tests |
| \`authenticated_client\` | \`conftest.py:250\` | Depends on \`app\`, no DB writes | Auth-only tests |
| \`db_app\` | \`conftest.py:273\` | \`drop_all()\` + \`create_all()\` + teardown \`drop_all()\` | DB-write tests |

## Connection Pool Behavior

Flask-SQLAlchemy auto-configures \`StaticPool\` for \`sqlite:///:memory:\` databases — this means all connections share the same in-memory database. This is set internally by Flask-SQLAlchemy, not in Shuffify's config.

**\`config.py:113-122\`** defines \`SQLALCHEMY_ENGINE_OPTIONS\` only for \`ProdConfig\` (PostgreSQL pooling). No explicit engine options exist for the test SQLite configuration.

### What was tried and didn't work

- **PRAGMA foreign_keys=OFF** (PR #198, then reverted): SQLite FK enforcement is OFF by default, and the PRAGMA was set on a throwaway connection that didn't affect the \`drop_all()\` connection.

### Hypotheses to investigate

1. **App context teardown releases the StaticPool connection** — If the \`with app.app_context()\` exit triggers connection cleanup before \`db.drop_all()\` can complete, the in-memory database is destroyed, causing "no such table" on subsequent \`DROP TABLE\` statements.

2. **Multiple \`create_app()\` calls share metadata but not pool state** — Both \`app\` and \`db_app\` fixtures call \`create_app("development")\`. If SQLAlchemy metadata is global but engine instances are per-app, table reflection during \`drop_all()\` may reference tables from one engine on another engine's connection.

3. **Test ordering creates FK constraint ordering issues** — The 15 models have a deep FK dependency tree rooted at \`users\` (11 tables reference it). If \`drop_all()\` attempts to drop child tables after the parent, SQLite may error even with FK enforcement off.

### Potential fixes to evaluate

1. **Use \`checkfirst=True\` explicitly** — \`db.metadata.drop_all(bind=db.engine, checkfirst=True)\` should skip missing tables
2. **Wrap teardown \`drop_all()\` in try/except** — swallow errors during cleanup since the in-memory DB is destroyed anyway
3. **Use a file-based SQLite per test** via \`tmp_path\` — eliminates StaticPool connection sharing entirely
4. **Scope \`db_app\` to \`class\` or \`module\`** — reduce the number of create/destroy cycles

## Effort

Medium — needs StaticPool investigation or fixture restructure. Non-blocking since errors are in teardown, not in test assertions.

chrisrogers37:
## Summary
- **Root cause**: `db_app` and `app` fixtures called `create_app("development")` then overrode `SQLALCHEMY_DATABASE_URI` to `sqlite:///:memory:`. However, the engine was already cached with DevConfig's file-based URI (`sqlite:///shuffify_dev.db`) from `_init_database`, so all "in-memory" tests were actually sharing a file-based SQLite — causing intermittent `no such table` errors during teardown.
- **Fix**: Added `TestConfig` class with the correct `sqlite:///:memory:` URI, updated fixtures to use `create_app("testing")`, and added resilient teardown (`try/except` + `db.engine.dispose()`).
- Fixed `test_hsts_not_present_in_debug_mode` which relied on DevConfig's `DEBUG=True` — now creates its own debug-mode app explicitly.

Fixes #202

## Test plan
- [x] All 10 `test_raid_dedupe.py` tests pass
- [x] 934 DB-heavy tests pass (models/services/routes)
- [x] `test_hsts_not_present_in_debug_mode` passes with explicit debug app
- [x] flake8 clean
- [x] Full suite: 1750 passed, 32 failed (pre-existing app context issues in job executor tests)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **db_app Fixture Flakiness in Batch Test Runs** - Fixed intermittent `no such table` errors during test teardown
+  - Root cause: fixtures called `create_app("development")` then overrode `SQLALCHEMY_DATABASE_URI` to `:memory:`, but the engine was already cached with the file-based URI from `_init_database`
+  - Added `TestConfig` class to `config.py` with `sqlite:///:memory:` URI so the engine is correctly initialized with StaticPool from the start
+  - Updated `db_app` and `app` fixtures to use `create_app("testing")`
+  - Added resilient teardown: `try/except` around `db.drop_all()` and `db.engine.dispose()` for proper cleanup
+
 ### Security
 - **OAuth CSRF State Parameter** - Added `state` parameter to OAuth authorization flow per RFC 6749 Section 10.12
   - Login route generates a cryptographically random state token via `secrets.token_urlsafe(32)` and stores it in session
--- a/config.py
+++ b/config.py
@@ -150,5 +150,23 @@ class DevConfig(Config):
     SCHEDULER_ENABLED = os.getenv("SCHEDULER_ENABLED", "true").lower() == "true"
 
 
+class TestConfig(Config):
+    """Test configuration for isolated, in-memory test runs.
+
+    Uses SQLite in-memory so Flask-SQLAlchemy auto-configures StaticPool,
+    ensuring all connections share the same database within a test.
+    """
+
+    CONFIG_NAME = "testing"
+    TESTING = True
+    SQLALCHEMY_DATABASE_URI = "sqlite:///:memory:"
+    SCHEDULER_ENABLED = False
+
+
 # Dictionary for easy config selection
-config = {"development": DevConfig, "production": ProdConfig, "default": DevConfig}
+config = {
+    "development": DevConfig,
+    "production": ProdConfig,
+    "testing": TestConfig,
+    "default": DevConfig,
+}
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,7 +6,7 @@ including mock Spotify clients, sample data, and Flask app contexts.
 """
 
 import pytest
-from unittest.mock import Mock, MagicMock
+from unittest.mock import Mock
 import time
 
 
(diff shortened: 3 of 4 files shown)


chrisrogers37 (commented):
## Review

Correct diagnosis and fix. The engine-caching root cause is accurately identified — Flask-SQLAlchemy initializes the engine at `init_app` time, so post-creation overrides to `app.config['SQLALCHEMY_DATABASE_URI']` were silently ignored and tests were hitting the file-based dev DB. Providing the correct URI through `TestConfig` before `create_app` returns is the right approach.

**Verdict: Approve** — correct fix, clean implementation, tests passing.

---

### Suggestion — `tests/conftest.py:318-321`

```python
try:
    db.drop_all()
except Exception:
    pass
```

`except Exception: pass` is broader than needed — it silently swallows non-DB errors (import failures, AttributeErrors, etc.). Since the intent is to guard against `InFailedSqlTransaction` after a messy test, narrow it:

```python
from sqlalchemy.exc import SQLAlchemyError
try:
    db.drop_all()
except SQLAlchemyError:
    pass
```

This preserves the defensive teardown behavior while keeping programming errors visible.

---

### Nit — `tests/test_app_factory.py:378`

```python
app = create_app("development")
app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///:memory:"
```

This is the same post-creation override the PR fixes in the fixtures. It's harmless here since `/health` doesn't touch the DB, but a one-line comment (e.g., `# /health doesn't use the DB — URI override is safe here`) would help future readers understand why this specific case is acceptable.

PR #369, opened 2026-05-22
security: regenerate session after OAuth login (session fixation)

### LINKED ISSUE #358: [HIGH] Session fixation — session not regenerated after OAuth login
## Summary

After successful OAuth callback, the session ID is not regenerated. An attacker who can set a victim's session cookie before authentication can hijack the authenticated session.

## Location

`shuffify/routes/core.py` lines 247-320 (callback handler)

## Current Behavior

```python
# Lines 191-196 (login route): clears data but keeps session ID
session.pop("spotify_token", None)
session.pop("user_data", None)
state = secrets.token_urlsafe(32)
session["oauth_state"] = state
session.modified = True

# Lines 250-305 (callback): stores auth data in SAME session
token_data = AuthService.exchange_code_for_token(code)
session["spotify_token"] = token_data
session["user_data"] = user_data
session.modified = True
# No session.regenerate() or equivalent
```

## Attack Scenario

1. Attacker generates a valid session by visiting the app
2. Attacker extracts the session cookie value
3. Attacker tricks victim into using that session cookie (via subdomain cookie injection, XSS on a sibling domain, or network-level attack)
4. Victim authenticates — OAuth tokens stored in the attacker's known session ID
5. Attacker replays the session cookie and gains access to the victim's Spotify account

## Recommended Fix

After successful token exchange in the callback handler, regenerate the session:

```python
# After line 249 (successful token exchange)
old_data = dict(session)
session.clear()
# Flask-Session will generate a new session ID
for key, value in old_data.items():
    if key != "oauth_state":  # Don't carry over consumed state
        session[key] = value
session["spotify_token"] = token_data
```

## Mitigating Factors

- `SameSite=Lax` makes cookie injection harder (but not impossible via subdomain or network attacks)
- `Secure=True` in production prevents network-level cookie interception
- 1-hour session TTL limits the attack window

chrisrogers37:
## Summary

- **Regenerate session ID after successful OAuth callback** to prevent session fixation attacks. After token exchange and user validation succeed, `session.clear()` is called before storing auth data — Flask-Session issues a new session ID, so any pre-auth cookie an attacker planted becomes useless.
- **New test** verifies pre-auth session data is wiped while auth data is preserved.

## Context

Closes #358. The OAuth callback stored auth tokens in the same session that existed before login. An attacker who could plant a session cookie (via subdomain injection, XSS on a sibling domain, or network-level attack) could hijack the authenticated session after the victim logs in.

## Changes

- `shuffify/routes/core.py`: Move `session["spotify_token"]` and `session["user_data"]` writes to after a `session.clear()` call. Token exchange and user validation happen first so we only clear on success.
- `tests/routes/test_core_routes.py`: New `test_session_regenerated_after_successful_oauth` — plants a `pre_auth_marker` in the session, runs OAuth callback, asserts marker is gone and auth data is present.
- `CHANGELOG.md`: Security entry under Unreleased.

## Test plan

- [x] `flake8 shuffify/` — 0 errors
- [x] `pytest tests/routes/test_core_routes.py` — 25/25 pass (including new test)
- [x] `pytest tests/` — 1870 passed, 1 skipped

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Dedicated `CSRFError` handler returns a clear "refresh the page" message
   - CSRF disabled in test config (`WTF_CSRF_ENABLED = False`) so test suite is unaffected
   - Closes #357
+- **Session regenerated after OAuth login to prevent session fixation** — After successful Spotify OAuth callback, the session is now cleared and regenerated with a new session ID before storing authentication data. Previously, the pre-authentication session ID was preserved, allowing an attacker who planted a session cookie to hijack the authenticated session. The fix calls `session.clear()` after token exchange and user validation, then re-populates only the auth data — any pre-auth session state (including attacker-planted markers) is discarded. Closes #358.
 
 ### Changed
 - **Source resolver hygiene — dead field removed, rollback hardening, characterization tests** - Low-priority cleanup chasing #314 and #315 in the scraper module
--- a/shuffify/routes/core.py
+++ b/shuffify/routes/core.py
@@ -247,10 +247,16 @@ def callback():
     try:
         # Exchange code for token
         token_data = AuthService.exchange_code_for_token(code)
-        session["spotify_token"] = token_data
 
         # Validate by fetching user data
         _, user_data = AuthService.authenticate_and_get_user(token_data)
+
+        # Regenerate session to prevent session fixation attacks.
+        # Clearing the session causes Flask-Session to issue a new
+        # session ID on the next response, so any pre-auth session
+        # cookie an attacker planted becomes useless.
+        session.clear()
+        session["spotify_token"] = token_data
         session["user_data"] = user_data
 
         # Upsert user record in database (non-blocking)
--- a/shuffify/routes/core.py
+++ b/shuffify/routes/core.py
@@ -8,14 +8,15 @@ import secrets
 from datetime import datetime, timezone
 
 from flask import (
-    session,
-    redirect,
-    url_for,
-    request,
+    current_app,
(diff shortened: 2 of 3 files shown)


chrisrogers37 (commented):

Fix is mechanically correct and the sequence is exactly right.

**Order of operations:** exchange → validate → `regenerate()` → `clear()` → write auth data. Session is only regenerated on the success path — a failed token exchange or user validation never reaches `regenerate()`, so error paths are unaffected.

**Why both calls are needed:** `session_interface.regenerate(session)` issues a new SID and marks the old server-side record for deletion, but does not wipe the session data dict. `session.clear()` then removes the stale pre-auth data. Without `clear()`, the new SID session would still carry pre-auth markers. Both calls are present and in the right order.

**Bonus fix:** The old code wrote `session["spotify_token"] = token_data` before `authenticate_and_get_user`. This PR moves that write to after user validation — no longer storing a token in the session for an unvalidated user. Correct.

**Flask-Session compatibility:** `current_app.session_interface.regenerate(session)` is the right API. `Flask-Session>=0.8.0` is already pinned in requirements and exposes `regenerate()` on the server-side session interface.

**Test:** Plants `pre_auth_marker`, runs callback, asserts marker gone and auth data present. The `pre_auth_marker not in sess` assertion always fires and is the real behavioral guard. The SID-change assertion is conditional on `getattr(sess, "sid", None)` — if the test Flask-Session backend doesn't expose `sid`, it silently skips. This means the core security property (new SID issued) may not be directly verified in CI. Not blocking — the behavioral test is sufficient to catch a missing `session.clear()` — but a follow-up that makes the SID assertion unconditional (e.g., by verifying the `Set-Cookie` header changes across the response) would close that gap.

1870 passing, 1 skipped — no regressions.

PR #370, opened 2026-05-22
security: add CSRF protection to all state-changing endpoints

chrisrogers37:
## Summary

- Adds Flask-WTF `CSRFProtect` to validate CSRF tokens on all POST/PUT/DELETE/PATCH requests, closing the vulnerability on **51 state-changing endpoints**
- Global `fetch()` wrapper in `base.html` auto-injects `X-CSRFToken` header on mutating requests — zero changes needed to existing AJAX calls (~75 fetch calls covered)
- Hidden `csrf_token` fields added to 3 HTML forms (settings, shuffle, undo) for defense in depth
- Dedicated `CSRFError` handler returns a clear "refresh the page" message
- CSRF disabled in test config so existing test suite is unaffected (1869 pass, 0 fail)

## Approach

Flask-WTF CSRFProtect is the standard CSRF solution for Flask apps. The implementation uses a meta-tag + fetch-wrapper pattern so all existing AJAX calls are protected without individual modification. The fetch wrapper handles both plain object headers and `Headers` instances.

## Files changed

| File | Change |
|------|--------|
| `requirements/base.txt` | Add Flask-WTF>=1.2.0 |
| `shuffify/__init__.py` | Initialize CSRFProtect in app factory |
| `shuffify/error_handlers.py` | Add CSRFError handler + registration |
| `shuffify/templates/base.html` | CSRF meta tag + global fetch wrapper |
| `shuffify/templates/settings.html` | Hidden csrf_token input |
| `shuffify/templates/macros/cards.html` | Hidden csrf_token inputs (shuffle + undo forms) |
| `config.py` | WTF_CSRF_ENABLED = False in TestConfig |
| `tests/test_refresh_and_registry.py` | Disable CSRF in dev-config test fixture |
| `CHANGELOG.md` | Security entry for #357 |

## Test plan

- [x] `flake8 shuffify/` — 0 errors
- [x] `pytest tests/ -v` — 1869 passed, 0 failed
- [x] `/simplify` review completed — hardened fetch wrapper for Headers instances, fixed error message ambiguity, removed style inconsistency

Closes #357

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #372, opened 2026-05-22
fix(security): CSP nonces + redact OAuth auth code from logs

### LINKED ISSUE #360: [MEDIUM] CSP uses unsafe-inline for script-src and style-src
## Summary

The Content-Security-Policy header includes `'unsafe-inline'` for both `script-src` and `style-src`, which weakens XSS protection. If an attacker finds an injection point, inline scripts will execute despite the CSP.

## Location

`shuffify/__init__.py` lines 382-394:

```python
response.headers["Content-Security-Policy"] = (
    "default-src 'self'; "
    "script-src 'self' https://cdn.tailwindcss.com "
    "https://cdn.jsdelivr.net 'unsafe-inline'; "    # <-- here
    "style-src 'self' 'unsafe-inline'; "             # <-- and here
    ...
)
```

## Impact

- `unsafe-inline` in `script-src` allows any injected `<script>` tag or inline event handler to execute
- `unsafe-inline` in `style-src` allows CSS injection (data exfiltration via CSS selectors)
- CSP becomes advisory rather than a hard security boundary against XSS

## Recommended Fix

1. Replace inline scripts with external `.js` files
2. Use nonce-based CSP: generate a per-request nonce, set `script-src 'nonce-{value}'`, and add `nonce="{value}"` to allowed `<script>` tags
3. For `style-src`: Tailwind utility classes are applied via class attributes (safe), but any inline `style=""` attributes need refactoring to classes

```python
import secrets
nonce = secrets.token_urlsafe(16)
response.headers["Content-Security-Policy"] = (
    f"default-src 'self'; "
    f"script-src 'self' https://cdn.tailwindcss.com "
    f"https://cdn.jsdelivr.net 'nonce-{nonce}'; "
    ...
)
```

## Mitigating Factors

- Jinja2 auto-escapes all template variables by default
- No `|safe` filter usage found on user-controlled data
- JavaScript uses `textContent` (not `innerHTML`) for user data
- `escapeHtml()` function used in `schedule-edit.js` for dynamic content
- No XSS injection vectors were identified in this audit

### LINKED ISSUE #361: [MEDIUM] OAuth authorization code logged in callback debug output
## Summary

The OAuth callback handler logs the full query string at DEBUG level, which includes the authorization code. If debug logging is inadvertently enabled in production, this sensitive token appears in logs.

## Location

`shuffify/routes/core.py` line 214:

```python
logger.debug(f"Callback received with args: {request.args}")
```

`request.args` contains:
- `code` — the OAuth authorization code (sensitive, one-time-use)
- `state` — the CSRF state token

## Risk

- Authorization codes are single-use but valid until exchanged — if logged and the log is accessed before exchange, token theft is possible
- Log aggregation services (Sentry, CloudWatch, etc.) may retain this data
- Violates the principle of least privilege for logging

## Recommended Fix

Log only non-sensitive parameters:

```python
logger.debug(
    "Callback received (state=%s, has_code=%s, error=%s)",
    request.args.get("state", "")[:8] + "...",
    bool(request.args.get("code")),
    request.args.get("error"),
)
```

## Mitigating Factors

- DEBUG level is disabled in production (`ProdConfig.DEBUG = False`)
- Authorization codes are single-use and short-lived
- Sentry PII filter (`_strip_pii`) strips some sensitive data, but this log line may not be captured by Sentry (it's below the configured logging integration threshold)

chrisrogers37:
## Summary

- **CSP nonce-based policy (#360)** — Generates a per-request cryptographic nonce (`secrets.token_urlsafe(32)`) and applies it to all 16 inline `<script>` and 4 inline `<style>` tags across 7 templates. CSP header now uses `'nonce-<value>'` instead of `'unsafe-inline'` for both `script-src` and `style-src`.
- **OAuth auth code redacted from logs (#361)** — The `/callback` debug log no longer dumps `request.args` (which contained the authorization code). Replaced with presence/absence indicators only.

Closes #360, closes #361

## Changes

| File | What |
|------|------|
| `shuffify/__init__.py` | `before_request` nonce generation, context processor injection, nonce-based CSP header |
| `shuffify/routes/core.py` | Redacted auth code from callback debug log |
| 7 template files | `nonce="{{ csp_nonce }}"` on all inline script/style tags |
| `tests/test_app_factory.py` | 3 new CSP nonce tests (no unsafe-inline, unique per request, in both directives) |
| `tests/routes/test_core_routes.py` | 1 new test asserting auth code never appears in logs |
| `CHANGELOG.md` | Security section entry |

## Test plan

- [x] `flake8 shuffify/` — 0 errors
- [x] `pytest tests/ -v` — 1873 passed, 0 failed
- [x] `/simplify` reviewed: no reuse, quality, or efficiency issues found
- [ ] Manual verification: load app in browser, confirm inline scripts/styles execute (nonces match CSP)
- [ ] Verify no `unsafe-inline` appears in response headers via browser dev tools


chrisrogers37 (commented):
**Verdict: Ship it** (comment-only — same-identity blocks Approve) — with one merge-order action required before deployment.

---

**CSP nonces (#360)**

Implementation is correct:

- `secrets.token_urlsafe(32)` in `before_request` — 256-bit entropy, generated fresh every request. Good. 
- Context processor injects `csp_nonce` via `getattr(g, "csp_nonce", "")` — the empty-string fallback is safe (an empty nonce simply won't match any script tag).
- `after_request` reads the same `g.csp_nonce` and embeds it in the CSP header. Same nonce for `script-src` and `style-src` — standard practice, fine.
- `'unsafe-inline'` removed from both directives. ✓
- Template count: 16 `<script>` + 4 `<style>` nonce-tags across 7 files — verified against every diff hunk. Matches PR description exactly.

Test suite covers: no `unsafe-inline` in header, unique nonce per request, nonce present in both `script-src` and `style-src`. All three are real guards.

**Auth code redaction (#361)**

Old: `logger.debug(f"Callback received with args: {request.args}")` — raw query string including `code` dumped to logs.

New logs `"present"/"absent"` for code and state (both correct — `code` is the credential, `state` is an OAuth CSRF nonce, both sensitive), and logs the `error` value directly (fine — OAuth error codes like `access_denied` are not credentials).

Test: plants `"AQDx7_SENSITIVE_AUTH_CODE_12345"` in the URL, captures at `DEBUG` level, asserts the raw string never appears. Direct mutation test — reverts the fix and the test fires.

---

**Merge-order action required before deployment**

This PR and PR #370 (CSRF fetch wrapper) both modify `base.html`. PR #370 added an inline `<script>` block that overrides `window.fetch` to inject `X-CSRFToken` on mutating requests. That script does **not** appear in this PR's `base.html` diff — the two PRs branched from different points and haven't seen each other's changes.

If they merge in either order without rebasing, the CSRF fetch wrapper `<script>` will have no `nonce` attribute and will be **blocked by the nonce-based CSP**. Result: `window.fetch` override never executes → all mutating AJAX calls send no `X-CSRFToken` → every state-changing request returns 400. The CSRF protection and the CSP protection will functionally cancel each other out.

Fix: whichever of these two PRs merges second needs a rebase, and the CSRF fetch wrapper block in `base.html` needs `nonce="{{ csp_nonce }}"` added before deploy.

Nothing wrong with either PR in isolation — this is a sequencing note for you.

Issue #367, opened 2026-05-22
[LOW] APScheduler uses pickle for job serialization in database


chrisrogers37:
## Summary

APScheduler serializes job arguments using Python pickle when storing them in the SQLAlchemy job store. Pickle deserialization of untrusted data is a known remote code execution vector.

## Location

`shuffify/scheduler.py` — APScheduler initialization with SQLAlchemyJobStore

## Risk

If an attacker gains write access to the `apscheduler_jobs` database table, they could craft a malicious pickled payload that executes arbitrary code when the scheduler deserializes it.

## Current Protections

- Jobs are created only through authenticated application routes
- Database access requires valid credentials
- Job definitions are server-controlled (user input is validated by Pydantic before reaching the scheduler)
- No user input is directly pickled

## Recommended Fix

This is an accepted-risk scenario for most APScheduler deployments. For hardening:

1. Ensure the `apscheduler_jobs` table has restricted write access at the database level
2. Consider migrating to APScheduler 4.x which supports JSON-serializable job stores
3. Monitor for unexpected job entries via the activity log

## Mitigating Factors

- Database is behind authentication (Neon PostgreSQL in production)
- No untrusted data flows into pickle serialization
- APScheduler is a widely-used library with this as a known design choice
- All job parameters are validated before scheduling

Issue #366, opened 2026-05-22
[LOW] No account lockout or brute-force protection beyond rate limiting


chrisrogers37:
## Summary

The application has rate limiting on auth endpoints (10/min login, 20/min callback) but no account-level lockout, progressive delays, or failed-attempt tracking.

## Current State

- Flask-Limiter applied to `/login` (10/min) and `/callback` (20/min)
- `LoginHistory` model tracks successful logins only
- No failed login counter or lockout mechanism

## Risk

An attacker can attempt OAuth state brute-forcing at 20 requests/minute indefinitely. While the 32-byte state token makes brute-forcing infeasible, the lack of lockout is a defense-in-depth gap.

## Recommended Fix

This is low priority given OAuth delegation, but for defense-in-depth:

1. Track failed callback attempts (state mismatch) per IP
2. After N failures, increase rate limit cooldown or block IP temporarily
3. Alert on unusual patterns (many state mismatches from one IP)

## Mitigating Factors

- OAuth 2.0 delegates authentication to Spotify — no local passwords to brute-force
- 32-byte state token (`secrets.token_urlsafe(32)`) is computationally infeasible to guess
- Rate limiting provides baseline protection
- 1-hour session TTL limits attack window

Issue #365, opened 2026-05-22
[LOW] Missing .dockerignore file


chrisrogers37:
## Summary

The repository has no `.dockerignore` file. Docker builds copy the entire build context into the daemon, which may include unnecessary or sensitive files.

## Risk

Without `.dockerignore`, the Docker build context includes:
- `.git/` directory (full repository history, potentially containing old secrets)
- `tests/` directory (unnecessary in production image)
- `documentation/` directory
- Any local `.env` files present during build
- `__pycache__/` directories
- `.flask_session/` local session files

While the Dockerfile itself is well-structured (non-root user, health checks), the build context may inadvertently include sensitive data in image layers.

## Recommended Fix

Create `.dockerignore`:

```
.git
.gitignore
.github
.env
.env.*
!.env.example
__pycache__
*.pyc
*.pyo
tests/
documentation/
.coverage
htmlcov/
.pytest_cache/
.mypy_cache/
.flake8
.flask_session/
*.db
*.sqlite3
venv/
.venv/
```

## Mitigating Factors

- The Dockerfile copies only specific directories (`requirements/` first, then `.`)
- No secrets are hardcoded in source files
- `.env` is gitignored (but may exist locally during build)

PR #373, opened 2026-05-22
security: low-severity hardening wave 3 (#364-#367)

### LINKED ISSUE #364: [LOW] Spotify OAuth tokens not revoked on logout
## Summary

The logout handler clears the Flask session but does not revoke the OAuth access or refresh tokens at Spotify. Tokens remain valid until they naturally expire.

## Location

`shuffify/routes/core.py` lines 336-349:

```python
@main.route("/logout")
def logout():
    # Records logout event, then:
    session.clear()
    # No Spotify token revocation call
```

## Risk

If an attacker obtains a session token (e.g., from a shared computer), the access token remains valid for its full TTL even after the user logs out. The encrypted refresh token in the database also remains active.

## Recommended Fix

Call Spotify's token revocation endpoint before clearing the session:

```python
import requests

@main.route("/logout")
def logout():
    token = session.get("spotify_token", {}).get("access_token")
    if token:
        try:
            requests.post(
                "https://accounts.spotify.com/api/token/revoke",
                data={"token": token},
                auth=(client_id, client_secret),
                timeout=5,
            )
        except Exception:
            pass  # Best-effort revocation
    session.clear()
```

Note: Spotify does not currently document a public revocation endpoint. If unavailable, document this as an accepted risk.

## Mitigating Factors

- Access tokens have short TTL (~1 hour)
- Session is stored server-side (Redis) — clearing the session removes the application's copy
- Encrypted refresh token in DB requires SECRET_KEY to decrypt

chrisrogers37:
## Summary

Addresses four LOW-severity security findings in a single changeset:

- **#364 — Best-effort OAuth token revocation on logout.** Logout now POSTs to Spotify's revocation endpoint before clearing the session. Spotify does not currently support RFC 7009, so the call is fire-and-forget (2s timeout, DEBUG logging). The DB-stored refresh token is intentionally preserved for scheduled operations.
- **#365 — `.dockerignore` added.** Excludes `.git`, `.env`, `__pycache__`, `tests/`, `documentation/`, `node_modules/`, `.flask_session/`, and other non-runtime files from Docker build context. Reduces image size and prevents accidental secret inclusion.
- **#366 — OAuth callback rate limit tightened.** `/callback` reduced from 20/min to 10/min per IP (matching `/login`). Traditional account lockout is not applicable: this is an OAuth-only app where users authenticate via Spotify, never through Shuffify. Documented in code.
- **#367 — APScheduler pickle serialization risk documented.** `SQLAlchemyJobStore` in APScheduler 3.x uses pickle with no config knob to swap serializers. The risk is low (exploitation requires DB write access). Documented in code with a note to re-evaluate on APScheduler 4.x upgrade.

## Test plan

- [x] `pytest tests/spotify/test_auth.py` — 4 new tests for `revoke_token` (200, non-200, network error, timeout)
- [x] `pytest tests/routes/test_core_routes.py` — 2 new tests for logout revocation (called with correct token, failure does not block logout)
- [x] `flake8 shuffify/` — clean
- [ ] Verify `.dockerignore` excludes expected files: `docker build --no-cache -t shuffify-test . && docker run --rm shuffify-test ls /app/tests` should fail (tests excluded)
- [ ] Verify rate limit: hit `/callback` 11 times in 1 minute from same IP, confirm 429 on 11th

Closes #364, #365, #366, #367

Generated with [Claude Code](https://claude.com/claude-code)


chrisrogers37 (commented):

All four fixes are correct. Reviewed each independently.

---

**#364 — Token revocation on logout**

Implementation is mechanically correct. The route calls `AuthService.revoke_access_token(access_token)` inside `try/except Exception: pass` before session clear — revocation can never block logout.

`revoke_token` in `auth.py` is a correct RFC 7009 call: `token` + `token_type_hint=access_token` in the request body, HTTP Basic auth using `client_id/client_secret`, `timeout=2`. The 2s timeout is appropriate for fire-and-forget. The comment is honest — Spotify doesn't honor this endpoint today, so the call returns non-200 and is silently ignored. When Spotify adds support, it starts working with no code change.

Tests are real guards:
- `test_logout_attempts_token_revocation`: `assert_called_once_with("test_token")` — remove the call from the route, test fails.
- `test_logout_continues_when_revocation_fails`: side_effect Exception, asserts 302 — remove the `except Exception: pass`, test fails.
- `revoke_token` unit tests cover: 200 → True, non-200 → False, network error → False, Timeout → False. All correct.

**#365 — .dockerignore**

Correct and comprehensive. Excludes `.env`, `.env.*`, `*.pem`, `*.key`, `.git`, `tests/`, `documentation/`, `__pycache__`, `venv`, `.flask_session`. The `*.md` exclusion is broad but `!requirements/*.md` is present to exempt any markdown in requirements/. `requirements/*.txt` files are not markdown — they're included in the context and available for `pip install`. Excluding `Dockerfile` from the build context is fine: Docker reads it from disk before the context transfer.

**#366 — Callback rate limit tightened**

`/callback` 20/min → 10/min, now matching `/login`. The added code comment correctly documents why traditional account lockout doesn't apply (OAuth-only app, credentials entered on Spotify's login page, not Shuffify). Change is correct and the reasoning is sound.

**#367 — APScheduler pickle documentation**

Comment-only. The characterization is accurate: APScheduler 3.x `SQLAlchemyJobStore` uses pickle with no serializer config knob; the risk is bounded because the pickle boundary is internal (app writes its own jobs, reads them back) and exploitation requires DB write access. Noted for re-evaluation at APScheduler 4.x upgrade. No behavioral change.

---

**Recurring pattern worth flagging:** This is the third PR in this wave where `shuffify/__init__.py` or `scheduler.py` picks up cosmetic line-length reformatting unrelated to the security fixes (89 deletions / 38 additions in `scheduler.py` are almost entirely formatting). Nothing wrong with the reformatting, but it adds diff noise that makes security changes harder to isolate in a `git blame`. Consider keeping formatting-only commits separate from security fixes in future waves.

Issue #354, opened 2026-05-20
Stale Dependabot config watches removed spotipy package


chrisrogers37:
## Finding

**Severity:** Low  
**Category:** Dependency hygiene  
**File:** `.github/dependabot.yml:40`

`dependabot.yml` watches for `spotipy` under `production-dependencies`, but `spotipy` is not in any requirements file. The project migrated to a direct HTTP client (`shuffify/spotify/http_client.py`). The stale pattern means Dependabot silently does nothing for that entry.

## Recommendation

Remove the `spotipy` pattern from `dependabot.yml`.

chrisrogers37:
Resolved in PR #376 (merged)

Issue #353, opened 2026-05-20
Test coverage gaps: activity routes, schedule_utils, algorithm boundaries


chrisrogers37:
## Finding

**Severity:** Low  
**Category:** Test coverage  
**Files:**
- `routes/activity.py` — no `test_activity_routes.py` exists. Every other route module has a test file.
- `services/schedule_utils.py` — tested only as a side-effect from schema tests. No dedicated `test_schedule_utils.py`. Missing: malformed time strings, `ValueError` for non-time-capable frequencies.
- `shuffle_algorithms/` — `AlbumSequenceShuffle` inlines URI extraction instead of using shared `extract_uris()` utility, inconsistent with all other algorithms.

## Recommendation

Add test files for activity routes and schedule_utils. Refactor `AlbumSequenceShuffle` to use `extract_uris()`.

chrisrogers37:
Resolved in PR #376 (merged)

Issue #352, opened 2026-05-20
Schedule.target_playlist_id uses String(64) while all other playlist ID columns use String(255)


chrisrogers37:
## Finding

**Severity:** Low  
**Category:** Schema consistency  
**File:** `shuffify/models/db.py:483-485`

`Schedule.target_playlist_id` is `String(64)` while every other table (`UpstreamSource`, `PlaylistSnapshot`, `PlaylistPair`, `RaidPlaylistLink`, `PendingRaidTrack`, `PlaylistPreference`, `TrackLock`) uses `String(255)`. Spotify playlist IDs are 22-char base-62 so `String(64)` works, but the inconsistency makes migrations harder to reason about and creates silent risk if Spotify changes ID format.

## Recommendation

Align to `String(255)` via migration for consistency.

chrisrogers37:
Resolved in PR #376 (merged)

Issue #351, opened 2026-05-20
Schema validation gaps: track URI format, ExternalPlaylistRequest mutual exclusion


chrisrogers37:
## Finding

**Severity:** Low  
**Category:** Validation  
**Files:**
- `schemas/requests.py:220-227` — `WorkshopCommitRequest.track_uris` validates `uri.startswith("spotify:track:")` but not the 22-char base-62 ID that follows. Malformed URIs pass validation, fail at the Spotify API.
- `schemas/requests.py:252-290` — `ExternalPlaylistRequest` accepts both `url` and `query` simultaneously. The code handles `url` first, but the schema doesn't document or enforce priority.

## Recommendation

Add a simple `len + isalnum` check after the prefix. Add mutual-exclusion or priority documentation to `ExternalPlaylistRequest`.

chrisrogers37:
Resolved in PR #376 (merged)

PR #376, opened 2026-05-22
fix: tech-debt sprint — indexes, validation, prod overlay, tests

### LINKED ISSUE #345: docker-compose.yml mounts host source into container — no prod override
## Finding

**Severity:** Medium  
**Category:** Infrastructure  
**File:** `docker-compose.yml:12`

The volume mount `- .:/app` overrides everything `COPY . .` placed in the image. Reasonable for local dev, but there is no `docker-compose.override.yml` or separate prod compose file. Running `docker-compose up` in production or CI gets host-mounted source (including `.env` and local dev files) inside the container, defeating the immutable image model.

## Recommendation

Split into `docker-compose.yml` (prod, no volume mount) and `docker-compose.override.yml` (dev, with volume mount). Add a `.dockerignore` to exclude sensitive files.

chrisrogers37:
## Summary

Six tech-debt issues resolved in one branch:

- **#345** — Add `docker-compose.prod.yml` that drops the host source mount for production deployments
- **#348** — Alembic migration adding missing indexes on `PendingRaidTrack.user_id`, `JobExecution.status`/`started_at`, `PlaylistPreference.spotify_playlist_id`
- **#351** — Stricter track URI regex validation (`spotify:track:<alphanumeric>`); `ExternalPlaylistRequest` now enforces mutual exclusion (url xor query)
- **#352** — Widen `Schedule.target_playlist_id` from `String(64)` to `String(255)` via Alembic migration, matching all other playlist ID columns
- **#353** — New test coverage: activity routes (3), schedule_utils helpers (13), cross-algorithm boundaries (6), WorkshopCommitRequest (5), ExternalPlaylistRequest (6)
- **#354** — Remove stale `spotipy` entry from Dependabot production-dependencies group

## Test plan

- [x] All 1912 tests pass (`pytest tests/ -v`)
- [x] flake8 clean (`flake8 shuffify/` — 0 errors)
- [x] `/simplify` review completed (findings logged, none blocking)
- [ ] Verify migrations apply cleanly on a Neon branch before prod
- [ ] Confirm `docker-compose -f docker-compose.yml -f docker-compose.prod.yml up` works as expected

Closes #345, #348, #351, #352, #353, #354

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -36,7 +36,6 @@ updates:
       production-dependencies:
         patterns:
           - "Flask*"
-          - "spotipy"
           - "gunicorn"
           - "redis"
           - "pydantic"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- **docker-compose.prod.yml** - Production override that drops the host source mount (`-v .:/app`) and adds `restart: unless-stopped`. Closes #345
+- **docker-compose source mount fix** - Moved host source mount (`-v .:/app`) from `docker-compose.yml` into new `docker-compose.override.yml` (auto-loaded in dev, skipped in prod). `docker-compose.prod.yml` adds `restart: unless-stopped`. Closes #345
 - **Database indexes** - Added missing indexes on PendingRaidTrack(user_id), JobExecution(status, started_at), PlaylistPreference(spotify_playlist_id) via Alembic migration. Closes #348
 - **Track URI format validation** - WorkshopCommitRequest now validates full `spotify:track:<22-char-id>` format instead of prefix-only check. Closes #351
 - **ExternalPlaylistRequest mutual exclusion** - Providing both `url` and `query` now raises a validation error. Closes #351
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **docker-compose.prod.yml** - Production override that drops the host source mount (`-v .:/app`) and adds `restart: unless-stopped`. Closes #345
+- **Database indexes** - Added missing indexes on PendingRaidTrack(user_id), JobExecution(status, started_at), PlaylistPreference(spotify_playlist_id) via Alembic migration. Closes #348
+- **Track URI format validation** - WorkshopCommitRequest now validates full `spotify:track:<22-char-id>` format instead of prefix-only check. Closes #351
+- **ExternalPlaylistRequest mutual exclusion** - Providing both `url` and `query` now raises a validation error. Closes #351
+- **Test coverage** - Added tests for activity routes, schedule_utils edge cases, cross-algorithm boundary conditions, and schema validation. Closes #353
+
 ### Changed
 - **Replaced deprecated `FLASK_ENV` with `APP_CONFIG`** - Config selector env var renamed from `FLASK_ENV` (removed in Flask 3.0) to `APP_CONFIG` across app factory, run.py, Dockerfile, and .env.example. Closes #355
 - **`SESSION_COOKIE_SECURE` defaults to `True`** - Base Config now defaults secure cookies on; DevConfig and TestConfig explicitly override to `False`. Closes #355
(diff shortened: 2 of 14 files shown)


chrisrogers37 (commented):
**Verdict: Mechanical fixes**

Mostly clean. Two specific issues need resolution; everything else is correct.

---

**#354 — Dependabot** — `spotipy` removed from production group. ✓

**#348 — Indexes migration** — Four indexes added: `PendingRaidTrack.user_id`, `JobExecution.status`, `JobExecution.started_at`, `PlaylistPreference.spotify_playlist_id`. Chain: `e5f6a7b8c9d0` → `f1a2b3c4d5e6` → `a2b3c4d5e6f7`. `JobExecution.__table_args__` in `db.py` matches the migration for those two indexes — consistent. ✓

**#352 — Column width** — `Schedule.target_playlist_id` widened String(64) → String(255) in both the model and the Alembic migration. `batch_alter_table` is correct for environments that don't support direct ALTER COLUMN (SQLite in tests). ✓

**#353 — Test coverage** — Activity route tests, schedule_utils tests (TIME_CAPABLE_FREQUENCIES, TIME_RE, build_cron), algorithm boundary tests (2-track, 500-track, same-artist, missing URI, section_count=1, two-track PercentageShuffle), WorkshopCommitRequest and ExternalPlaylistRequest schema tests all look correct. The boundary tests use `algo.shuffle(tracks)` and assert count; this tests real behavior. ✓

**db.py reformatting** — 426 deletions / 140 additions, nearly all mechanical line-length normalization. Semantic changes are only the String(64) → String(255) and the `JobExecution.__table_args__` addition, both correct. ✓

---

**Issue 1 — TRACK_URI_RE doesn't enforce 22-char Spotify IDs (CHANGELOG claim is wrong)**

The implemented regex:
```python
TRACK_URI_RE = re.compile(r"^spotify:track:[a-zA-Z0-9]{1,}$")
```

`{1,}` is `+` — it accepts any length ≥ 1. `spotify:track:x` passes validation.

The CHANGELOG entry says: *"validates full `spotify:track:<22-char-id>` format"*. That's not what the code does.

**Fix:** either enforce the actual Spotify ID length (`{22}`) or correct the CHANGELOG to say "validates alphanumeric URI format" without the 22-char claim. The test `test_rejects_prefix_only` only covers the zero-char case — add `test_rejects_short_id` to exercise the length constraint if you enforce it.

---

**Issue 2 — docker-compose.prod.yml may not actually remove the source mount**

Docker Compose merge semantics for `volumes:` concatenate the two files' volume lists — the override file can add volumes but cannot remove volumes that appear in the base file.

The override as written:
```yaml
services:
  app:
    volumes:
      - flask_session:/app/.flask_session
    restart: unless-stopped
```

If `docker-compose.yml` has `- .:/app` in the `app` service, running `docker compose -f docker-compose.yml -f docker-compose.prod.yml up` would result in BOTH volumes present — the source mount would NOT be removed.

If the source mount lives in `docker-compose.override.yml` (auto-loaded in dev) rather than the base file, this pattern is correct: running explicitly with `-f docker-compose.yml -f docker-compose.prod.yml` skips the override.yml and the source mount is absent.

Before merging, confirm which file actually contains `- .:/app` and add a note to the PR body about the correct invocation. If it's in the base file, the overlay needs a different approach (e.g., restructure to a dev overlay pattern).

chrisrogers37 (commented):

Both requested fixes verified on current HEAD (`7feb333`).

---

**Issue 1 — TRACK_URI_RE** ✅

Regex is now `r"^spotify:track:[a-zA-Z0-9]{22}$"` — exactly 22 alphanumeric chars enforced. Mutation check: reverting to the old `startswith("spotify:track:")` causes `test_rejects_prefix_only` to fail (old code would not raise on `"spotify:track:"`; test expects `ValidationError`). Test is live. Workshop route tests now use properly-formed 22-char URIs, which would also fail against a loose regex.

One minor gap: `test_rejects_short_id` (21-char ID) was called out in the prior review but not added. The regex is correct; the missing boundary test is not a blocker.

---

**Issue 2 — docker-compose source mount** ✅

Source mount (`- .:/app`) removed from `docker-compose.yml` (base) entirely. Moved into new `docker-compose.override.yml` (auto-loaded by Docker Compose in dev). `docker-compose.prod.yml` adds only `restart: unless-stopped`. Running prod explicitly with `-f docker-compose.yml -f docker-compose.prod.yml` skips the override — no source mount present. Correct compose pattern; merge-semantics concern from prior review is resolved.

---

**CI:** Shows pending / 0 statuses — no pipeline results visible via API. PR body claims 1912 tests passing. Confirm CI green before merge.

Ready to merge.

Issue #350, opened 2026-05-20
Spotify client minor issues: search limit capped at 10, retry/refresh budget overlap


chrisrogers37:
## Finding

**Severity:** Low  
**Category:** Spotify API  
**Files:**
- `spotify/api.py:833,923` — `search_tracks` and `search_playlists` clamp limit to `max(1, min(limit, 10))`. Spotify supports up to 50 for search. Artificially throttles workshop search quality.
- `spotify/http_client.py:162-188` — Token refresh after 401 consumes a retry loop iteration via `continue`. The retry-on-network-error budget and the one-token-refresh budget share the same counter.
- `spotify/api.py:306-309` — `item.get("track") or item.get("item")` fallback for Spotify's key migration is correct but the `"item"`-primary path has no test.

## Recommendation

Raise search limit cap to 50 (Spotify max). Consider separating the refresh attempt from the retry budget. Add a test for the `"item"` key path.

PR #378, opened 2026-05-25
fix: search_playlists total_tracks returns 0 — wrong key lookup

### LINKED ISSUE #335: search_playlists total_tracks always returns 0 — wrong key lookup
## Finding

**Severity:** Medium  
**Category:** Bug  
**File:** `shuffify/spotify/api.py:858-878`

The code tries `item.get("items", item.get("tracks", {}))` to find the track count. Spotify's search response uses `"tracks": {"total": N}` on playlist summary objects. The lookup order is inverted — `"items"` doesn't exist on playlist summaries, so `total_key` falls through to `{}` and `total_tracks` is always 0 in search results.

## Recommendation

Fix the lookup to `item.get("tracks", item.get("items", {}))` and add a test covering the expected response shape.

chrisrogers37:
## Summary
- Fixed `search_playlists` returning `total_tracks: 0` for every playlist due to incorrect key lookup
- The code tried `item["items"]` (which doesn't exist on playlist search results) before falling back to `item["tracks"]` — the fallback was correct but the primary lookup always matched `None`, causing `.get("total", 0)` to return 0
- Simplified to `item.get("tracks", {})` directly

Closes #335

## Test plan
- [ ] Existing test suite passes (1911 passed, pre-existing `test_search_clamps_limit_max` failure unrelated — tracked in #350)
- [ ] `flake8 shuffify/` clean
- [ ] Manual verification: search for playlists, confirm `total_tracks` is non-zero

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Hardcoded `SECRET_KEY` fallback removed from base Config** - `Config.SECRET_KEY` no longer defaults to `"a_default_secret_key_for_development"`. DevConfig auto-generates a random key; TestConfig uses an explicit test key; ProdConfig requires the env var (unchanged). Closes #341
 
 ### Fixed
+- **`search_playlists` total_tracks always returned 0** - Fixed wrong key lookup: the code tried `item["items"]` (which doesn't exist on playlist search results) before falling back to `item["tracks"]`. Simplified to use `item.get("tracks", {})` directly. Closes #335
 - **`logging.basicConfig(level=DEBUG)` no longer runs at import time** - Moved into `create_app()` with level set to INFO in production, DEBUG otherwise. Prevents debug logging from polluting any process that imports the shuffify package. Closes #338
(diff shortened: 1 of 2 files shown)


chrisrogers37 (commented):
**Verdict: Ship it**

**Fix correctness:** `item.get("tracks", {})` is the correct lookup. The Spotify simplified playlist object (returned in search results) has `"tracks": {"href": ..., "total": N}` at the top level — no `"items"` key exists on the playlist object itself. The old comment claiming "API version differences" was speculative; there is no Spotify API variant that uses `"items"` for the per-playlist total reference.

**Old code defect:** `item.get("items", item.get("tracks", {}))` — Python evaluates the default expression eagerly, so this is semantically "if 'items' key exists, use it; otherwise use 'tracks'." If an individual playlist object unexpectedly carried an `"items"` key (e.g. a future API variant or a mis-shaped response), it would shadow the correct "tracks" lookup. Simplifying to the single correct key removes that fragility.

**Mutation test:** Reverting to the old code — all current tests pass unchanged, because the test fixture items have no `"items"` key, so old code falls through to "tracks" and produces the same result. The PR doesn't add a test that exercises the shadowing path (an item with `{"items": None, "tracks": {"total": 50}}`). Low-value gap: Spotify never sends this shape, so a test for it would be testing a hypothetical. Acceptable.

**`item is None` guard above the change:** retained. Correct placement.

PR #379, opened 2026-05-25
fix: logout handler — log swallowed exception, use cached user lookup

### LINKED ISSUE #349: Logout handler: swallowed exception + double user lookup
## Finding

**Severity:** Low  
**Category:** Code smell  
**File:** `shuffify/routes/core.py:341-366`

Two issues in the `logout` handler:

1. The activity-log `try/except Exception: pass` block (lines 353-366) swallows exceptions silently with no logging. Every other non-blocking failure in the codebase uses `logger.warning()`.
2. `UserService.get_by_spotify_id(user_data["id"])` is called twice in sequence — once for the logout event, once for the activity log. The second lookup is an identical query; the result of the first is never reused.

## Recommendation

Add `logger.warning()` to the except block. Reuse the first user lookup result.

chrisrogers37:
## Summary
- Token revocation `except Exception: pass` now logs via `logger.warning()` — consistent with every other non-blocking failure in the codebase
- Replaced `UserService.get_by_spotify_id(user_data["id"])` with `get_db_user()`, which uses the session-cached PK for a faster lookup and avoids duplicating the string-key query
- Also switched the second `except` block from f-string to `%s`-style logging (avoids string formatting when the log level is suppressed)

Closes #349

## Test plan
- [ ] 9 logout-related tests pass (`pytest -k logout`)
- [ ] Full suite passes (1922 passed, pre-existing `test_search_clamps_limit_max` failure unrelated — tracked in #350)
- [ ] `flake8 shuffify/routes/core.py` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - **`search_playlists` total_tracks always returned 0** - Fixed wrong key lookup: the code tried `item["items"]` (which doesn't exist on playlist search results) before falling back to `item["tracks"]`. Simplified to use `item.get("tracks", {})` directly. Closes #335
+- **Logout handler: swallowed exception + redundant user lookup** - Token revocation failure now logs a warning instead of silently swallowing with bare `except: pass`. Replaced direct `UserService.get_by_spotify_id()` call with the session-cached `get_db_user()` helper. Closes #349
 - **`logging.basicConfig(level=DEBUG)` no longer runs at import time** - Moved into `create_app()` with level set to INFO in production, DEBUG otherwise. Prevents debug logging from polluting any process that imports the shuffify package. Closes #338
 - **`REDIS_URL` now fails loudly in production** - `_init_redis` raises `RuntimeError` if Redis is unavailable in production instead of silently falling back to filesystem sessions. Development still falls back gracefully. Closes #339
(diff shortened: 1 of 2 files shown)


chrisrogers37 (commented):
**Verdict: Ship it**

**Exception handling:** `except Exception: pass` → `except Exception as e: logger.warning(...)`. Correct. Silent swallows on token revocation failures were masking auth service errors invisibly. Logging at warning level is right — revocation failure is non-fatal for logout but worth knowing about.

**`get_db_user()` substitution:** the helper is already imported (`from shuffify.routes import get_db_user`, line 26) and used elsewhere in `core.py` (line 79). The old code re-fetched via `UserService.get_by_spotify_id(user_data["id"])` — a live DB query using a raw session key lookup — when the session-cached helper already has the resolved user. The replacement is correct and cheaper.

**Null safety:** new code: `if db_user:` before using it. If `get_db_user()` returns `None` (no session, expired token), the activity logging block is skipped cleanly. The surrounding `try/except Exception` catches anything `get_db_user()` might raise.

**Logging format:** `f"Failed to record logout: {e}"` → `"Failed to record logout: %s", e` — correct; lazy `%`-style avoids string formatting when the logger's level is above WARNING.

**Session.clear() and redirect:** unchanged and unconditional — logout completes regardless of activity-recording failures. Correct.

One non-blocking note: the token revocation `try/except` block and the activity-recording `try/except` block are now both warning-level failures with no observable difference to the user. If token revocation fails silently but the session is cleared, the token remains valid until it expires. Worth a comment noting that behavior, but not blocking.

PR #383, opened 2026-05-27
fix(executors): per-playlist execution lock to prevent concurrent-job races

chrisrogers37:
## Summary

Yesterday at 09:00 UTC, schedule 4 (`shuffle` on **Long Live The Boys**) failed for the first time in 10+ days:

```
Schedule 4: shuffle verification failed —
expected 49 tracks, got 44, missing 5, extra 0
```

Root cause was a race between schedules sharing a cron firing time on the same playlist:

| Schedule | Job | Target | `started_at` | Duration | Result |
|---|---|---|---|---|---|
| 4 | shuffle | Long Live The Boys | 09:00:01.438 | 2.7s | **failed** (verification: 49 expected, 44 seen) |
| 10 | rotate | Long Live The Boys | 09:00:01.440 | 13.1s | success, +5 tracks |

Both schedules fire at `0 9 * * *`. APScheduler handed them to the executor 2 ms apart. Rotate's "add 5 new tracks" step landed inside shuffle's verification window — exactly accounting for the `+5` arithmetic mismatch shuffle observed.

The F2 snapshot rollback (PR #306) could not fire because user 1 has `auto_snapshot_enabled = False` in `user_settings` (since 2026-03-31), so `_restore_job_snapshots` found nothing to restore and fell through to plain `failed`. **F2 is working as designed; the absent rollback is an operational consequence of auto-snapshots being off, not a code bug.** Re-enabling auto-snapshots restores the safety net independently of this PR.

## Fix

New module `shuffify/services/playlist_lock.py` exposes a `playlist_lock` context manager that takes a Postgres session-level `pg_try_advisory_lock` keyed on a stable blake2b-64 signed hash of the target playlist ID. `JobExecutorService.execute` wraps its work in this lock so two schedules targeting the same playlist serialize regardless of cron alignment.

Design details:
- **Dedicated connection** borrowed from the engine pool (not `db.session`) — SQLAlchemy commits inside the executor body don't release it, since session-scoped advisory locks die when the session's connection returns to the pool.
- **No-op on SQLite** — dev/test is single-process, no race possible.
- **60-second timeout** with 0.25s poll interval. On timeout the run is skipped with a `WARNING` log and the next cron fire retries — losing one run is preferable to corrupting the playlist with interleaved writes.
- Follows the same `pg_try_advisory_lock` pattern already established in `shuffify/scheduler.py:81` for scheduler-instance singletonization.

## Why advisory locks vs. alternatives

- **Re-time crons (move schedule 4 to 09:20):** fixes today's specific collision but the next time the user adds a schedule the same trap recurs. Doesn't generalize.
- **Application-level lock table:** needs schema migration, cleanup logic for stale locks, retry semantics on crash recovery. Heavier than needed.
- **APScheduler `max_instances=1`:** prevents the same job from running twice concurrently; doesn't help across different schedules targeting the same playlist.
- **In-memory mutex:** works inside one process but breaks the moment you scale to multi-worker.

Postgres advisory locks: zero schema impact, OS-free, auto-released on connection close, already a project pattern.

## Test plan

- [x] `flake8 shuffify/` — clean
- [x] `pytest tests/services/test_playlist_lock.py tests/services/test_job_executor_service.py::TestExecuteLocking -v` — 9/9 pass (key derivation, SQLite no-op, Postgres acquire/release, timeout path, exception cleanup, executor skip-on-timeout, executor happy path)
- [x] `pytest tests/ -q` — **1913 passed, 1 skipped.** Two pre-existing failures in `tests/routes/test_activity_routes.py` predate this branch — confirmed to fail on clean main, out of scope here.
- [ ] After merge + deploy: schedule 4 fires at 09:00 UTC tomorrow without a verification failure. Expected behavior: either schedule 4 or schedule 10 holds the lock; the other waits a few seconds then proceeds. Both end up `success`.

## Notes / follow-ups (not in scope)

- The "no snapshot available" message inside `_restore_job_snapshots` doesn't distinguish "user disabled auto-snapshots" from "snapshot service errored". Worth a structured `SCHEDULE_RUN_ROLLBACK_SKIPPED` activity log entry in a future pass.
- User 1 has auto-snapshots off; users 2 and 3 have them on. Consider re-enabling for user 1 to get F2 protection back.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **Per-playlist execution lock** - New `shuffify.services.playlist_lock.playlist_lock` context manager serializes `JobExecutorService.execute()` runs that target the same playlist, using Postgres session-level advisory locks on a dedicated pool connection. Prevents the race where two schedules sharing a cron firing time on the same playlist interleave reads and writes (e.g. a shuffle's post-write verification reading mid-flight rotate state and failing with `expected N tracks, got N-K`). No-op on SQLite (dev/test, single-process). Lock acquisition has a 60-second timeout; on timeout the run is skipped with a `WARNING` log and the next cron fire retries.
 - **docker-compose source mount fix** - Moved host source mount (`-v .:/app`) from `docker-compose.yml` into new `docker-compose.override.yml` (auto-loaded in dev, skipped in prod). `docker-compose.prod.yml` adds `restart: unless-stopped`. Closes #345
 - **Database indexes** - Added missing indexes on PendingRaidTrack(user_id), JobExecution(status, started_at), PlaylistPreference(spotify_playlist_id) via Alembic migration. Closes #348
 - **Track URI format validation** - WorkshopCommitRequest now validates full `spotify:track:<22-char-id>` format instead of prefix-only check. Closes #351
--- a/shuffify/services/executors/base_executor.py
+++ b/shuffify/services/executors/base_executor.py
@@ -15,6 +15,7 @@ from typing import List
 
 from shuffify.models.db import db, Schedule, JobExecution, User
 from shuffify.services.base import safe_commit
+from shuffify.services.playlist_lock import playlist_lock
 from shuffify.services.token_service import (
     TokenService,
     TokenEncryptionError,
@@ -309,8 +310,6 @@ class JobExecutorService:
         other's verification (e.g. a shuffle + a rotate that both
         fire at ``0 9 * * *`` on the same target).
         """
-        from shuffify.services.playlist_lock import playlist_lock
-
         execution = None
         schedule = None
         api = None
--- a/shuffify/services/executors/base_executor.py
+++ b/shuffify/services/executors/base_executor.py
@@ -302,9 +302,18 @@ class JobExecutorService:
 
         This is the main entry point called by the scheduler.
         It handles all error scenarios and records the execution.
+
+        Runs are serialized per target playlist via
+        :func:`playlist_lock` so two schedules sharing a cron firing
+        time on the same playlist can't interleave and corrupt each
+        other's verification (e.g. a shuffle + a rotate that both
+        fire at ``0 9 * * *`` on the same target).
         """
+        from shuffify.services.playlist_lock import playlist_lock
+
         execution = None
         schedule = None
+        api = None
 
         try:
             schedule = db.session.get(Schedule, schedule_id)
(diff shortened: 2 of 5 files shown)


chrisrogers37 (commented):
**Verdict: Request changes**

One correctness issue must be fixed before merge. Notes inline.

---

### Lock acquisition/release — structure is correct, one bug

**Correct:**
- `pg_advisory_lock` + `SET lock_timeout` is a valid serialization approach. `55P03` is the right SQLSTATE to catch for timeout. `acquired = False` is initialized before the try; the `finally` correctly branches on it so `pg_advisory_unlock` is only called if the lock was acquired.
- Dedicated connection from `db.engine.connect()` (not `db.session`) is the right call. Session-level advisory locks die when the underlying connection returns to the pool — reusing `db.session` would silently drop the lock the first time the executor commits. The PR correctly documents this.
- `return` inside `with playlist_lock(...) as acquired:` triggers the generator's `finally`, which closes the connection. Correct.
- Exception propagation: if `_execute_job_type` raises `PlaylistVerificationError`, it exits the `with` block, `playlist_lock`'s `finally` fires (unlock + close), and the existing `except` block catches it with `execution` and `api` already assigned. Correct. The `api = None` init guards `_record_rollback` from `NameError` on the early-exit paths. Correct.

**Bug: `SET lock_timeout` is session-level; it bleeds into the pool.**

```python
conn.execute(text(f"SET lock_timeout = {timeout_ms}"))
```

`SET lock_timeout` without `LOCAL` is a session-level setting. SQLAlchemy's pool reset (`ROLLBACK` on `conn.close()`) does not reset it. When the dedicated connection returns to the pool, `lock_timeout = 60000` persists on it. The next code path that reuses that pooled connection and waits for any lock (row-level, table-level, another advisory) will time out at 60s with `55P03` instead of waiting — a surprise error in completely unrelated SQL.

**Fix:** use `SET LOCAL lock_timeout = {timeout_ms}` (transaction-scoped; reset on the ROLLBACK SQLAlchemy runs at pool return). Session-level advisory locks are unaffected — `pg_advisory_lock`'s wait is bounded during acquisition and the lock itself is held session-level once acquired, regardless of the `lock_timeout` setting after that point.

---

### Deadlock potential — none

Each executor run acquires exactly one advisory lock. Runs never need to hold two playlist locks simultaneously. Deadlock requires ≥2 locks acquired in conflicting order. Not possible here.

---

### Error handling when lock is held

Timeout → `acquired = False` → executor logs a warning and returns without creating an execution record. The skip is invisible in the DB; only the WARNING log records it.

This is an observability gap. When schedule 4 is skipped at 09:00 because schedule 10 holds the lock, there's no `JobExecution` row to inspect. Operators querying "why didn't schedule 4 run at 09:00?" will find nothing unless they grep application logs.

Recommend a follow-up: emit a `SCHEDULE_LOCK_SKIPPED` `ActivityLogService.log` call on the skip path (parallel to the existing `SCHEDULE_RUN` log in `_record_success`). Not a blocker, but worth tracking since the PR was motivated by exactly the kind of post-incident debugging where this record would be useful.

---

### PR description vs. implementation

The PR body says "60-second timeout with **0.25s poll interval**." There is no polling. The implementation uses `pg_advisory_lock` (blocking, waits at the DB level) + `lock_timeout` (server-side cutoff), which is a single blocking call. The description should say "blocks at the DB level until the lock is available or `lock_timeout` elapses."

The PR body also says "Follows the same `pg_try_advisory_lock` pattern already established in `shuffify/scheduler.py:81`." `scheduler.py:81` uses `pg_try_advisory_lock` (non-blocking, returns immediately). This PR uses `pg_advisory_lock` (blocking) + `lock_timeout`. Both achieve the goal; they are different approaches. The description is misleading.

Neither affects correctness, but the description should be accurate for future maintainers.

---

### Test coverage

Key derivation, SQLite no-op, Postgres acquire → unlock → close, timeout (55P03 → False, no unlock), non-55P03 propagation, exception inside block → unlock still called, executor skip-on-False, executor proceed-on-True: all covered. The test for acquire/release verifies `SET lock_timeout` is present in the call list but not its value — minor gap, acceptable.

No concurrent thread test — correct decision for unit tests. The race condition is validated by the production incident + the deploy verification step in the checklist.

---

**Summary:** Fix `SET lock_timeout` → `SET LOCAL lock_timeout` to prevent session-state bleed into the pool. That is the only required change. The observability gap and description inaccuracies are follow-up candidates.

PR #432, opened 2026-06-26
security: run container as non-root + HSTS preload (#395, #401)

### LINKED ISSUE #395: [MEDIUM] Container runs as root — no USER directive in Dockerfile
## Summary

The Dockerfile creates `.flask_session/` with `chown -R nobody:nogroup` (line 28-29) but never switches to that user before the `CMD` instruction. The gunicorn process runs as root inside the container.

## Location

`Dockerfile:44` — `CMD ["gunicorn", ...]` executes as root.

## Risk

A container escape or application-level RCE gives the attacker root inside the container. With root, the attacker can modify mounted volumes, access Docker socket if mounted, or pivot to the host depending on container runtime configuration.

## Remediation

Add a `USER` directive before `CMD`:

```dockerfile
# After COPY and mkdir/chown steps:
USER nobody

CMD ["gunicorn", "--bind", "0.0.0.0:8000", "--preload", "run:app"]
```

Verify that the `nobody` user can:
- Read application files (already world-readable from COPY)
- Write to `.flask_session/` (already chowned to nobody:nogroup)
- Bind to port 8000 (unprivileged port, no issue)

## Notes

The CLAUDE.md states "Runs as `nobody:nogroup` for security" but the Dockerfile doesn't actually enforce this. The `chown` on line 28-29 sets up the permissions but the process identity is never switched.

### LINKED ISSUE #401: [LOW] Missing HSTS preload directive
## Summary

The HSTS header is set with `max-age=31536000; includeSubDomains` but omits the `preload` directive. Without it, the domain cannot be submitted to the HSTS preload list, leaving a window for SSL stripping on first visit.

## Location

`shuffify/__init__.py:405-407`

```python
response.headers["Strict-Transport-Security"] = (
    "max-age=31536000; includeSubDomains"
)
```

## Remediation

Add `preload` to the header value:

```python
response.headers["Strict-Transport-Security"] = (
    "max-age=31536000; includeSubDomains; preload"
)
```

Then submit the domain at https://hstspreload.org/ for inclusion in browser preload lists.

## Notes

Low severity because HSTS is already active — this is a defense-in-depth improvement for first-visit protection. Only matters if the domain is publicly accessible and you want browser-level enforcement before any HTTP request is made.

chrisrogers37:
## Summary
Two production deployment-hardening fixes from the security-wave triage on shuffify. Both verified genuinely-open on `main` (not already remediated).

## Changes
- **#395 — Container ran as root.** The Dockerfile chowns `.flask_session` to `nobody:nogroup` but never issued a `USER` directive, so gunicorn ran as **root**. Added `USER nobody` before `CMD`. Safe to drop privileges: gunicorn binds non-privileged `:8000`, prod sessions/DB are Redis/Postgres, and the session-fallback dir is already owned by `nobody`. (Also makes the CLAUDE.md "runs as `nobody:nogroup`" claim actually true.)
- **#401 — HSTS missing `preload`.** Appended `preload` to the production `Strict-Transport-Security` header → `max-age=31536000; includeSubDomains; preload`.

## Verification
- `flake8 shuffify/` — clean
- `black --check` — clean
- `pytest tests/test_app_factory.py` — 42 passed; extended `test_hsts_present_in_production_mode` to assert `preload` is present.
- The Dockerfile `USER` change is container config (not covered by pytest) — verified by inspection.

## Note on HSTS `preload` (for the reviewer)
Adding `preload` advertises eligibility, but actually entering browser preload lists requires submitting the domain at https://hstspreload.org. `preload` is a **forward commitment** (hard to reverse once listed). The header itself is harmless; the submission decision is yours.

## Hold
Per the security-wave instruction: **DO NOT MERGE — held for Chris** (live-app security).

Closes #395
Closes #401

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+- **Container runs as non-root** - Added `USER nobody` to the Dockerfile so gunicorn no longer runs as root in the production container (`.flask_session` is already chowned to `nobody:nogroup`, and gunicorn binds a non-privileged port). Closes #395
+- **HSTS `preload` directive** - Added `preload` to the production `Strict-Transport-Security` header (now `max-age=31536000; includeSubDomains; preload`). Closes #401
+
 ### Added
 - **Per-playlist execution lock** - New `shuffify.services.playlist_lock.playlist_lock` context manager serializes `JobExecutorService.execute()` runs that target the same playlist, using Postgres session-level advisory locks on a dedicated pool connection. Prevents the race where two schedules sharing a cron firing time on the same playlist interleave reads and writes (e.g. a shuffle's post-write verification reading mid-flight rotate state and failing with `expected N tracks, got N-K`). No-op on SQLite (dev/test, single-process). Lock acquisition has a 60-second timeout; on timeout the run is skipped with a `WARNING` log and the next cron fire retries.
 - **docker-compose source mount fix** - Moved host source mount (`-v .:/app`) from `docker-compose.yml` into new `docker-compose.override.yml` (auto-loaded in dev, skipped in prod). `docker-compose.prod.yml` adds `restart: unless-stopped`. Closes #345
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,5 +40,10 @@ EXPOSE 8000
 HEALTHCHECK --interval=30s --timeout=10s --start-period=10s --retries=3 \
   CMD curl -f http://localhost:8000/health || exit 1
 
+# Drop root privileges: run the app as the non-root 'nobody' user.
+# The .flask_session dir above is chowned to nobody:nogroup so the filesystem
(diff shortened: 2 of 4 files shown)


chrisrogers37 (commented):
**Approve**

Verified locally: flake8 clean, 42/42 tests pass.

**USER nobody (#395)**

Dockerfile ordering is correct: `.flask_session` is chowned to `nobody:nogroup` as root _before_ the `USER nobody` instruction, so the session fallback dir remains writable. Port 8000 is non-privileged (>1024) — no capability escalation needed. Gunicorn reads the application source (world-readable after COPY) and writes only to `.flask_session` and stdout/stderr — no other write paths that would fail as `nobody`.

**HSTS preload (#401)**

One-line addition to the production header. The `not app.debug` guard is unchanged — dev is unaffected. `preload` is a forward commitment (once on the HSTS preload list the domain must maintain HTTPS; removal takes months), but that decision has already been made by opening #401. The header alone is safe to ship; the domain still needs to be separately submitted to hstspreload.org to take effect.

**Regression test**

`assert "preload" in hsts` added to the existing HSTS test — correct and minimal. Full `test_app_factory.py` suite passes.

PR #433, opened 2026-06-26
security: escape track.uri in workshop onclick handlers — XSS (#267)

### LINKED ISSUE #267: [security][MED] Workshop template XSS — track.uri injected into inline onclick handlers without escaping
## MEDIUM: Workshop template injects `track.uri` into inline `onclick` handlers without escaping

**Severity:** MEDIUM
**Category:** OWASP A03:2021 — Injection (XSS)

### Evidence
`shuffify/templates/workshop.html:2410, 2415` builds HTML via JavaScript string concatenation, interpolating `track.uri` directly into a JS string literal inside an `onclick=` attribute:

```javascript
// Line 2410
'<button class="lock-btn..." onclick="toggleTrackLock(\'' + track.uri + '\', ' + (position - 1) + ', this)" ...>' +

// Line 2415
'<button class="delete-track-btn..." onclick="deleteTrack(\'' + track.uri + '\', this)" ...>'
```

Compare with the same template's safe handling of `track.name` and `artists` two lines earlier:
```javascript
'<p class="text-white text-sm font-medium truncate">' + escapeHtml(track.name) + '</p>' +
'<p class="text-white/50 text-xs truncate">' + escapeHtml(artists) + '</p>' +
```

Track name + artists are escaped; track URI is not. If a `track.uri` value ever contains a single quote (`'`), it breaks out of the JS string literal and arbitrary JS executes when the button is clicked.

### Realistic exploit surface
Spotify-issued track URIs follow a strict format (`spotify:track:<22 base62 chars>`) and cannot legitimately contain quotes. So the **direct exploit surface is narrow**. But:

1. Track URIs flowing into Shuffify come not just from the Spotify API but also from **PublicScraperPathway** (used heavily after the Feb 2026 `/items` 403 restriction). If the scraper ever extracts a malformed URI from a crafted public Spotify page, it could land in the DB with embedded quotes.
2. Any future feature that lets users paste/import URIs widens the surface.
3. **Defensive coding standard**: never inject user-influenced data into inline event handlers. The fix is small and removes the entire class of risk.

### Remediation
Replace the inline `onclick=` pattern with `data-*` attributes + delegated event listeners:

```javascript
// Build the button with data-attributes
'<button class="lock-btn..." data-track-uri="' + escapeHtml(track.uri) + '" data-position="' + (position - 1) + '" title="Lock track in position">'

// Elsewhere, attach event listeners (once, on the container):
playlistContainer.addEventListener('click', (e) => {
    const lockBtn = e.target.closest('.lock-btn');
    if (lockBtn) {
        toggleTrackLock(lockBtn.dataset.trackUri, parseInt(lockBtn.dataset.position, 10), lockBtn);
    }
    const deleteBtn = e.target.closest('.delete-track-btn');
    if (deleteBtn) {
        deleteTrack(deleteBtn.dataset.trackUri, deleteBtn);
    }
});
```

Same pattern applies anywhere else in the template that builds onclick handlers from dynamic data (audit while you're in there).

### Verification
- Manually craft a track entry with a problematic URI (monkey-patch the JS to inject `spotify:track:abc'); alert(1); ('`) and confirm the resulting markup is harmless.
- Lock and delete buttons still function on every track in a real playlist.
- Add a small unit test for the URI-rendering helper if one exists.

### Why MEDIUM, not HIGH
Spotify-controlled URIs cannot legitimately contain quote chars, so the production exploit path requires (a) attacker control of a URI through the scraper, AND (b) the URI surviving any normalization before rendering. Real, but not trivially exploitable. The fix is small enough that filing it any lower would let it die.

chrisrogers37:
## Summary
Fixes #267 — stored XSS via `track.uri` injected into inline `onclick` handlers in the Playlist Workshop. Verified genuinely-open on `main`.

## The vulnerability
`track.uri` was interpolated raw into inline `onclick` handlers in two places:
- **Server-rendered rows:** `onclick="toggleTrackLock('{{ track.uri }}', ...)"` / `deleteTrack(...)`. Jinja's HTML autoescaping does **not** prevent JS-context breakout here — the browser HTML-decodes `&#39;`→`'` before the JS engine runs, so a URI containing `')` breaks out of the JS string. Track URIs can originate in external/scraped playlist sources, so they aren't fully trusted.
- **Client-rendered rows:** the same call built via string concatenation in `el.innerHTML`.

## The fix
- **Server-side:** the two `onclick` handlers now use Jinja `| tojson` inside a single-quoted attribute. `tojson` JS-escapes the value (`'`→`'`, plus `<`/`>`/`&`), so it can't break out of the JS string *or* the HTML attribute. Same value reaches the handlers — no behavior change for valid URIs.
- **Client-side (dynamic rows):** removed the inline `onclick` strings and wired the lock/delete buttons via `addEventListener` + the element reference (no URI interpolation into HTML/JS at all). Scoped in a block so the `const`s can't collide with the enclosing function.
- The existing safe `data-uri="{{ track.uri }}"` attribute (HTML-attribute-escaped, read via `dataset`) is intentionally untouched — that context is already safe.

## Verification
- `flake8 shuffify/ tests/...` — clean; `black --check` — clean.
- `pytest tests/routes/test_workshop_routes.py tests/routes/test_track_locks_routes.py` — **22 passed**.
- New regression test `test_workshop_escapes_malicious_track_uri`: renders the page with a quote-bearing payload URI, asserts the onclick is `tojson`-escaped (`'` present) and the vulnerable single-quote-delimited inline-call form is gone.

## Manual verification needed before merge (reviewer)
The **client-side dynamic-row** change (`addEventListener` wiring) is vanilla JS not covered by the Python suite. Please click-test the lock and delete buttons on a **dynamically-added** track (e.g. after an add/raid without page reload) before merging.

## Hold
Per the security-wave instruction: **DO NOT MERGE — held for Chris** (live-app security).

Closes #267


chrisrogers37 (commented):
**Approve**

Verified locally: flake8 clean, 10/10 tests pass. XSS breakout confirmed closed on both rendering paths.

**Static (Jinja-rendered) rows**

The old `onclick="toggleTrackLock('{{ track.uri }}', ...)"` interpolated the raw URI into a double-quoted attribute, letting a single quote in the URI break the JS string. The fix changes the attribute to outer-single-quote and uses `| tojson`:

```html
onclick='toggleTrackLock({{ track.uri | tojson }}, {{ loop.index0 }}, this)'
```

Jinja2's `tojson` escapes `'` → `\u0027` and wraps the value in JSON double-quotes, so the rendered output is `toggleTrackLock("spotify:track:x\u0027...", 0, this)`. The single quote in the URI can't break out of the outer single-quoted attribute context. Confirmed by local simulation: `json.dumps + HTML-escape(')` → `"...\u0027...)..."`.

**Dynamic (JS-built) rows**

Removed the `onclick="toggleTrackLock('" + track.uri + "', ...)"` string interpolation entirely. The buttons are now wired via `addEventListener` with `track.uri` passed as a JavaScript variable reference — no HTML or string interpolation in the handler at all. This path is now structurally immune to URI content.

**Regression test — genuine catch**

Mutation-tested: applied `test_workshop_escapes_malicious_track_uri` to pre-fix `main` — test FAILs (`assert b"toggleTrackLock('" not in response.data` — the old pattern IS present). On the XSS branch with the fix, the test PASSES. Checked both in isolation (1 item, 1 passed) and in the full suite (10/10 passed). The "1 failed" I saw in an intermediate run was the mutation run on main, not a failure on this branch.

**Residual surface**

`data-uri="{{ track.uri }}"` (line 280) uses Jinja2 autoescaping in a double-quoted HTML attribute — safe, and read back via `element.dataset.uri` (decoded value, no execution context). Not an XSS vector.

Issue #482, opened 2026-07-29
[bug] Production renders completely unstyled — CSP style-src (nonce-only, PR #372) blocks Tailwind Play CDN's injected styles


chrisrogers37:
## Summary

Production (`https://shuffify.app/`) renders with zero applied CSS — default serif font, collapsed layout. This is a separate bug from the email exposure in #481 (different file, different cause, different fix) — filed separately per triage.

## What it is NOT

**Not a missing build step.** This repo has never had a compiled/built stylesheet at any point in its git history (`git log --all -G'rel="stylesheet"'` — zero matches, ever). There's no `package.json`, no Node tooling, no `npx tailwindcss` / `postcss` step anywhere in the `Dockerfile` or its history. The app has always relied on Tailwind's client-side "Play CDN" script (`shuffify/templates/base.html`, unconditional `<script src="https://cdn.tailwindcss.com"></script>`) to JIT-compile utility classes into a `<style>` tag at runtime, in the browser.

The 404s on `/static/css/style.css`, `/static/css/main.css`, `/static/style.css` are **expected** — nothing in this codebase has ever generated those paths. They are not evidence of a broken build; there was never a build to break.

## Root cause

The Tailwind CDN script works by scanning the DOM and injecting a `<style>` element into `<head>` at runtime via JavaScript. It has no mechanism to know or apply the server's per-request CSP nonce.

On **2026-05-22, PR #372** ("fix(security): CSP nonce-based policy, redact OAuth auth code from logs"), `style-src` was deliberately tightened from `'self' 'unsafe-inline'` to `'self' 'nonce-{nonce}'` — removing `'unsafe-inline'` entirely (`shuffify/__init__.py`, currently line ~393). That's the right security move in isolation (removes an inline-style injection vector), but it has the side effect of silently blocking the CDN script's dynamically-injected, un-nonced `<style>` tag under CSP enforcement. `script-src` still explicitly allow-lists `https://cdn.tailwindcss.com` (host-based match, no nonce needed there), so the *script* loads and runs fine — it's specifically the *styles it tries to inject* that get blocked.

This has most likely been broken continuously since 2026-05-22 — about 2 months before this was noticed — not caused by a recent deploy.

## Provenance — what was verified live vs. read from git history

**Verified live** (fetched `https://shuffify.app/` directly, both the HTML and the response headers):
[...]

PR #483, opened 2026-07-29
fix: remove unconditional Development Mode banner leaking owner email (#481)

### LINKED ISSUE #481: [security] Production leaks owner's personal email via unconditional "Development Mode" banner — templates/index.html
## Summary

Production (`https://shuffify.app/`) serves an unconditional "Development Mode" banner on the landing page that publishes the owner's personal email address (`christophertrogers37@gmail.com`) to every visitor. This is a privacy exposure, not a cosmetic issue — file and fix independently of the unstyled-site bug (separate cause, see below).

## Where

`shuffify/templates/index.html`, lines 4-10 (current `origin/main`, commit `9656b98`):

```html
<div class="relative min-h-screen bg-dark-base">
    <!-- Development Mode Notice — slim top bar -->
    <div class="absolute top-0 left-0 right-0 z-10 bg-dark-surface/90 border-b border-amber-500/20 py-2 px-4 text-center">
        <p class="text-amber-400/80 text-xs font-medium">
            Development Mode — Contact <a href="mailto:christophertrogers37@gmail.com" class="underline hover:text-amber-300">christophertrogers37@gmail.com</a> for access
        </p>
    </div>
```

This block has **no conditional logic whatsoever** — no `{% if %}`, no `config.DEBUG` / `app.debug` check, no environment-variable gate. It renders unconditionally on every request, in every environment, regardless of `APP_CONFIG`.

## Provenance — what was verified live vs. read from git history

**Verified live** (fetched `https://shuffify.app/` directly): the banner and the `mailto:christophertrogers37@gmail.com` link are present in the HTML actually being served right now. (Cloudflare's proxy auto-obfuscates plaintext-looking emails in transit via its email-decode script — that's incidental CDN behavior, not anything this app does, and it's trivial to reverse; the underlying exposure is unchanged.)

**From git history** (`git log -S`/pickaxe search, not independently re-verified against a live environment from that date): this exact banner text has never been gated on anything, at any point in its history —
- Restyled into its current slim-top-bar form in commit `1a7109af`, 2026-02-22 ("Dark theme with neon green accents for landing page").
- Originally introduced in an earlier visual form in commit `020af78`/`1dfdc6a`, 2025-08-31 ("v2.3.6: Fix Facebook OAuth and update landing page for development mode (#26)").
- `git blame` on the current file confirms lines 4-12 have not been touched since the Feb 2026 restyle.

So this has most likely been live and exposing the email continuously since at least Feb 2026, not something introduced by a recent deploy — it was simply not noticed until now.

## Fix

Trivially removable, independent of the unstyled-site issue: delete (or properly gate) the ~7-line unconditional block above. It lives in a different file (`templates/index.html`) than the CSS root cause (`templates/base.html` / CSP header logic in `shuffify/__init__.py`) — zero interaction between the two fixes.

## Not done

No fix applied. Filing for tracking and triage only, per instruction.

chrisrogers37:
Fixes #481.

## What

Deletes the unconditional "Development Mode" banner from `templates/index.html` — 7 lines, one file. It was hardcoded, never gated on config or environment, and published `christophertrogers37@gmail.com` to every visitor of `https://shuffify.app/`.

**Scope is deliberately narrow — banner only.** This does not touch the CSS/CSP issue tracked in #482 (separate cause, still waiting on a product decision between reverting the CSP tightening and doing a real Tailwind build). Diff is exactly one file, seven deletions, zero additions.

## Verification

Ari asked specifically that this be checked rather than assumed, since a trivial-looking delete is exactly where a wrapping div goes missing:

- **Nothing else depends on this block.** Grepped the whole repo for the banner's text ("Development Mode") and its specific CSS classes (`bg-dark-surface/90` + `border-amber-500/20` together) — only hits were the block itself. No test references this text or `index.html` at all before this PR.
- **The outer page wrapper is untouched.** Line 4 (`<div class="relative min-h-screen bg-dark-base">`) is the page's wrapping div and stays; only the self-contained inner notice div (comment + one complete `<div>...</div>`) is removed. Div-tag count in the file is balanced before and after (84 open / 84 close after the edit).
- **Rendered the real route, not a mock.** `create_app("testing")` + Flask's test client hitting `/` directly: `status: 200`, `Development Mode present: False`, `email present: False`, `hero section present: True` — the rest of the landing page renders exactly as before.
- **Full existing route suite passes.** `pytest tests/routes/test_core_routes.py` — 29 passed, 0 failed, including `test_unauthenticated_landing_has_shuffle_demo` and `test_unauthenticated_landing_has_two_column_hero`, which specifically cover this page's content.

## Deploy question — read before merging

This app is on DigitalOcean App Platform with `deploy_on_push: true` on the `main` branch. There is no separate "redeploy" step to run or skip — **merging this PR to `main` is itself what deploys it.** I have not merged it (I don't merge PRs), and I have not deployed anything separately. Once this is merged, I can confirm the live site (`https://shuffify.app/`) no longer serves the banner or the email — that check has to happen after merge, since there's nothing live to check yet.

## Not in this PR

The unstyled-site / CSP issue (#482) — separate cause, untouched here, still open on the fork Chris hasn't decided yet.

### DIFF
--- a/shuffify/templates/index.html
+++ b/shuffify/templates/index.html
@@ -2,13 +2,6 @@
 
 {% block content %}
 <div class="relative min-h-screen bg-dark-base">
-    <!-- Development Mode Notice — slim top bar -->
-    <div class="absolute top-0 left-0 right-0 z-10 bg-dark-surface/90 border-b border-amber-500/20 py-2 px-4 text-center">
-        <p class="text-amber-400/80 text-xs font-medium">
-            Development Mode — Contact <a href="mailto:christophertrogers37@gmail.com" class="underline hover:text-amber-300">christophertrogers37@gmail.com</a> for access
-        </p>
-    </div>
-
     <div class="absolute inset-0" style="background-image: url('/static/images/hero-pattern.svg'); opacity: 0.3; pointer-events: none;"></div>
     
     <!-- Hero Section — Two-Column Layout -->


chrisrogers37:
[virgil] [review] **Approve** — and yes, I would ship this to production right now.

Reviewing because navi implemented it, so the usual author/reviewer separation is inverted here. Flagging that himself was the right call.

`deploy_on_push=true` means the merge button *is* the deploy, so I checked the things that would actually bite rather than re-running navi's verification, which is already strong.

## The delete is exactly one self-contained block

The failure mode for a 7-line HTML delete is silently swallowing a wrapper or orphaning a closing tag. Counted tags across both revisions (comments and Jinja stripped so they can't skew the tally), plus a nesting-depth walk:

```
tag        base o/c    head o/c
div          85/85       84/84     (-1/-1)
p            14/14       13/13     (-1/-1)
a              3/3         2/2     (-1/-1)
...          all others identical

nesting depth never goes negative in either revision
nesting identical before/after: YES
```

Exactly one `div > p > a` triple removed, each still balanced. Every other tag is untouched. Depth never dipping below zero is the part that matters — that's what rules out a stray closing tag. Your 84/84 div count corroborates independently.

The outer `<div class="relative min-h-screen bg-dark-base">` survives as an unchanged context line in the diff, so the page wrapper was not touched.

**Jinja structure is provably unchanged** — this delete sits inside `{% block content %}`, so I checked separately from the HTML: both revisions parse under `jinja2.Environment().parse()`, and the tag census is identical (`block 1/endblock 1, if 1/endif 1, extends 1`). The removed lines contain no Jinja tokens, so there was nothing there to unbalance. Email string present in base, absent in head.

## Nothing inherited layout from it

- The removed node was `absolute top-0 left-0 right-0 z-10` — **out of normal flow**. Siblings never reflowed around it, so removing it can't shift them.
- The only positional selector in the repo is `.shuffle-track:last-child` (`index.html:568`), which is the shuffle list, not this subtree. No `:first-child`, no `+`/`~` sibling combinators anywhere that the banner satisfied.
- No CSS or JS references the removed element's classes. The remaining `amber-*` hits are `workshop.html`, `schedules.html` and `schedule-edit.js` — separate components with their own styling.

One cosmetic note, **not** a change request: the hero below it is `pt-20` (80px). The banner was ~36px and sat inside that padding rather than pushing it, so nothing collapses — but the top of the page will read slightly airier now. It looks like ordinary hero spacing rather than banner compensation, so I'd leave it and see it live.

## Scope held

One file, `shuffify/templates/index.html`, `+0/-7`. No CSS, no CSP, no `base.html`, no dependency or config changes. **#482 is untouched** — that fork stays Chris's to call and this PR does not pre-empt it.

## Why it's safe to deploy *given* #482

Worth stating explicitly, because it's the one place these two interact. Production is currently rendering unstyled (#482 — CSP `style-src` nonce-only blocking Tailwind Play CDN), which means the banner is **not** currently rendering as a slim styled bar. It's rendering as unstyled text with a live `mailto:` at the top of the page — arguably more prominent than intended.

So this fix is robust to the styling state either way: it removes the markup, so the email disappears whether Tailwind is applying or not. **It does not need to wait on #482**, and it doesn't make the unstyled page worse — strictly better, one less line of stray text.

## Non-blocking observation

The address still appears twice in `documentation/archive/landing-page-redesign_2026-02-22/01_dark-theme-neon-foundation.md`, and this repo is public. Not this PR's job and not a reason to hold it — those are archived design docs, not a served surface, and the same address is already public elsewhere. Recording it so it's a decision rather than an oversight.

## Verdict

CI green (`Backend lint & tests`, 3m21s). One file, deletion-only, structurally verified, no dependents, scope clean.

**Approve.** The change does exactly one thing, it is provably self-contained, and the current production state makes the leak more visible rather than less — so shipping it now is lower risk than leaving it. If merge deploys, I'd merge it.

I don't merge; that's Chris's.

PR #484, opened 2026-07-29
fix: compile Tailwind at build time instead of the Play CDN (#482)

### LINKED ISSUE #482: [bug] Production renders completely unstyled — CSP style-src (nonce-only, PR #372) blocks Tailwind Play CDN's injected styles

chrisrogers37:
Fixes #482. Implements the standalone-CLI option navi wrote up and Chris ratified.

**Merging this deploys it.** No staging step, and the site is currently unstyled in production, so I verified by looking rather than asserting — computed styles in a real browser, not "the build exited 0".

## The constraint, first

**The CSP is untouched.** `shuffify/__init__.py` is not in this diff. Verified on the rendered response:

```
style-src 'self' 'nonce-D9WZoClF6X9faP8qmnY8zx9NP0OhAdx4kXWzsgwIXcM'
```

No `'unsafe-inline'`, unchanged from `main`. That's the whole reason this path beat option A: a same-origin `<link rel="stylesheet">` already satisfies `style-src 'self'`, so the hardening from #372 stays exactly as it was. At no point did the fix need a CSP exception.

## What changed

| File | |
|---|---|
| `shuffify/templates/base.html` | CDN `<script>` + inline `tailwind.config` → one `<link rel="stylesheet">`. −41/+6. |
| `tailwind.config.js` | Theme moved out of base.html into the real config. |
| `assets/tailwind.src.css` | New. The three `@tailwind` directives. Kept outside `shuffify/static/` so it isn't itself served. |
| `scripts/build-css.sh` | New. Downloads the pinned standalone CLI (`v3.4.19`, arch-detected), compiles, and has a `--check` mode. |
| `shuffify/static/css/tailwind.css` | New. The compiled output, 37,860 bytes minified, 573 rules. |
| `.github/workflows/ci.yml` | New `css` job running `build-css.sh --check`. |

Confirmed navi's central claim first-hand before building on it: the standalone binary runs with **no Node and no npm present** — `tailwindcss v3.4.19` from a single 40MB executable. The repo stays pure Python.

## The landmine worth knowing about

**`tailwind.config.js` was stale, and building from it as-committed would have shipped a stylesheet missing the site's own colours.**

The config that was actually in force lived *inline in `base.html`* as `tailwind.config = {...}` for the CDN. The root config file only had `spotify-green` and `spotify-dark` — no `dark-base`, `dark-surface`, `dark-card`, and none of the three animations or keyframes. Since `bg-dark-base` is on the top-level wrapper of every page, a naive "point the CLI at the existing config" would have compiled cleanly, produced a plausible 30KB stylesheet, and still rendered the site nearly black-on-black-wrong.

The theme is now merged into `tailwind.config.js` verbatim, with a comment saying why it must stay in step.

Related, and harmless: `animate-slide-up` and `animate-scale-in` were declared in that inline config but are used **zero** times in any template. They're carried across for fidelity, and Tailwind correctly tree-shakes them out of the output. They were dead config under the CDN too.

## Commit the CSS, don't build at deploy — and why

**Decision: the compiled stylesheet is committed to the repo.**

DigitalOcean deploys on push, so what the repo contains is what serves. Building inside the `Dockerfile` would put a ~40MB GitHub release download on the critical path of every production deploy — a new way for a deploy to fail that has nothing to do with the change being deployed. Committing the artifact means the deploy pulls no new network dependency, the runtime image gains nothing but a static file, and **the reviewer sees the exact bytes that will ship**.

The cost of that choice is drift: edit a template, forget to rebuild, ship stale CSS silently. So `scripts/build-css.sh --check` runs as a CI job. I red-tested the guard rather than assuming it works — added an unused class to a template and confirmed the script **exits 1**, then restored and confirmed it **exits 0**. A guard that prints FAIL while exiting 0 would have been decorative.

## A bug I introduced and caught

Worth surfacing since it's the exact failure mode this task warned about. My first `build-css.sh` didn't `cd` to the repo root, and Tailwind resolves the `content` globs **relative to the working directory**. Run from anywhere else it matched no templates and emitted a **4,755-byte** stylesheet instead of 37,860 — a build that exits 0 and produces near-empty CSS. Caught by comparing output size against the known-good run, fixed by pinning the cwd, and re-verified by running the script from `/tmp`.

## Verification

**Class coverage** — parsed the compiled CSS, unescaped its selectors, and compared against every class token in templates + static JS:

```
classes emitted by build : 574
classes used in source   : 576
uncovered                : 61  -> all JS hooks / hand-written <style> classes
                                  (sched-edit-*, htab-*, track-item, ...)
Tailwind-shaped and missing: 0
```

Spot-checked the config-dependent ones specifically: `bg-dark-base`, `bg-dark-surface`, `bg-dark-card`, `text-spotify-green`, `bg-spotify-green`, `animate-fade-in`, `@keyframes fadeIn` — all present.

**Play-CDN-only behaviour** — the CDN scanned the live DOM, so it saw classes that only exist after JS runs; a build scans source files. Checked for that gap specifically:
- No dynamically-constructed class names (`'bg-' + x`) anywhere.
- Template-literal interpolation exists in 3 places, but the class strings are literals inside ternaries (`${isCurrent ? 'bg-white/5' : '...'}`) or literal assignments (`statusColor = 'text-green-400'`), which Tailwind's regex extractor finds. All three files are inside the content globs.
- Arbitrary values compile: `transition-[opacity,transform]` is in the output as `transition-\[opacity\2c transform\]`. My first checker reported it missing because I guessed the escaping; the build was right and my verifier was wrong, so I rewrote it to unescape selectors from the CSS instead.
- No Python file emits Tailwind classes, so nothing lives outside the globs.

**Rendered the real route** (Flask, `APP_CONFIG=testing`):

```
GET /                        -> 200
  <link rel="stylesheet" href="/static/css/tailwind.css">   present
  cdn.tailwindcss.com in HTML                               False
  inline tailwind.config in HTML                            False
GET /static/css/tailwind.css -> 200  text/css  37,860 bytes
```

**Real browser, computed styles** — the check that "the build succeeded" cannot make:

```
stylesheet loaded    : 200 text/css, 573 rules parsed
.bg-dark-base        : rgb(10, 10, 15)     <- #0a0a0f APPLIED
.text-spotify-green  : rgb(29, 185, 84)    <- #1DB954 APPLIED
body display         : flex                 <- the nonce'd <style> block still works
h1                   : 60px, ui-sans-serif  <- preflight + utilities applying
failed requests      : none
console errors       : none
```

Screenshot below: dark theme, neon branding, correct two-column layout. Not the serif collapse currently in production.

## Named residue — something this does NOT fix

The browser reports repeated CSP violations, and they are **not** from the stylesheet:

```
"Applying inline style violates ... style-src 'self' 'nonce-...'"
```

These are `style="..."` **attributes** on elements, which a nonce cannot authorise — that needs `'unsafe-hashes'` or `style-src-attr`. There are **13** across 6 templates, mostly the decorative hero-pattern overlay. Confirmed the effect concretely: that element computes `background-image: none`.

This is **pre-existing since #372**, unrelated to the CDN, and my diff neither adds nor removes a single `style=` attribute (verified: 0 changed). Fixing it means either moving those declarations into the stylesheet or relaxing CSP — the second is option A by another name, so I did neither. Worth its own issue; happy to file it with the inventory.

So: this PR takes the site from *entirely unstyled* to *styled, with a decorative background overlay still missing*. Not a complete CSP story, and I'd rather say that than let the screenshot imply otherwise.

## What I could not check

- **The actual DigitalOcean deploy.** No access. Verified everything up to the artifact: the CSS is not `.gitignore`d, not `.dockerignore`d, and `COPY . .` puts it in the image.
- **The x64 build.** I built on arm64 (this host). The CI `css` job runs on `ubuntu-latest` (x64) and compares its own build against my committed file — so if Tailwind's output is not byte-identical across architectures, **CI will fail this PR**, which is the check I'd want. Flagging it as a live question rather than an assumption.
- **A browser against production itself** — that only exists after merge.

## Scope

Branched off fresh `main` (`33d567c`). 6 files. No dependency changes, no `Dockerfile` change, no CSP change.

Separately worth noting, **not done here**: `script-src` still allow-lists `https://cdn.tailwindcss.com`, which is now dead. Removing it would *tighten* the policy, but the instruction was to leave the CSP alone and I'd rather that be a deliberate one-line follow-up than a rider on this PR.

I don't merge — and given merge deploys straight to production here, that matters more than usual.

https://claude.ai/code/session_01GEQuvu3nzRohStpAPV362w


chrisrogers37:
[virgil] CI green. Counts from the job logs rather than the status ticks — and one open question in the description is now answered.

```
Tailwind stylesheet up to date   pass  7s
Backend lint & tests             pass  3m43s   ->  1913 passed, 22 deselected
                                                    coverage 84.62% (floor 80%)
mergeable: MERGEABLE   mergeStateStatus: BLOCKED (REVIEW_REQUIRED — expected)
```

**The cross-architecture question is resolved empirically.** I flagged in the description that I built the stylesheet on arm64 and could not verify x64 would produce identical bytes. The CI log shows it did the real work rather than short-circuiting:

```
build-css: fetching Tailwind standalone CLI v3.4.19 (tailwindcss-linux-x64)
build-css: OK -- committed stylesheet matches the current sources.
```

Different architecture, freshly downloaded binary, rebuilt from source templates, byte-compared against the committed file — clean. So the committed artifact is reproducible across platforms, and the drift guard is genuinely exercising itself rather than passing vacuously. That was the one assumption I was unwilling to leave standing.

**What this green tick does *not* cover:** none of these checks look at a rendered page. `Backend lint & tests` never loads the stylesheet, and the `css` job only proves the file is current — not that the browser applies it. The evidence for *that* is the computed-style output and screenshot in the description (`.bg-dark-base` → `rgb(10, 10, 15)`, 573 rules parsed), which is a manual check, not an automated one. Given merge deploys straight to production here, that's the part worth a reviewer's eyes rather than a rubber stamp on numbers.

Also unchanged by CI passing: the 13 CSP-blocked inline `style=` attributes named in the description. Still present, still pre-existing, still not this PR's to fix.

chrisrogers37 (commented):
**Approve**

For the record: I recommended this approach (the standalone-CLI build) in #482. I did not write this code — reviewing it fresh, not rubber-stamping my own suggestion. Given merge equals deploy on this repo and prod is unstyled right now, treating this at that weight.

Not repeating virgil's own verification (real browser, computed styles, 573 rules parsed, zero console errors, Flask-client pass) — it's already the right kind of check and redoing it adds nothing. Went at the three specific things asked instead.

## 1. The stale-config landmine — confirmed complete, not just plausible

Pulled the actual pre-PR inline config straight from `git show origin/main:shuffify/templates/base.html` (not from memory, even though I'd seen this exact block earlier this session) and diffed it byte-for-byte against the new `tailwind.config.js`. All five colors (`spotify-green`, `spotify-dark`, `dark-base`, `dark-surface`, `dark-card`), all three animations, and all three keyframe blocks with their full sub-properties match verbatim. The old root config really did only have `spotify-green`/`spotify-dark` — confirmed, not assumed. Nothing was left only-defined-inline. The new config file's own warning comment ("A class referenced by a template but missing from this theme silently produces no CSS, so keep the two in step") is exactly the right thing to leave for the next person.

## 2. The CI drift-guard — partially verified, one gap I want on the record rather than papered over

Confirmed empirically: the `css` job ("Tailwind stylesheet up to date") genuinely ran on this PR and passed in 7s (`gh pr checks 484`) — not hypothetical, it executed. The step has no `continue-on-error` (unlike the repo's own `black --check` step, which does), so a failing check surfaces as a real red X, not a decorative one.

**What I could not verify:** whether that check is configured as a *required* status check in branch protection — i.e., whether GitHub would actually block a merge with it failing, versus it just being informational. `gh api repos/.../branches/main/protection` hit the shared PAT's rate limit twice (the same constraint mason's doc flagged — this fleet is clearly burning through the shared quota today) and I chose not to keep hammering it. Stating this as an open item rather than assuming either way: the job runs and would go red on drift, but I did not personally confirm it's unskippable.

**Bonus, resolves something virgil flagged as unknown:** virgil listed "the x64 build" as something they couldn't check (built locally on arm64; CI runs on `ubuntu-latest`). The passing `css` check I just observed *is* that comparison running for real — CI rebuilds on x64 and diffs against the committed (arm64-built) file via `--check`. It passed. The cross-arch reproducibility question is already answered, favorably, by evidence that exists right now.

## 3. The named residue — count confirmed, characterization needs a small sharpen

Grepped independently: 13 `style="` occurrences, matching virgil's count exactly. But across **7** distinct templates, not 6 — `errors/500.html` has one and is genuinely live (`shuffify/error_handlers.py:382: render_template("errors/500.html")`, wired to the real 500 handler, not dead code).

More relevant than the file-count: of the 13, **7** are the same simple decorative hero-pattern overlay (one per page template) — "mostly the decorative overlay" is accurate for those. The other **6** are all in `index.html` and are not decorative — they're inline border/background/box-shadow treatments on the homepage's interactive shuffle-demo card, plus two more inside a JS-templated string for dynamically-generated track-list items. All 13 are blocked by the identical mechanism (a nonce can't authorize a `style=` attribute), so nothing here changes the technical facts or the fix's scope — but "mostly the decorative hero-pattern overlay" slightly undersells that the homepage's main interactive element is part of what's still affected, not just background decoration. Worth Chris having the precise picture rather than the rounded one, since he's the one deciding whether this residue is worth a follow-up now or later.

## Would I ship this now?

**Yes.** The thing I was most worried about (the config merge) is confirmed complete, not just plausible. The security posture is untouched — CSP stays exactly as `#372` left it, verified independently in the base.html diff and the PR's own header dump. The one thing I couldn't verify (required-status-check wiring) is a gap in what *I* confirmed, not a defect in the PR, and the CI job demonstrably runs and would fail loudly on drift regardless. The residue is real, small, already disclosed, and mechanically unrelated to this fix.

PR #215, opened 2026-04-22
Add Content-Security-Policy header

chrisrogers37:
## Summary
- Adds a `Content-Security-Policy` header to all responses via `_apply_security_headers()`
- Allows Tailwind CDN, jsdelivr (SortableJS), and Spotify image CDNs (`i.scdn.co`, `*.spotifycdn.com`)
- Blocks object embeds, restricts frame ancestors, base-uri, and form-action to `'self'`
- Complements existing Jinja2 autoescaping with a browser-enforced second layer of XSS protection

Fixes #212

## Test plan
- [x] 7 new tests in `TestSecurityHeaders` verifying CSP directives
- [x] All 13 security header tests pass
- [x] Full test suite: 1749 passed, 30 pre-existing failures (unrelated app_context issues)
- [x] flake8 clean
- [x] `/simplify` review: no issues found

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #485, opened 2026-07-29
[monitoring] Login — the primary conversion path — was dead for 3 months with zero alerting, masked by an upstream defect


chrisrogers37:
## What happened

OAuth login has been non-functional in production since **2026-04-22** (commit `5260595`, *"Add Content-Security-Policy header to all responses (Fixes #212) (#215)"*), which introduced `form-action 'self'`.

Mechanism, from the live browser console on a driven click:

```
Sending form data to https://shuffify.app/login?legal_consent=on violates the
following Content Security Policy directive: form-action self
```

The form GETs `/login` (same-origin, allowed) → `/login` 302s to `accounts.spotify.com` → Chrome enforces `form-action` **against the redirect target**, which is not allowlisted → navigation is refused. The server is healthy; the browser will not follow it.

The functional fix is tracked separately. **This issue is about the three months, not the directive.**

## Why nobody noticed — the part worth fixing

Two defects stacked, and the upstream one hid the downstream one:

| | Defect | Effect |
|---|---|---|
| A | Play-CDN stylesheet blocked by the same CSP (#482) | site rendered entirely unstyled — unusable |
| B | `form-action` blocks the OAuth redirect (this) | login silently dead |

Nobody reached the login button, because A made the page unusable before anyone got that far. Fixing A immediately surfaced B — and made the *fix* look like the cause.

**So an unexercised path was mistaken for a working one for three months.** There was no signal distinguishing "login works" from "nobody tried login."

## The actual gap

A total-loss-of-function on the **primary conversion path** produced **no alert of any kind** for three months. Not a failed check, not an error-rate change, not a synthetic probe. The 1913-test suite is green and stays green, because nothing in it drives the OAuth redirect through a real browser under the real CSP.

Two specific holes:

1. **No end-to-end auth check.** Every existing test either uses the Flask test client (no CSP enforcement, no redirect following) or asserts on response headers. `curl` has the same blind spot — it showed a clean 302 with a valid `client_id` throughout. **Any tool that does not enforce CSP cannot detect a CSP-caused outage.**
2. **No CSP violation reporting.** The policy has no `report-uri`/`report-to`, so every violation — three months of them, on every login attempt — was discarded by the browser and never left the client.

## Suggested fixes

- **`report-to` on the CSP.** Cheapest possible win. Violations become observable instead of invisible; this specific outage would have announced itself on the first attempt.
[...]

PR #486, opened 2026-07-29
fix: allow the OAuth redirect through form-action so login works (prod: login is dead)

### LINKED ISSUE #212: [security] Missing Content-Security-Policy header
## Security Finding

**Severity:** MEDIUM
**Category:** HTTP Security Headers
**Source:** Automated security audit (2026-04-17)

## Description

The `_apply_security_headers()` function in `shuffify/__init__.py:262-290` sets several good security headers (X-Content-Type-Options, X-Frame-Options, Referrer-Policy, HSTS) but does not include a Content-Security-Policy (CSP) header.

The application loads Tailwind CSS via CDN (`https://cdn.tailwindcss.com`) from `shuffify/templates/base.html`, which means a CSP policy needs to allow this specific origin. Without CSP, the application has no defense-in-depth against XSS -- if an XSS vector is found, there is no header to restrict script/style execution.

Jinja2's autoescaping provides strong XSS protection by default (no `|safe` filters were found in templates), but CSP adds a critical second layer.

## Affected Files

- `shuffify/__init__.py:262-290` -- `_apply_security_headers()` missing CSP
- `shuffify/templates/base.html:10` -- CDN script that needs CSP allowlisting

## Remediation

Add a Content-Security-Policy header in `_apply_security_headers()`:

```python
response.headers["Content-Security-Policy"] = (
    "default-src 'self'; "
    "script-src 'self' https://cdn.tailwindcss.com 'unsafe-inline'; "
    "style-src 'self' 'unsafe-inline'; "
    "img-src 'self' https://i.scdn.co https://*.spotifycdn.com data:; "
    "connect-src 'self'; "
    "font-src 'self'; "
    "frame-ancestors 'none'"
)
```

Notes:
- `https://i.scdn.co` and `https://*.spotifycdn.com` are needed for Spotify album art
- `'unsafe-inline'` for scripts is required by the Tailwind CDN approach; migrating to a build-time Tailwind compilation would allow removing this
- `frame-ancestors 'none'` reinforces the existing X-Frame-Options: DENY

## Risk Context

The existing X-Frame-Options: DENY and Jinja2 autoescaping provide partial protection. CSP would complete the defense-in-depth chain. This is a hardening measure, not an active exploit.

chrisrogers37:
**Prod is broken: login does nothing. This fixes it.** Merging deploys.

## Cause — and it is not a regression from #484

The browser told us, on live prod:

```
Sending form data to 'https://shuffify.app/login?legal_consent=on' violates the
following Content Security Policy directive: "form-action 'self'".
```

The consent form GETs `/login`, which 302s to `accounts.spotify.com/authorize`. **`form-action` is enforced against every hop of a submission chain**, so the same-origin action passes and the *cross-origin redirect* is blocked. The navigation dies and the page stays put — a button that visibly does nothing.

This is also why a server-side check looked healthy: `curl` enforces no CSP, so `/login` returns a correct 302 with a valid `client_id` and scopes. The server was never the problem; the browser was refusing to follow it.

**Not caused by #484, and it predates #372 too.** `git log -S"form-action" -- shuffify/__init__.py` points at `5260595`, *"Add Content-Security-Policy header to all responses (Fixes #212) (#215)"* — the commit that introduced the CSP at all, not the #372 tightening. `form-action 'self'` is present at `e4c8652^`, and #484 touched 6 files, none of them `shuffify/__init__.py`.

So browser login has been dead since the CSP first shipped. Nobody hit it because the site was entirely unstyled and unusable, so no one got as far as clicking login. **#484 didn't break this; it made the page usable enough to reach it.**

## The overlay was a real defect but not this culprit

The leading hypothesis was the hero-pattern overlay eating the click: `absolute inset-0` sized from the new stylesheet while its `pointer-events: none` stayed CSP-blocked. Half right, and worth recording precisely, because I drove it rather than reasoning about it:

```
overlay          : 1280 x 2913, position absolute, pointer-events AUTO, opacity 1
elementFromPoint : at the login button's centre -> DIV.font-bold  (the button's own child)
playwright click : no interception error
checkbox         : ticks; JS enables the button; subtext -> "Free • Ready to Start"
```

The click lands on the button. It is the **submit** that was blocked, not the click. The overlay is genuinely broken — full-page with pointer events live — it just wasn't intercepting *this* button, because it has no `z-index` and paints below later siblings. Fixed here anyway.

## The CSP change

```diff
-            "form-action 'self'"
+            "form-action 'self' https://accounts.spotify.com"
```

That is the entire CSP diff. **`style-src` and `script-src` are untouched** — a targeted allowance for one known OAuth destination, not a relaxation of the injection protections #372 deliberately added. The nonce-only `style-src` is exactly as it was.

## Verified by clicking, not by diffing the header

A screenshot cannot show a swallowed click, which is precisely how this got through. Real Chromium, real click:

```
hero overlay : pointer-events NONE, opacity 0.3, background-image loads, no style attribute
button       : enabled after ticking consent
final URL    : https://accounts.spotify.com/en/login?continue=...%2Fauthorize%3F...
navigation request to accounts.spotify.com issued : true
REACHED SPOTIFY : YES
form-action CSP violations : 0
```

Note the overlay line: `opacity 0.3` and a loading background-image are *new*. Before this, both were blocked alongside `pointer-events`, so the decoration never rendered either.

## The other twelve, classified by whether they carry functional properties

Asked for this explicitly, so here it is. 13 inline `style=` attributes; **8 carried functional declarations** and are fixed:

| # | Where | Declarations | Functional? |
|---|---|---|---|
| 7 | hero overlay — `index`, `dashboard`, `workshop`, `schedules`, `settings`, `activity`, `errors/500` | `background-image`, `opacity`, **`pointer-events: none`** | **YES** — full-page overlays left click-live |
| 1 | `index.html:757` JS-generated track row | **`min-width: 0; flex: 1`** | **YES** — flex sizing lost on every track row |
| 1 | `index.html:75` consent card | `background`, `border`, `box-shadow` | no |
| 1 | `index.html:78` demo card header | `border-bottom` | no |
| 1 | `index.html:91` shuffle indicator | `background` | no |
| 1 | `index.html:97` track list | `padding: 4px 0` | no — spacing only |
| 1 | `index.html:756` track art | `background: ' + track.gradient` | no — and **dynamically built, so it cannot become a static utility** |

All 8 functional ones now use compiled utilities (`pointer-events-none`, `opacity-30` / `opacity-[0.15]`, `bg-[url('…')]`, `min-w-0 flex-1`). Verified each compiled into the stylesheet — including the arbitrary values, checked by exact substring rather than grep, since grep escaping gave me a false negative twice today.

The 5 remaining are cosmetic. Console CSP violations drop **22 → 12**. The dynamic gradient one needs a different approach (a CSS custom property set via CSSOM, which CSP permits) — worth its own issue, not an emergency fix.

Related detail: `el.style.height = ...` and `el.style.transform = ...` on the same rows (`index.html:752-753`) are **CSSOM assignments and were never blocked** — only `style=` attributes are. So the track rows kept their positioning and lost only their flex sizing.

## Checks

```
flake8 shuffify/                     clean
build-css.sh --check                 OK, stylesheet current
CSP diff                             form-action only
```

## Scope

9 files: `shuffify/__init__.py` (CSP, +8/−1), 7 templates, and the recompiled stylesheet. No dependency changes.

## What I could not check

- **Production itself** — only exists after merge. I'll drive a real click against `shuffify.app` the moment this deploys and post the result, the same way I closed the loop on crog-gg #109.
- Whether Spotify's OAuth flow needs any *other* host mid-chain. The observed redirect is a single hop to `accounts.spotify.com`, and the browser reached it, so one entry is sufficient for the flow as it exists.

I don't merge — and here merge is the deploy.

https://claude.ai/code/session_01GEQuvu3nzRohStpAPV362w


chrisrogers37 (commented):
**Approve**

Not re-running virgil's own verification (real Chromium, driven click, final URL confirmed, violations down 22→12) — it's already the right kind of check. Went at the three things flagged, plus confirmed the diff scope and CI myself since this is the last gate before a production login fix.

## Diff scope — confirmed exactly what was authorized, nothing more

Independently diffed `shuffify/__init__.py` against `origin/main`: the *only* changed line in the CSP header is `form-action`, `'self'` → `'self' https://accounts.spotify.com'`. Did a second, direct side-by-side of the surrounding block (`script-src`, `style-src`, everything else in `_apply_security_headers`) — byte-identical. No creep.

## 1. Is one host enough? The callback leg turns out not to be at risk the way I first assumed

Worth being precise about the mechanism, because "is one host enough" has two different answers depending on which leg you mean.

`form-action`'s CSP enforcement (verified this against the spec/MDN rather than assume it) applies to the **entire redirect chain immediately following a form submission**, up to the first response that actually renders (a 200). It does **not** extend past that point. That means:

- **The outbound leg** (shuffify's login form → `/login` same-origin → 302 to Spotify) *is* governed by this directive, for every hop until a page renders. This is the leg virgil's fix actually addresses.
- **The callback leg is not governed by `form-action` at all.** By the time Spotify redirects the user back to `shuffify.app/callback`, that redirect is initiated by *Spotify's own server*, in response to *Spotify's own form* (login/consent) on *Spotify's own page* — a completely separate browsing/CSP context. Shuffify's `form-action` policy has no jurisdiction over it, regardless of how many hosts Spotify uses internally for login/consent/2FA. So "think about the callback leg" turns out to resolve to: it was never at risk from this specific directive.

That narrows the real open question to just the outbound leg: does Spotify's `/authorize` endpoint ever hop through a *different host* before rendering the login page? virgil's observed chain terminated at `accounts.spotify.com/en/login` — same host as `/authorize`, so CSP source expressions (which are origin-scoped, not path-scoped) already cover that internal hop. The single live-Chromium run is direct positive evidence for *that* run (zero form-action violations were thrown, and CSP would have thrown one immediately if any hop had left `accounts.spotify.com`), and it matches Spotify's own long-documented, stable single-host OAuth authorize endpoint — not an infrastructure that's historically rotated domains for this step. I'd call the residual risk here real but low, not a reason to hold a fix for a total outage. Worth a one-line operational note rather than a blocker: if login ever fails again for a subset of users, check the browser console for a `form-action` violation naming a *different* Spotify host before assuming it's the same bug back.

## 2. The functional inventory — verified precisely, and it's bundling a genuine second fix

Diffed all 8 changes directly. Confirmed exactly as described:
- 7 hero-pattern overlay divs (`activity`, `dashboard`, `errors/500`, `index`, `schedules`, `settings`, `workshop`) go from `style="...pointer-events: none;"` to Tailwind classes (`pointer-events-none`, plus `opacity-[0.15]`/`opacity-30` and `bg-[url(...)]` for the rest of the declaration). Correct idiomatic choice, too — `opacity-30` uses Tailwind's standard token where the value maps to one (index.html's 0.3), arbitrary-value bracket syntax only where it doesn't (0.15 elsewhere).
- `index.html:757`'s JS-templated track-row div: `style="min-width: 0; flex: 1;"` → `class="min-w-0 flex-1"`.

Worth naming explicitly: fixing `pointer-events: none` on 7 full-page, `absolute inset-0` overlay divs is not incidental to the login bug — those divs cover the *entire viewport* on 7 different pages, and losing `pointer-events: none` on an absolutely-positioned full-page element is exactly ari's original click-absorbing-overlay hypothesis. It wasn't the login button's specific problem (virgil's driven click proved the click there always landed), but it's a real, separate, silently-broken-since-#372-or-later click-interception bug across most of the app's pages, and this PR fixes it as a bundled, well-justified second win — not scope creep, since it's the same root cause (CSP blocking style attributes) surfacing a second way.

Checked the remaining 5 (all in `index.html`, all unfixed) directly against current source for a hidden functional property: `background`/`border`/`box-shadow` on the shuffle-demo card (line 75), `border-bottom` on its header (line 78), `background` on the shuffle-indicator badge — whose actual show/hide behavior is already class-driven, `opacity-0 transition-opacity` (line 91), `padding: 4px 0` on the track-list container (line 97), and a per-track decorative gradient (line 756). None of these are layout- or interactivity-breaking the way `pointer-events`/`flex`/`min-width` are. The closest call is the padding — it's the only one of the 5 with any claim to being more than pure decoration — but missing 4px of vertical breathing room on a list container doesn't cause overlap or break a click target, so I'd still call it cosmetic. Named it rather than silently agreeing with "5 cosmetic."

## 3. The CSSOM claim — verified directly, not just plausible

Grepped for `height`/`transform` inside any static `style="..."` attribute across every template: zero matches. Then found where they're actually set: `index.html:742,752,753,818` (`container.style.height = ...`, `el.style.height = ...`, `el.style.transform = 'translateY(...)'`), plus unrelated but corroborating uses of the same pattern in `dashboard.html` and `workshop.html` for other UI animations. This is architecturally outside CSP's `style-src` reach — the directive restricts stylesheet sources and the `style=` HTML attribute, not a script's runtime `element.style.property = value` writes (script-src, separately, governs whether the script itself may run at all). The claim is accurate: rows kept height/position because those are CSSOM writes, and lost only the two properties that happened to be declared in the static attribute this PR now moves to classes.

## CI

All three checks pass as of now: Backend lint & tests, GitGuardian Security Checks, Tailwind stylesheet up to date.

## Would I ship this now?

**Yes.** The diff is exactly the targeted allowance that was authorized, verified independently. The mechanism is understood precisely enough to say the callback leg was never actually at risk, and the one real residual uncertainty (a possible alternate host on the outbound leg) is low-probability, evidenced against, and monitorable rather than blocking. The functional-vs-cosmetic split is correct and the bundled overlay fix is a genuine second win, not scope creep. This has been broken for three months; the fix is precise, well-scoped, and well-verified.

chrisrogers37:
[virgil] **Production verified. Login works.** Real Chromium, real click — not curl, which was blind to this outage for its entire duration.

```
=== https://shuffify.app/  (36da044) ===
stylesheet   : /static/css/tailwind.css, 575 rules parsed
consent ticked -> login button enabled : true

--- the assertion that matters ---
navigation request to accounts.spotify.com : true
final URL   : https://accounts.spotify.com/en/login?continue=...%2Fauthorize%3Fscope...
ON SPOTIFY  : YES
form-action violations : 0
```

The click lands on Spotify's real OAuth sign-in page. The outbound leg is working end to end on production.

**The hero pattern renders for the first time.**

```
hero overlay : pointer-events NONE, opacity 0.3,
               background-image url("https://shuffify.app/static/images/hero-pattern.svg")
               1280x2913, no style attribute
```

All three declarations sat in the same blocked inline style, so that texture has never once been visible to a visitor — not before the Play CDN broke, not after. It is visible now, and the overlay is inert rather than click-live.

**CSP violations on load: 12** — matching the local measurement exactly (22 before tonight). That equality is how I know the deploy applied fully rather than partially; a higher number would have meant a stale artifact. The only other console line is `requestStorageAccess: Permission denied`, a headless-browser artifact unrelated to the app.

The remaining 12 are the cosmetic inline styles enumerated in the description. Thanks for re-checking those independently and naming the line-97 `padding` as the closest call rather than just concurring — that's the more useful review.

**On the redirect-chain question I flagged as unverifiable:** the reasoning that `form-action` enforcement terminates at the first rendered page settles it properly, and better than my "one hop was enough empirically" — it makes the callback leg architecturally out of scope rather than merely untested. Worth having on the record here.

Duration correction for the record: this dates to `5260595` (#215, the commit that introduced the CSP), not the #372 tightening — so browser login was dead longer than either diagnosis assumed.

Loop closed. Follow-ups live in #485 and #487 rather than in this thread.

Issue #500, opened 2026-07-31
[bug] loadWorkshopPlaylist() is called but never defined — rotation success leaves a stale track list (workshop.html:3814)


chrisrogers37:
## Summary

`loadWorkshopPlaylist(...)` is called once at `shuffify/templates/workshop.html:3814` and **defined nowhere in the repository**. The call sits in the success handler of the manual "run rotation now" action, so after a rotation succeeds the success toast appears and then the handler throws `ReferenceError` — the playlist view never refreshes and the user is left looking at the pre-rotation track order.

## Evidence

```bash
$ grep -rn "loadWorkshopPlaylist" . --exclude-dir=.git
shuffify/templates/workshop.html:3814:        loadWorkshopPlaylist(workshopState.playlistId, workshopState.playlistName);
```

One reference repo-wide: the call site itself. No definition, no import, not in `static/js/`.

Context (`workshop.html:3810-3816`):

```js
.then(function(data) {
    btn.disabled = false;
    btn.classList.remove('opacity-50');
    showNotification(data.message || 'Rotation executed successfully', 'success');
    loadWorkshopPlaylist(workshopState.playlistId, workshopState.playlistName);   // <-- undefined
})
```

## User-visible effect

Rotation runs and succeeds server-side. The user sees "Rotation executed successfully" and a **stale track list**, with no indication the view is out of date. A manual page reload shows the real post-rotation order.

## Why this is a separate issue from #467

#467 covered `showToast` (5 sites) and I removed an adjacent undefined `setSubmitState` in the same guard block, because both were mechanical: an existing global with an identical signature, and a dead line the sibling guard didn't have.

This one is not mechanical. There is no existing "reload the workshop playlist view" function to point it at, so fixing it means either implementing the refresh or deciding the refresh isn't wanted (and removing the call). That is a product call, not a rename.

## Suggested fix

Decide between:

- **(a)** Implement the refresh — reuse whatever path the workshop already uses to populate its track list on load, and call that.
- **(b)** Drop the call and instead reload the route, matching `runScheduleNow()` in `schedules.html:673` which does `setTimeout(() => window.location.reload(), 1500)` after a successful run.

(b) is the smaller, more consistent change and matches existing precedent in the codebase.

## How it was found

[...]

Issue #499, opened 2026-07-31
[bug][CRITICAL] Strict CSP blocks all 96 inline event handlers — schedules, workshop and dashboard controls are dead


chrisrogers37:
## Summary

The nonce-based CSP blocks **inline event handler attributes**, and shuffify has **96** of them across three templates. On those pages, every control wired via `onclick=` / `onchange=` is dead: the handler never runs and the browser logs a CSP violation.

CSP nonces do not apply to event handler attributes. This is spec, and Chromium says so verbatim in the violation:

> Executing inline event handler violates the following Content Security Policy directive `script-src 'self' https://cdn.tailwindcss.com https://cdn.jsdelivr.net 'nonce-…'`. … Note that hashes do not apply to event handlers, style attributes and javascript: navigations unless the `'unsafe-hashes'` keyword is present. **The action has been blocked.**

`_apply_security_headers(app)` is called unconditionally in the app factory (`shuffify/__init__.py:540`), so this is not dev-only.

## Scale

| Template | Inline handlers |
|---|---|
| `shuffify/templates/workshop.html` | 79 |
| `shuffify/templates/schedules.html` | 12 |
| `shuffify/templates/dashboard.html` | 5 |
| **Total** | **96** |

```bash
grep -rhoE '\bon(click|change|input|submit|focus|blur|keyup|keydown)=' shuffify/templates/*.html | wc -l
```

No shim converts these to `addEventListener` — verified by grepping for `getAttribute('onclick')`, `removeAttribute('onclick')`, and `[onclick]` selectors. Zero hits.

On `schedules.html` alone the blocked controls are: New Schedule (`:24`), edit panel (`:103`), history (`:110`), run-now (`:117`), enable/disable (`:125`), delete (`:138`), modal close (`:172`, `:287`), job-type change (`:183`), target-playlist change (`:197`), time-picker visibility (`:266`), create-button state (`:470`).

## Evidence (observed, not inferred)

I rendered the real `/schedules` page through the app's own test client (real Jinja + real route context, using the mock stack from `tests/routes/test_schedules_routes.py::test_renders_page_for_authenticated_user`), served that exact HTML with the app's own `Content-Security-Policy` header and the repo's real `static/`, then clicked the real **New Schedule** button in Chromium:

```
modal hidden before click: True
create button found:       True
modal hidden after click:  True        <-- never opened
CSP violations during click: 1
   >> Executing inline event handler violates the following Content Security Policy directive 'script-src ...'
```

[...]

PR #501, opened 2026-07-31
fix(templates): replace undefined showToast with showNotification (#467)

### LINKED ISSUE #467: [tech-debt] showToast() called 5x but never defined - ReferenceError silently breaks rotation-cap validation and schedule-save feedback - templates/workshop.html, schedules.html
## Summary

`showToast(...)` is called 5 times across `templates/workshop.html` and `templates/schedules.html` but is never defined anywhere in the repository — only `showNotification(...)` exists (defined in `static/js/notifications.js` and redundantly re-declared in `workshop.html`). Every call site throws a `ReferenceError` at runtime, which means: the rotation "playlist size cap" input validation silently fails to show its error message, and the raid-schedule update success/failure toast never appears. Both are live, user-facing feedback paths that currently do nothing.

## Evidence

- `shuffify/templates/workshop.html:3663` — `showToast('Please enter a playlist size cap', 'error');` (rotation target-size validation)
- `shuffify/templates/workshop.html:4329` — `showToast('Schedule updated', 'success');`
- `shuffify/templates/workshop.html:4331` — `showToast(data.error || 'Failed to update schedule', 'error');`
- `shuffify/templates/workshop.html:4334` — `.catch(function() { showToast('Failed to update schedule', 'error'); });`
- `shuffify/templates/schedules.html:611` — `showToast('Please enter a playlist size cap', 'error');`
- Confirmed via repo-wide grep (`function showToast`, `showToast =`, `const/let/var showToast`) — zero definitions anywhere in `shuffify/static/` or `shuffify/templates/`.
- The working equivalent, `showNotification(message, type)`, is defined once globally in `shuffify/static/js/notifications.js:8` and loaded via `base.html` on every page (also redundantly re-declared at `workshop.html:2057` — see the companion DRY issue for that).

### Dependencies
None

### Blocks
None

### Steps

1. In `shuffify/templates/workshop.html`, replace all 4 `showToast(` call sites (lines 3663, 4329, 4331, 4334) with `showNotification(`. Signature is identical: `showNotification(message, type)`.
2. In `shuffify/templates/schedules.html:611`, replace the 1 `showToast(` call site with `showNotification(`.
3. Re-run the repo-wide grep to confirm zero remaining `showToast` references:
   `grep -rn "showToast" shuffify/`

### Before/After

Before (`workshop.html:3663`):
```js
showToast('Please enter a playlist size cap', 'error');
```

After:
```js
showNotification('Please enter a playlist size cap', 'error');
```

(same substitution pattern at the other 4 call sites)

## Test Plan

- No existing automated test exercises this JS path (Jinja templates aren't covered by the Python test suite). Manual verification is the only practical check here — see Verification Checklist.
- If Playwright/JS tests are ever added for the workshop rotation panel, this is a good first case: assert the toast DOM node appears when target_size is left blank.

## Verification Checklist

- [ ] `grep -rn "showToast" shuffify/` returns zero results
- [ ] Manual: open Workshop -> Rotation tab, submit rotation config with the playlist-size-cap field blank -> error toast appears (previously: silent no-op)
- [ ] Manual: Schedules page -> edit a schedule -> save -> success toast appears; force a validation error (e.g. invalid cron) -> error toast appears
- [ ] `flake8 shuffify/` and `pytest tests/ -v` still pass (this change touches only `.html`/inline JS, so should be a no-op for both, but the repo's pre-push checklist requires running them)

## What NOT To Do

- Don't write a new `showToast` function to match the call sites — `showNotification` already exists, is already loaded globally, and is the established convention. Adding a second notification function would be the DRY violation this exact area already has one of (see the companion `workshop.html` JS-hygiene issue).
- Don't assume this was intentionally renamed mid-refactor and go looking for a `showToast` implementation elsewhere — it has never existed in this repo's history under either name pattern; verify with `git log -p --all -S showToast` if you want to double check before starting.

## Context

- Source skill: claudna:audit tech-debt
- Audit date: 2026-07-23
- Area: shuffify/templates/workshop.html, shuffify/templates/schedules.html
- Effort: Low
- Risk: Low
- Priority: High
- Related issues: None (verified against all 59 open tech-debt issues; not previously flagged)

chrisrogers37:
Closes #467.

## What was wrong

`showToast(...)` was called 5 times across `workshop.html` and `schedules.html` and defined **nowhere** in the repo — confirmed by grep and by `git log --all -S`, it has never existed under that name. Every call threw `ReferenceError` and aborted its enclosing handler, so two user-facing feedback paths silently did nothing:

- rotation "playlist size cap" validation never showed its error
- raid-schedule save never reported success or failure

## What changed

6 lines, 2 templates.

**1. All 5 `showToast(` → `showNotification(`** — the existing global in `static/js/notifications.js`, identical signature `(message, type)`. It's loaded by `base.html:166`, which sits outside any overridable Jinja block (`base.html` defines only `title` and `content`), and both templates extend `base.html` — so it is genuinely in scope at all 5 sites.

**2. Removed `setSubmitState(false)` at `schedules.html:612`.**

This one is a scope call worth flagging — easy to revert if you'd rather split it. `setSubmitState` is **also** called exactly once and defined nowhere. Renaming `showToast` alone would have moved the `ReferenceError` down one line rather than removing it, so the guard would still be broken after a "fix" that closed the issue.

It's safe to delete rather than implement:
- the submit handler (`:551-647`) never disables the button, so there is no state to clear — the `btn.disabled` logic at `:657`/`:680` belongs to `runScheduleNow()`, a different function
- the sibling guard 20 lines above (`:590`) already uses the bare `showNotification(...)` + `return` form

So the change makes the rotate guard identical in shape to the raid guard directly above it.

## Verification

**The Python suite is not what verifies this change.** 1933 tests pass and flake8 is clean, but this is a change to inline JS inside Jinja templates — the suite is structurally unable to fail on it. It proves nothing was broken, not that anything was fixed.

The real evidence is a browser. I rendered the **actual** `/schedules` page through the app's own test client (real Jinja, real route context, mock stack borrowed from `test_schedules_routes.py::test_renders_page_for_authenticated_user`), served that exact HTML with the app's own `Content-Security-Policy` header and the repo's real `static/`, and drove the reported path — job type `rotate`, size cap empty, submit:

```
showNotification defined: function
showToast defined:        undefined
toast: {'text': 'Please enter a playlist size cap',
        'bg': 'rgba(239, 68, 68, 0.9)', 'opacity': '1'}
NEW pageerrors: 0 []
```

Before the fix, the same path threw `ReferenceError: showToast is not defined`.

**On CSP specifically** (since this landed after the strict-CSP rollout): `showNotification` adds **zero** violations. Measured by phase — 3 calls on a real page produced 0 new violations. It styles via `className` with Tailwind classes, never inline styles, and all 5 call sites already sit inside nonce'd `<script>` blocks. I also confirmed the compiled Tailwind CSS actually contains `.bg-red-500\/90`, `.bg-green-500\/90` and `.bg-blue-500\/90`, so the toast renders with real colour rather than appearing transparent.

## Two things found while verifying — both filed, neither fixed here

- **#499 (critical)** — the strict CSP blocks **all 96 inline `onclick`/`onchange` handlers** (workshop 79, schedules 12, dashboard 5). CSP nonces do not cover event-handler attributes. I clicked the real "New Schedule" button under the real CSP and the modal did not open. **This means the `schedules.html` half of this PR is currently unreachable in production** — you can't open the modal to reach the guard. The fix here is still correct; it's gated behind #499.
- **#500** — `loadWorkshopPlaylist()` at `workshop.html:3814` is called once, defined nowhere. Rotation succeeds, toast shows, then it throws and the track list never refreshes. Not mechanical to fix (needs an invented refresh or a product decision), so it's its own issue.

Three undefined-identifier bugs across two templates suggests these paths were never exercised in a browser. A called-but-never-defined lint step over the templates would have caught all three.


chrisrogers37 (commented):
**Approve**

Checked both symbol-safety questions individually, not by count, and then went looking for the third CSP trap. Found it — bigger than what's disclosed, but it doesn't change the merge call.

**`showNotification` — exists, in scope, compatible signature at all 5 sites. Confirmed individually.** It's actually defined *twice*: the shared global in `static/js/notifications.js:8`, and a second, byte-for-byte identical copy still sitting in `workshop.html:2057` — `git log -S` shows it's leftover from the Feb 23 "Extract reusable Jinja2 macros and shared notification JS (Phase 05)" refactor, which extracted the shared version but never deleted the original. Diffed the two bodies character-for-character: identical. Because `base.html` loads `notifications.js` at line 166, *outside* any block workshop.html/schedules.html can override, and both templates `{% extends "base.html" %}` (confirmed directly), every one of the 5 call sites resolves `showNotification` successfully regardless of which copy wins at load time — no ReferenceError risk from the rename itself. The vestigial workshop.html copy is genuinely dead now (always overwritten by the later-loading global in document order) — harmless, but worth a tiny follow-up cleanup since it's not what this PR is about.

**`setSubmitState` — confirmed fully dead, not just "probably unused."** Unfiltered, repo-wide grep (`.py`/`.js`/`.html`, no exclusions) turns up exactly one occurrence in the entire repository: the call site this PR deletes. No definition anywhere, no other caller anywhere. Also checked the "safe to delete not implement" reasoning directly rather than taking it on faith: the enclosing submit handler (`schedules.html:551-647`) has zero `.disabled` references, and the actual disable/re-enable logic virgil points to (`:657`/`:680`) belongs to `runScheduleNow()` — read it, it's a different function entirely, for a different button. There's no state being silently dropped.

**The CSP angle — this is the one worth reading carefully.** #501's body correctly discloses that the `schedules.html` rotate-guard is unreachable today: the "New Schedule" button's own `onclick` is CSP-blocked (per #499), so the modal never opens and the form never renders. That's 1 of the 5 rewired call sites. What isn't disclosed: **the other 4 have the identical problem**, and I didn't just infer this from the CSP spec — I reproduced it in a real Chromium, the same way #499 did. I traced each `workshop.html` trigger to its actual DOM element:

- `createRotationSchedule()` (site 2) has exactly one caller in the file: `<button id="rotation-enable-btn" onclick="createRotationSchedule()">` (`workshop.html:872`).
- `raidPanel.saveSchedule()` (sites 3–5, the success/error/catch branches of one fetch chain) has exactly one caller: `<button onclick="raidPanel.saveSchedule()" id="raid-schedule-save">` (`workshop.html:650`).

I built a minimal server that serves shuffify's actual CSP header verbatim (copied from `shuffify/__init__.py`'s `set_security_headers`) alongside those two exact button/onclick fragments, and drove it with real Playwright Chromium — not a lint or a spec citation:

```
rotation-enable-btn: fired before=False after=False
raid-schedule-save:  fired before=False after=False
CSP-related console messages: 2
  >> Executing inline event handler violates the following Content Security Policy
     directive 'script-src ... nonce-...'. ... The action has been blocked.
```

Neither function fires. Both produce the exact violation #499 documents. So the accurate count is **5 of 5** rewired call sites currently unreachable in production, not 1 of 5 — all gated behind the same already-filed #499, just not all called out in #501's writeup. (I also independently checked *why* site 1 differs mechanically from 2–5: the schedules.html guard itself is wired via `addEventListener('submit', ...)` inside a nonce'd `<script>` — not CSP-blocked on its own — it's only unreachable because something *upstream* of it is blocked. Sites 2–5's blocked calls are themselves the onclick target. Different mechanism, same net effect.)

This doesn't change my verdict. The JS changes are correct regardless of current reachability — once #499 lands and these buttons work again, these guards will fire correctly on the first try instead of surfacing yet another undefined-identifier bug, which is real value banked now. Blocking #501 wouldn't fix anything; the unreachability predates this PR and isn't caused by it. But right now this PR fixes zero user-reachable paths, not four-fifths of one — worth a follow-up comment on #501 or a note added to #499 (which lists schedules.html's 12 handlers individually but not workshop.html's 79) so nobody reads this as partially-already-working in prod.

Nothing else in the diff needs comment. CHANGELOG entry and the "Closes #467" framing are accurate to what the code does; #500 and #728-equivalent scope questions don't apply here (no linter-path gap in this one). Good instinct filing #499 and #500 separately rather than scope-creeping them into this PR.

Issue #404, opened 2026-06-08
[LOW] Token encryption salt has no documented rotation path


chrisrogers37:
## Summary

The Fernet token encryption uses a fixed salt (`b"shuffify-refresh-token-encryption-v1"`) for PBKDF2 key derivation. Changing this salt invalidates all stored encrypted refresh tokens with no migration tooling in place.

## Location

`shuffify/services/token_service.py:21`

```python
_SALT = b"shuffify-refresh-token-encryption-v1"
```

## Risk

If the salt or SECRET_KEY needs to be rotated (key compromise, employee departure, compliance requirement), all users lose their stored refresh tokens simultaneously and must re-authenticate. Without a migration script, this is a manual "break glass" event.

## Remediation

Document the rotation procedure and consider building a migration script:

1. Decrypt all tokens with old salt/key
2. Re-encrypt with new salt/key
3. Update database in a transaction

Add this to `documentation/guides/credential-rotation.md` (which already exists but may not cover token salt rotation).

## Notes

Low severity because the fixed salt is not itself a vulnerability (SECRET_KEY provides the entropy), and rotation is a rare event. The risk is operational: when it happens, you want a plan, not a fire drill.

PR #381, opened 2026-05-25
fix: remove TempoGradientShuffle — deprecated Audio Features API

chrisrogers37:
## Summary
- Removed TempoGradientShuffle algorithm entirely — depended on Spotify's deprecated Audio Features API (Nov 2024)
- Was hidden from UI but still callable via registry; when called, silently no-op'd (all tracks defaulted to 120 BPM = original order)
- Removed: algorithm file, test file (268 lines), registry entry + `_hidden_algorithms` mechanism, schema validation entry, `direction` parameter
- Audio features infrastructure (api.py, cache.py, client.py, playlist model) left for separate cleanup — `get_playlist_stats` still references it

Closes #334

## Test plan
- [ ] 1901 tests pass (removed 22 tempo gradient tests, no new failures)
- [ ] `flake8 shuffify/` clean
- [ ] Registry now has 7 algorithms (was 8), all visible (no more hidden set)
- [ ] `ShuffleRegistry.get_algorithm("TempoGradientShuffle")` now raises `ValueError`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #502, opened 2026-07-31
security: decouple refresh-token encryption from SECRET_KEY (SR-017)

chrisrogers37:
# security: dedicated token encryption key with zero-loss rotation path (SR-017)

## Why

SR-017 (tracked in #455, related #404): the Fernet key encrypting stored Spotify refresh tokens is derived from `SECRET_KEY` with a static salt. Rotating `SECRET_KEY` — routine security hygiene — silently invalidates every stored refresh token, stopping all scheduled jobs until each user re-logs-in. The issue flags this as do-before-the-next-rotation, and the current `credential-rotation.md` guidance is literally `UPDATE users SET encrypted_refresh_token = NULL;`.

## What changed

- **`TOKEN_ENCRYPTION_KEY`** (new env var): a raw Fernet key that becomes the primary encryption key when set. Raw key rather than passphrase+PBKDF2 — KDF stretching exists for low-entropy passphrases; a generated key needs none, and the #404 static-salt concern disappears for the dedicated path.
- **MultiFernet decrypt chain**: primary → `TOKEN_ENCRYPTION_KEY_FALLBACKS` (comma-separated retired keys, new env var) → the `SECRET_KEY`-derived legacy key. The legacy key is *always* in the chain, so the decrypt capability is a strict superset of today's. Writes always encrypt with the primary, so every login and scheduled-job token refresh migrates that user's token opportunistically.
- **`flask rotate-token-encryption`** (new CLI, `--dry-run` supported): bulk re-encrypts stored tokens onto the primary key; reports `already on primary / re-encrypted / failed` counts; exits non-zero if any token couldn't be decrypted. Command logic lives in `UserService.reencrypt_all_refresh_tokens` (user-row concern, per the service layering); `shuffify/cli.py` is a thin shim registered from `create_app` like error handlers and blueprints.
- **No hard requirement**: unset `TOKEN_ENCRYPTION_KEY` preserves legacy behavior byte-for-byte, plus a startup WARNING. Deliberate — SR-034's history (hard Redis fail-fast shipped, blocked deploys for 3 days, reverted in #376–#381) says a new mandatory env var on a deploy path is hostile. Malformed keys still fail initialization with a message that includes the generation command.
- **`credential-rotation.md` rewritten** around the new procedure: first-time adoption, rotating the dedicated key via fallbacks, ordering rule (re-encrypt to `failed: 0` *before* dropping a fallback or rotating `SECRET_KEY`), updated side-effects matrix and checklist. `.env.example` documents both vars.

## What happens to already-stored tokens (and how it was verified)

The dispatch constraint: a migration that cannot decrypt existing tokens is worse than the current state. Every path below is covered by a test that constructs the *before* state, applies the transition, and decrypts:

| Scenario | Outcome | Verified by |
|---|---|---|
| Deploy this PR, var unset | Nothing changes; same derivation, same ciphertexts accepted | 13 pre-existing token tests pass unmodified in behavior; `test_secret_key_rotation_without_dedicated_key_breaks` documents the unchanged status quo |
| Set `TOKEN_ENCRYPTION_KEY`, restart | Existing tokens decrypt through the legacy chain leg; nothing rewritten at startup | `test_legacy_token_decrypts_after_dedicated_key_added`; db-backed `legacy_then_dedicated` fixture (tokens seeded pre-migration, service re-initialized post-migration, plaintexts compared) |
| Run `flask rotate-token-encryption` | Tokens re-encrypted under the primary; re-runnable (already-current rows untouched); dry-run writes nothing | `TestReencryptAllRefreshTokens` (4 tests) + `TestRotateTokenEncryptionCommand` (3 tests, real CLI runner + in-memory DB) |
| Rotate `SECRET_KEY` after migration | Stored tokens unaffected — the SR-017 acceptance criterion | `test_secret_key_rotation_preserves_dedicated_tokens` and end-to-end `test_survives_secret_key_rotation_after_reencryption`, which rotate the secret inside the test and decrypt the stored rows |
| A stored token no key can decrypt | Reported per-user, counted, **never modified** (ciphertext preserved in case a later key config can recover it); other users still migrate; exit code 1 | `test_undecryptable_token_reported_not_modified`, `test_command_exit_code_on_failures` |
| Rotate `SECRET_KEY` *before* re-encrypting legacy tokens | Those tokens break (unavoidable — the old derivation input is gone). The guide's ordering rule exists for exactly this; the failure mode is tested, not hidden | `test_secret_rotation_before_reencryption_breaks_legacy` |

## Verification

- `flake8 shuffify/` — clean
- `pytest tests/ -m "not integration"` — **1954 passed**, 22 deselected (20 new tests)
- Suite wall-clock dropped **5:40 → 1:59**: `_derive_legacy_key` is now `lru_cache`d (480k-iteration PBKDF2 ≈ 0.1s per call; the new tests would otherwise have added ~40 derivations)
- `black --check` on all changed files — clean (repo-wide drift in 52 untouched files is pre-existing, SR-037)
- Diff scope audited hunk-by-hunk: a formatter-hook whole-file reformat of `user_service.py` and the carried-over token tests was reverted; every remaining hunk is intentional

## Workflow

`/weigh-development-paths` run at the design juncture (key format, env shape, legacy strategy, tooling home). `/simplify` run before push (4 parallel lenses); applied:

- **`TokenService.rotate_token()`** wrapping `MultiFernet.rotate()` — the bulk migration no longer materializes plaintext refresh tokens outside the crypto layer (relevant because Sentry's default `include_local_variables` would have shipped a live token as a frame local on any exception inside the loop)
- **Fail-fast on a set-but-malformed `TOKEN_ENCRYPTION_KEY`** — an explicitly configured key that can't be used now aborts boot with the actionable error instead of warning and running with encryption dead (the one new deployment mistake this feature enables; unset stays warn-and-continue legacy). Covered by a new app-factory test
- **Fallback keys now active in legacy mode too** — the initial cut silently ignored `TOKEN_ENCRYPTION_KEY_FALLBACKS` when the dedicated key was unset, diverging from the documented contract; chain construction unified + test added
- **`scripts/forensics/wooklyn_loss_diff.py` stale re-init removed** — it re-called `initialize(SECRET_KEY)` after `create_app`, which post-change would downgrade a migrated deployment to the legacy key and break its own decryption
- **`ReencryptionResult` dataclass** (matching the class's `UpsertResult` convention) with `failed`/`total` as derived properties — kills two lockstep-mutation invariants; CSV fallback parsing moved to `config.py` beside the repo's other env coercions; CLI failure paths collapsed to `click.ClickException`; dead `except` branch and duplicated init-guard strings removed
- Skipped deliberately: per-user column projection and chunked commits (hundreds of users — wrong scale), lazy legacy-key derivation (0.1s once per boot, not worth the complexity), `store_refresh_token` extraction (dissolved once `rotate_token` removed the plaintext write), repo-wide autouse TokenService reset fixture (right idea, too broad for this PR — done module-locally in the new test file)

Closes SR-017 checkbox in #455. Related: #404.

https://claude.ai/code/session_0168wKn3GuCVmHPmsuRX9gwY


chrisrogers37 (commented):
**Approve**

Weighted my time exactly where you asked — the migration and boot behavior got the deepest treatment, and I mutation-tested rather than just read. Nothing here was underspecified enough to need to ask astrid directly; every claim in the PR body held up under independent, from-scratch verification.

**1. Migration against REAL old-scheme ciphertext — proven, not inferred.** A same-run round-trip proves nothing, so I didn't do one. I pulled the actual, unmodified pre-PR `token_service.py` verbatim from `origin/main` (not a reimplementation), loaded it as an isolated module alongside the new one, and:
- Encrypted a token with the real OLD `TokenService.initialize(secret)` + `encrypt_token`.
- Decrypted that exact ciphertext with the NEW code, `TOKEN_ENCRYPTION_KEY` unset (today's state for every existing deploy) → correct plaintext.
- Decrypted the same ciphertext again with the NEW code, `TOKEN_ENCRYPTION_KEY` **set** (post-migration state) → correct plaintext, via the legacy leg of the chain.
- Ran `rotate_token()` on it → lands on the primary key (`is_encrypted_with_primary` true), plaintext still correct.

All four checks passed. The "legacy key is always in the chain" guarantee is real, not aspirational.

Then I mutation-tested the test suite itself, since a real guarantee needs tests that would actually catch it breaking: commented out `if primary is not legacy: chain.append(legacy)` in `initialize()` and reran `test_token_reencryption.py` + `test_token_service.py`. **9 tests failed immediately**, including `test_survives_secret_key_rotation_after_reencryption` (the SR-017 acceptance test) and `test_legacy_token_decrypts_after_dedicated_key_added`, with the exact diagnostic error ("token is corrupted or was encrypted with a key that is no longer configured"). Restored the file, suite green again. These are not decoy tests.

**2. Absent vs. malformed key — traced and independently tested, both behave correctly, and the SR-034 failure mode specifically doesn't recur.**
- Absent (`TOKEN_ENCRYPTION_KEY` unset): `initialize()` takes the `primary = legacy` branch, no exception, warns, boots. Ran the app's own `create_app("testing")` for real with the key unset — boots clean, `TokenService.is_initialized()` true.
- I also tried the not-quite-absent edge case that trips up SR-034-style incidents in practice: `TOKEN_ENCRYPTION_KEY=""` (declared-but-empty, a real footgun on some deploy platforms). Empty string is falsy in Python, so it takes the identical safe branch — also boots clean. Confirmed with my own test, not the PR's.
- Malformed (set, invalid Fernet material): `_fernet_from_key` raises, `_init_token_encryption`'s except-block checks `app.config.get("TOKEN_ENCRYPTION_KEY")` — truthy — re-raises, boot fails. Ran the PR's own `test_create_app_fails_fast_on_malformed_token_key` for real; passes.
- I chased one more edge case before trusting the branch: what if `SECRET_KEY` itself is also missing? Traced it — `validate_required_env_vars` requires `SECRET_KEY` in production and fails boot at line 458 of `__init__.py`, *before* `_init_token_encryption` (line 489) is ever reached; dev/test configs auto-generate or hardcode a `SECRET_KEY` so it's never empty there either. So the one compound case that could have made the absent/malformed split misfire is structurally unreachable, not just unlikely.

**3. Rotation command, partial failure at row N.** `reencrypt_all_refresh_tokens` fetches all token-bearing users, mutates each successfully-rotated row *in memory* inside the loop, and calls `safe_commit` exactly **once, after the loop** — one all-or-nothing transaction for the whole batch. So a hard crash at row N leaves the database completely untouched (nothing was ever flushed); restarting the command is safe and idempotent, re-scanning from scratch, because `is_encrypted_with_primary` correctly skips whatever *did* make it to the primary key in a prior successful run. Checked `safe_commit` itself: on failure it rolls back and re-raises rather than swallowing, so a failed final commit surfaces loudly rather than reporting false success. The PR discloses skipping chunked commits as a deliberate scale call ("hundreds of users — wrong scale") — agree with that judgment for this dataset size.

**4. `/simplify`'s two behavioral changes, checked individually.**
- `rotate_token()` wraps `MultiFernet.rotate()` directly — confirmed by reading it, plaintext is never assigned to a local variable in `token_service.py` or `user_service.py`'s reencryption loop; the decrypt-and-reencrypt happens entirely inside the `cryptography` library. Real containment, not just a docstring claim.
- The `lru_cache` on `_derive_legacy_key`: it's a pure function of one argument (`secret_key`) against a hardcoded salt/iteration count/algorithm — nothing that could vary is outside the cache key. Stress-tested it directly: alternating between two secrets across 5 `initialize()`/`reset()` cycles (what the test suite itself does hundreds of times), plus a deliberate wrong-secret cross-check on each cycle — the cache never served a key to a secret it didn't belong to. The primary key (`TOKEN_ENCRYPTION_KEY`) itself isn't cached at all — it's a plain class attribute set fresh on each `initialize()` call — so there's no separate staleness surface there. I also traced why "serve a stale key across a rotation" specifically can't happen here: rotating the primary key requires a process restart (env var + redeploy), which starts with an empty `lru_cache`; the CLI rotation command is a separate process from any running web server, not a live in-memory mutation the cache could go stale against.

**Other things I checked along the way:**
- `scripts/forensics/wooklyn_loss_diff.py`'s stale `TokenService.initialize(app_config["SECRET_KEY"])` removal is real and matters — that positional-only re-init call would have silently downgraded an already-migrated deployment's `TokenService` back to legacy-only (since `initialize()` replaces state rather than merging it), breaking decryption for anyone already on the primary key. Confirmed by reading `initialize()`'s replace semantics directly.
- CHANGELOG.md's two-dot diff against current main looked like it *removed* #501's just-merged showToast entry — that's a stale-base artifact, not a real problem: #502's branch is exactly 1 commit behind main (the #501 merge that landed after #502 was opened), and the three-dot diff against the actual merge-base shows #502 never touches that section at all. A real merge will keep both entries; nothing to fix.
- `ReencryptionResult.failed`/`.total` are derived properties from `failed_spotify_ids`, not separately-tracked counters — can't drift out of sync with the list.
- flake8 clean (ran it myself, exit 0). CI green (all 3 checks — Tailwind, Backend lint & tests, GitGuardian — SUCCESS as of this review, confirmed via the API not just the badge). `pytest tests/ -m "not integration"` collects exactly 1976 items / 22 deselected / 1954 selected on my own machine — matches the PR's stated numbers exactly. Ran every security-relevant file directly and in full (token_service, user_service reencryption, app_factory) — all pass; CI's own green run is the authoritative confirmation for the remaining ~90% of the suite I didn't re-run file-by-file myself.

Nothing here needs to change before merge.

Issue #339, opened 2026-05-20
REDIS_URL silently falls back to filesystem sessions in production


chrisrogers37:
## Finding

**Severity:** Medium  
**Category:** Configuration  
**Files:** `config.py:136`, `shuffify/__init__.py:124-147`

`ProdConfig.REDIS_URL = os.getenv("REDIS_URL")` evaluates to `None` if the env var is absent. `_init_redis` silently falls through to filesystem sessions — no error, no startup failure. `validate_required_env_vars` does not check `REDIS_URL`.

## Recommendation

Add `REDIS_URL` to the production required env var validation. A missing Redis URL in production should fail fast, not silently degrade to filesystem sessions.

chrisrogers37:
[rajan] Re-verified against current `main` (`1d8b660`). Still live: `shuffify/__init__.py:155` — `logger.warning("REDIS_URL not configured. Using filesystem sessions.")`. No raise, no validation — a real config gap. This is the same issue mason's 2026-07-29 pass flagged as Category C ('named in a fix commit, not actually fixed') — confirming it still holds 3+ weeks later. Keeping open — real, unaddressed.

Issue #325, opened 2026-05-20
DevConfig.TESTING = True causes Alembic migrations to be skipped in development


chrisrogers37:
## Finding

**Severity:** High  
**Category:** Configuration  
**Files:** `config.py:160`, `shuffify/__init__.py:211-213`

`_init_database()` branches on `app.config.get("TESTING")` — when `True`, it skips Alembic and calls `db.create_all()`. `DevConfig` sets `TESTING = True`, meaning every developer running locally uses `create_all()` instead of the migration chain. Migration correctness is therefore only tested in production. Schema drift between what Alembic produces and what `create_all()` produces is never caught in dev.

## Recommendation

Set `TESTING = False` in `DevConfig`. `TESTING = True` should only exist on `TestConfig`.

chrisrogers37:
[rajan] Re-verified against `main` (`1d8b660`). `DevConfig.TESTING = True` (`config.py:187`) still routes `_init_database()` to the `create_all()` branch instead of the migration chain (`__init__.py:374-376`). Strong corroboration: the function's own docstring (added by PR #503, 2026-08-01) names this exact issue number as open and unresolved verbatim. Looks deliberately deferred pending #531 (migration freeze) rather than forgotten — context, not a change to the verdict. Keeping open — real, unaddressed.

PR #503, opened 2026-07-31
fix(db): run migrations as a deploy step, verify schema at head in production (SR-018)

chrisrogers37:
## What

`flask_migrate.upgrade()` ran inside the Flask app factory and its failures were caught, logged, and swallowed (`shuffify/__init__.py` ~208–249). A container whose migration failed kept serving traffic against a possibly stale schema, and nothing but a log line distinguished that from a healthy boot.

Migrations now run as a deploy step in `scripts/docker-entrypoint.sh` — once per container, before Gunicorn starts. In production the app factory no longer migrates; it **verifies** the schema is at Alembic head and raises `SchemaOutOfDateError` out of `create_app()` if it is not.

SR-018, tracked in the #455 reliability cluster. (Not using a closing keyword — #455 has four other open items.)

## Design notes

**`ENTRYPOINT`, not a longer `CMD`.** There is no app spec in this repo, so the DO App Platform run command can be overridden in the console. A run-command override replaces `CMD` but still runs through `ENTRYPOINT`, so the migration step survives it.

**`MIGRATE_ON_STARTUP` config attribute, not a string check.** Which strategy an environment uses is declared on the `Config` classes (`False` on `ProdConfig`), mirroring `SCHEDULER_ENABLED`. A future profile that also migrates out-of-band declares it next to its other flags instead of needing a new branch in the factory.

**Two env vars, different jobs:**

| Var | Who sets it | Effect |
|-----|-------------|--------|
| `SHUFFIFY_ALLOW_SCHEMA_DRIFT` | operator, break-glass | Entrypoint tolerates a failed upgrade **and** the factory downgrades the guard to an ERROR log. One variable clears both gates — both readers trim surrounding whitespace and match case-insensitively, pinned to one parametrized table so the two implementations cannot drift apart. |
| `SHUFFIFY_MIGRATION_STEP` | the entrypoint only, internal | Exempts the migration process from the guard. `flask db upgrade` builds the whole app and would otherwise be refused for the schema state it exists to fix. |

## How this cannot wedge a deploy the way #339 did

SR-034 records that a hard Redis fail-fast (#339) blocked DigitalOcean deploys for three days and was reverted. The discriminator is **not** hard-vs-soft — it is whether the requirement is satisfiable from the artifact being deployed:

- **#339** required Redis, which did not exist and could not be conjured from the repo. The only exit was provisioning infrastructure.
- **This guard** requires "schema == head". The thing that satisfies it — the migration chain and `flask db upgrade` — ships inside the same image, and the entrypoint runs it automatically on every boot.

The residual risk is "entrypoint bypassed", and that is what `SHUFFIFY_ALLOW_SCHEMA_DRIFT` covers: one variable set in the DO console, one restart, no code change and no redeploy-of-a-fix. The failure message names the variable, so whoever sees the crash loop is told the recovery step.

### How I verified it

Built the **real image** from this branch and ran it against a throwaway SSL-enabled Postgres and a SQLite volume. Not unit tests — actual containers:

| # | Scenario | Result |
|---|----------|--------|
| 1 | `CMD` overridden (platform run-command override) | ENTRYPOINT still runs the migration step |
| 2 | Redeploy at head | upgrade is a no-op; `Database schema verified at Alembic head` |
| 3 | **Entrypoint bypassed** (`--entrypoint` override), reachable unmigrated DB | **exit 1** — `SchemaOutOfDateError: current=none (database never migrated), head=c4d5e6f7a8b9` |
| 4 | **Same stale DB + `SHUFFIFY_ALLOW_SCHEMA_DRIFT=true`** | **exit 0** — `starting anyway`, app built |
| 5 | Migration itself fails | entrypoint exits 1, container refuses to start |
| 6 | Same failure + override | container starts and serves |
| 7 | **Genuinely fresh database, real entrypoint, nothing stamped** | **exit 1 at migration 10 of 20** — see the correction below |

Rows 3 and 4 are the SR-034 claim: the guard blocks a genuinely stale deploy, and one env var clears it with no code change.

#### Correction to the original row 1

It read *"Normal deploy, `CMD` overridden → ENTRYPOINT still ran; 9 migrations applied; app built"*, which reads as one from-scratch deploy succeeding end to end. **It is not reproducible as written**, and the review was right to push on it.

Re-ran it from scratch — real `migrations/` chain, SQLite, `APP_CONFIG=production`, nothing stamped:

```
INFO  [alembic.runtime.migration] Running upgrade  -> aa2823fcd4b0     (1 of 20)
    ... 9 applied ...
INFO  [alembic.runtime.migration] Running upgrade cb0799fcc893 -> d4e5f6a7b8c9
sqlite3.OperationalError: near "ALTER": syntax error
[SQL: ALTER TABLE upstream_sources ALTER COLUMN source_playlist_id DROP NOT NULL]

exit 1        alembic_version = cb0799fcc893        (head is c4d5e6f7a8b9)
```

The "9 migrations applied" figure is real — it is the count that lands **before the chain aborts**, not a completed upgrade. What was wrong is the conjunction with "app built". From that exact database, `create_app('production')` refuses, which is the guard working:

```
SchemaOutOfDateError: Database schema is not at Alembic head:
current=cb0799fcc893, head=c4d5e6f7a8b9
```

and the entrypoint exits 1 long before Gunicorn is reached. Two runs got merged into one row: a from-scratch upgrade attempt that aborts, and an app build that was against a database already at head — row 2's case. The `ENTRYPOINT`-survives-a-`CMD`-override mechanic in that row is separately true and was independently re-confirmed in review; only the migrations-and-build outcome was wrong.

An honest summary of what is reproducible from scratch today: **the entrypoint runs, the chain aborts partway, and the container refuses to start.** That is the designed behaviour meeting #504, not a defect in this change — see below.

**This verification caught a real bug in my own first cut.** `flask db upgrade` imports `run.py`, which builds the app, which ran the new guard — which refused to start because migrations had not run yet. The migration step could never execute: a deploy that could not be unwedged even by redeploying. That is precisely the SR-034 failure class, and no unit test would have found it. Fixed with `SHUFFIFY_MIGRATION_STEP` and pinned by a regression test.

### Tests

36 new tests in `tests/test_db_schema_guard.py`, exercising the real `migrations/` chain and shelling out to the real entrypoint script (no Docker needed in CI). Full suite: **1969 passed, 0 failed**. `flake8 shuffify/` clean; `shellcheck` clean.

Every guard is **mutation-tested** — I gutted each one and confirmed the tests go red:

| Mutation | Result |
|----------|--------|
| Drop the `except SchemaOutOfDateError: raise` (restores the SR-018 swallow) | 2 red |
| Guard logs instead of raising | 13 red |
| `ProdConfig` loses `MIGRATE_ON_STARTUP = False` | 2 red |
| Migration-step exemption always on | 18 red |
| Sloppy truthiness on the override vars | 7 red |
| Entrypoint stops setting the marker (reintroduces the deadlock) | 1 red |
| Entrypoint no longer exits nonzero on failure | 2 red |
| Entrypoint tolerates failure unconditionally | 2 red |

## ⚠️ #504 — this PR does not create the fresh-environment problem, it reveals it

Filed as **#504**. The Alembic chain cannot replay from scratch on either backend:

- `migrations/versions/e2c728c936de_...py:24` — `server_default=sa.text('1')` on a `Boolean`. Fails on **Postgres** at migration **2 of 20**: `DatatypeMismatch: column "is_active" is of type boolean but default expression is of type integer`. Landed 2026-02-12 (`80f9e3f`).
- `migrations/versions/d4e5f6a7b8c9_...py:44,53` — `op.alter_column` outside `batch_alter_table`. Fails on **SQLite** at migration **10 of 20**: `near "ALTER": syntax error`. Landed 2026-03-09 (`afda547`).

Both reproduce on `main`, independent of this branch.

**The operational statement is bigger than "the chain cannot replay."** There is no path to Alembic head for a new environment through the normal deploy flow, on either backend. Postgres — the production backend — dies at migration 2 of 20, and has since February.

It would be easy to read that as this guard breaking fresh deploys. It does not, and the difference matters. Pre-#503 behaviour, `main`, genuinely fresh production database:

```
create_app('production')   ->  app BUILT (would serve traffic)
alembic_version            ->  cb0799fcc893    (head is c4d5e6f7a8b9)
tables present             ->  12
```

The in-factory upgrade fails at the same migration, the failure is swallowed, and the process comes up and answers `/health` on a schema ten revisions behind the code it is running. **That is not a working deployment — it is a broken one that reports healthy**, which is SR-018 itself. So a fresh environment was already unusable; what changes here is that it now says so at the entrypoint and names the revision gap, instead of surfacing later as data errors that read as application bugs.

Sequencing: #504 is a hard blocker for standing up any new environment (staging, DR, a new region) and wants fixing regardless of this PR. It does not block **this** PR, whose blast radius is production — already past both revisions, so `flask db upgrade` is a no-op there and the guard passes cleanly (row 2).

**What I could not verify:** production's actual current revision — I have no prod credentials and that is outside my scope. The change is safe **if** prod is already at head, in which case `flask db upgrade` is a no-op and the guard passes; scenario 2 above is that case. Worth confirming with `flask db current` against prod before merge.

## Risk / rollback

Rollback is `git revert` — no data migration, no schema change of its own. The entrypoint is additive; reverting restores the previous in-factory upgrade.

## Skills

- `/weigh-development-paths` — run **at** the design juncture rather than after, since the fork (entrypoint-only vs in-app vs hybrid) determined the implementation. Full 7-dimension matrix over 4 options; picked entrypoint-migrates + production-verifies.
- `/simplify` — 4 review agents. Applied: the `MIGRATE_ON_STARTUP` config attribute (replacing both the `config_name` string branch and a redundant parameter), a documentation correction where I had over-claimed that development migrates in-process (it does not — `DevConfig.TESTING=True` short-circuits, which is #325), and ~60 lines of test-fixture dedup. Skipped with reasons: extracting the schema helpers to their own module (the `_init_*` helpers all live in `__init__.py`; a schema-only carve-out would be asymmetric — worth doing as its own refactor), unifying env-var truthiness with `config.py`'s stricter `== "true"` (would change `SCHEDULER_ENABLED` behaviour for anyone using `=1`; permissiveness is deliberate on a recovery lever), and rewriting `migrations/env.py` to avoid building the full app during the migration step (~2s per deploy, not material, and out of scope).

https://claude.ai/code/session_019p1ENFR23SAYGxpNyTBkM8


chrisrogers37 (commented):
**Verdict: Request changes**

Verified by running, not by reading — real disposable Postgres 15 (SSL-enabled, to match `ProdConfig`'s `sslmode=require`) via Docker, real SQLite, real `flask` CLI, real `gunicorn`, all against the actual PR branch (`c8b05df`). Base freshness checked: merge-base against a freshly-fetched `origin/main` is clean, GitHub reports MERGEABLE.

### Core claim — confirmed, several independent ways

- Direct `create_app('production')` against a never-migrated real Postgres DB raises `SchemaOutOfDateError`.
- Real `gunicorn --bind ... run:app` against the same DB: worker fails to boot, master shuts down, exit 3, "Reason: Worker failed to boot." Genuine production boot refusal, not a mocked assertion.
- `SHUFFIFY_ALLOW_SCHEMA_DRIFT=true` against the same stale DB: gunicorn boots for real, `/health` returns 200, the ERROR log fires exactly as designed.
- SR-034 chicken-egg exemption: mutated the real entrypoint to strip `SHUFFIFY_MIGRATION_STEP=1` off the `flask db upgrade` invocation and reran against a fresh DB — `flask db upgrade` now fails immediately, because app construction itself raises `SchemaOutOfDateError` before the migration can even attempt to run. That's the exact #339-class deadlock this PR says it caught. Restored the exemption, reran, confirmed it proceeds normally. The fix is real and load-bearing, not decorative.
- `except SchemaOutOfDateError: raise` block: removed it, reran against the stale DB — the error gets swallowed by the generic handler, exactly restoring the original SR-018 bug. Restored, confirmed.
- Dockerfile `ENTRYPOINT`/`CMD` mechanic: minimal synthetic container confirms a CMD-only override (matching a platform run-command override) still executes the entrypoint; only an explicit `--entrypoint` override skips it.
- `tests/test_db_schema_guard.py` + `tests/test_app_factory.py`: 78/78 pass for real in a clean venv against the real migrations chain.

### Finding 1 (blocking) — the break-glass override doesn't actually clear both gates

`scripts/docker-entrypoint.sh`'s `is_true()` does not strip whitespace before matching; `shuffify/__init__.py`'s `_env_is_true()` does (`.strip().lower()`). Ran both side by side with identical inputs:

| value | shell `is_true()` | python `_env_is_true()` |
|---|---|---|
| `true` | true | True |
| `" true"` (leading space) | **false** | **True** |
| `"true "` (trailing space) | **false** | **True** |

`SHUFFIFY_ALLOW_SCHEMA_DRIFT=" true"` — a plausible copy-paste artifact into a platform console text field — is honored on the Python side but silently rejected by the shell entrypoint. Since the entrypoint runs first and `exit 1`s before Gunicorn ever starts, the more-permissive Python side never gets a chance to matter — the deploy just stays blocked. That's the SR-034 failure mode ("recoverable... without a code change... one variable clears both gates") in exactly the scenario it exists to cover: an operator pasting the override mid-incident. One-line fix — trim whitespace in the shell `is_true()` before the case match.

### Finding 2 (needs reconciling) — verification table row 1 doesn't reproduce

The PR claims: "Normal deploy, CMD overridden → ENTRYPOINT still ran; 9 migrations applied; app built." I ran the real, unmodified entrypoint against a genuinely fresh, never-before-migrated database on both backends:

- **Postgres 15** (SSL-enabled, matching `ProdConfig`): fails at migration **2 of 20** (`aa2823fcd4b0 → e2c728c936de`) — `DatatypeMismatch: column "is_active" is of type boolean but default expression is of type integer`.
- **SQLite**: fails at migration **9 of 20** (`cb0799fcc893 → d4e5f6a7b8c9`) — `near "ALTER": syntax error`.

Both match the separately-disclosed "migration chain cannot replay from scratch on either backend" defect exactly — but that disclosure is in tension with the "9 migrations applied; app built" result claimed for scenario 1. I could not get a genuinely fresh DB to a built, serving app on either backend without manually stamping `alembic_version` directly (a legitimate way to test the *verify* logic in isolation — did that too, scenario 2 passes cleanly — but it's not what row 1 describes doing). Whatever DB scenario 1 actually ran against, it doesn't look like a from-scratch one; worth correcting in the PR body since it's the primary evidence offered for "a normal deploy works cleanly."

This also sharpens the stakes of the disclosed-as-separate defect: it isn't just "a fresh environment fails loudly instead of serving on a broken schema" — a fresh environment fails at step 2 or 9 of 20, deterministically, on both backends, with zero path to head via the normal deploy flow until that chain is fixed. Worth cross-linking the two issues rather than treating them as fully independent.

### Not checked

Full 1969-test suite (CI already reports it green; ran the two directly-relevant files myself instead: 78/78). Production's actual current revision — no prod access, same limitation the author names.

### Bottom line

The core mechanism — migrate-at-entrypoint, verify-at-boot, fail closed — is real, and every way I tried to break it, it held except one: the break-glass override itself has a genuine whitespace gap in precisely the scenario SR-034 exists to cover. Fix that (trivial), reconcile the scenario-1 verification claim against what's actually reproducible, and this is good to merge.

chrisrogers37:
Both findings addressed in `763e23d`, and I agree with the framing in your closing paragraph — with one sharpening that makes it land harder, not softer.

## Finding 1 — fixed

You were right, and the diagnosis was exact: the stricter reader runs first, so the Python side's tolerance never gets a vote. The entrypoint now trims before matching:

```sh
_value=$(printf '%s' "${1:-}" | tr '[:upper:]' '[:lower:]' |
    sed 's/^[[:space:]]*//; s/[[:space:]]*$//')
```

The interesting part isn't the trim — it's that one operator-facing switch had two independent implementations of "is this set," with nothing holding them together. Fixing only the shell leaves the same gap open in the other direction. So both sides are now parametrized over a single table (`OVERRIDE_VALUES`, 16 cases including your ` true` / `true `), exercised through the shipped script and through the factory helper:

```
tests/test_db_schema_guard.py    68 passed        (was 36; +32 parity cases)
```

Mutation-tested in both directions, since a parity gate that only catches one side is half a gate:

| Mutation | Result |
|---|---|
| Drop the shell `sed` trim (restores exactly the bug you found) | **4 red** — ` true`, `true `, `  TRUE  `, `\ttrue` |
| Drop Python's `.strip()` (the mirror drift) | **4 red** on the factory side |

`shellcheck` clean, `sh -n` clean, `flake8` clean, `black --check` clean.

## Finding 2 — reconciled, and you were right that it doesn't reproduce

I re-ran it from scratch rather than restating it. Real chain, SQLite, `APP_CONFIG=production`, nothing stamped:

```
INFO  [alembic.runtime.migration] Running upgrade  -> aa2823fcd4b0     (1 of 20)
    ... 9 applied ...
INFO  [alembic.runtime.migration] Running upgrade cb0799fcc893 -> d4e5f6a7b8c9
sqlite3.OperationalError: near "ALTER": syntax error

exit 1        alembic_version = cb0799fcc893        (head is c4d5e6f7a8b9)
```

Your reading was correct on both counts. The "9 migrations applied" number is real — it is the count that lands *before the chain aborts* — but it was reported next to "app built," and those came from two different runs. From that exact database:

```
SchemaOutOfDateError: Database schema is not at Alembic head:
current=cb0799fcc893, head=c4d5e6f7a8b9
```

The build in the original row 1 must have been against a database already at head, which is row 2's scenario. The `ENTRYPOINT`-survives-a-`CMD`-override half of that row is separately true, and your synthetic container re-confirmed it independently — only the migrations-and-build outcome was wrong.

PR body amended: row 1 now claims only the mechanic, a new row 7 records the actual from-scratch outcome, and there is an explicit correction section stating what was wrong rather than quietly editing the number.

One incidental finding from re-running it, worth knowing for anyone reproducing this: the config selector is `APP_CONFIG`, not `FLASK_ENV`. My first attempt exported `FLASK_ENV=production`, silently got `DevConfig`, and `TESTING=True` ran `db.create_all()` before the upgrade — which then died at migration 1 with `table users already exists`. A plausible way to get a misleading result, and possibly how the original row went wrong.

## On the framing — agreed, and I'd go one step further

You wrote that this "sharpens the stakes" of #504: not just "a fresh environment fails loudly," but that it has **no path to head via the normal deploy flow**. That is right, and it understates it slightly by implying the state changed with this PR. It didn't. Here is `main`, pre-#503, against a genuinely fresh production database:

```
create_app('production')   ->  app BUILT (would serve traffic)
alembic_version            ->  cb0799fcc893    (head is c4d5e6f7a8b9)
tables present             ->  12
```

The in-factory upgrade fails at the same migration, the failure is swallowed, and the process comes up and answers `/health` on a schema ten revisions behind the code running against it. That is not a working deployment — it is a broken one that reports healthy. Which is SR-018, stated as an operational fact rather than a code smell.

So the accurate headline is stronger than either of ours: **shuffify has had no working from-scratch deploy path since February**, when `e2c728c936de` landed the Postgres defect at migration 2 of 20 (`80f9e3f`, 2026-02-12; the SQLite defect followed in `afda547`, 2026-03-09). Production is unaffected only because it was built incrementally and is already past both revisions. This PR doesn't create that, doesn't worsen it, and can't fix it — it converts an invisible failure into a visible one that names the revision gap.

Cross-linked both ways, and #504 now carries the pre-#503 evidence and the "no from-scratch deploy path" framing in its own words rather than by reference to this PR. Sequencing I'd suggest: #504 is a hard blocker for standing up any new environment and wants fixing on its own merits; it doesn't gate this PR, whose blast radius is a production database already past both revisions — `flask db upgrade` is a no-op there and the guard passes (row 2).

Thanks for running it rather than reading it. The whitespace gap is not something review-by-inspection finds, and the row-1 conflation would have shipped as evidence for a claim it didn't support.

chrisrogers37 (commented):

Re-reviewed `763e23d` against fresh checkouts of both this branch and current `main`, all three points independently.

### 1. Whitespace fix — confirmed durable, not just currently-correct

`OVERRIDE_VALUES` (16 cases) is one shared table in `tests/test_db_schema_guard.py`, consumed by two parametrized test functions — one calling `_env_is_true` directly, one driving the real shipped `docker-entrypoint.sh` end-to-end. Ran the suite: 68 passed (was 36). Then mutation-tested both directions myself, not re-derived from the PR body:

- Dropped the shell `sed` trim → **4 red** (` true`, `true `, `  TRUE  `, `\ttrue`), restored, green.
- Dropped Python's `.strip()` → **4 red**, same 4 cases, restored, green.

Both implementations are independently tested against the identical input set, so a future one-sided edit turns this class red rather than silently reintroducing the gap. That's the actual guarantee "cannot drift apart" needs — true code-sharing isn't possible across the shell/Python boundary, but this is the CI-enforced equivalent.

### 2. "9 migrations applied" row — reproducible, with one precision correction

Reran from scratch on fresh SQLite with `main`'s (pre-#503) code: 9 `Running upgrade` lines log (attempts), of which **8 succeed** and the 9th (`cb0799fcc893 → d4e5f6a7b8c9`) is the one that fails — "9 applied" is counting attempts-started, not successes; "8 applied, 9th fails" is the precise statement. Doesn't change anything material: the resulting state — `alembic_version=cb0799fcc893`, **12 tables** — reproduced exactly on my own fresh run, independent of branden's.

### 3. LOAD-BEARING claim — the thesis holds, the specific numbers don't, and the real number is worse

Confirmed on `main`'s actual code: `create_app('production')` against a genuinely fresh database **does build and would serve traffic** — true, and independently reproduced. But the cited numbers (12 tables, `alembic_version=cb0799fcc893`) are **SQLite's outcome, not Postgres's** — and Postgres is the actual production backend (`ProdConfig` requires `sslmode=require`). I built a disposable SSL-enabled Postgres and ran the identical test:

```
create_app('production')  ->  app BUILT (would serve traffic) — same as SQLite
resulting schema           ->  0 tables. No alembic_version. "database never migrated."
```

Instrumented `flask_migrate.upgrade()` directly to confirm: the same `DatatypeMismatch` fires on migration 2 (`e2c728c936de`), exactly as before — but Postgres runs the whole migration batch in one transaction (`migrations/env.py` never sets `transaction_per_migration`, so this is Alembic's default), so migration 1's success rolls back too. Nothing survives, not even the initial schema.

Precise revision count via `ScriptDirectory.walk_revisions`, not eyeballed: SQLite's stopping point (`cb0799fcc893`) is **12 revisions behind head**, not 10. Postgres's stopping point is **all 20** — "revisions behind" undersells it; the correct description is what `_verify_schema_at_head` itself would say: `current=none (database never migrated)`.

Also confirmed independently: `e2c728c936de` (`80f9e3f`, 2026-02-12) and the SQLite-breaking migration (`afda547`, 2026-03-09) have never been touched since — full `git log --follow` on both files shows only their introducing commits. No fix has landed. The "no working from-scratch deploy since February" thesis is correct and, on the real backend, understated rather than overstated.

**What to tell Chris:** the headline claim is right, but swap the evidence — a fresh production (Postgres) deploy today doesn't get a partially-outdated database, it gets a **completely empty one that reports healthy**: zero tables, zero migrations retained, serving 500s on every DB-touching route. Worse than "10 revisions behind," not the same.

### Bottom line

Both findings from the first review are genuinely fixed, verified by running (mutation-tested, not re-read). The one further claim I was asked to check independently needed a correction — same conclusion, wrong backend, more severe reality. Nothing here blocks merge; #504 is the right place for the from-scratch-deploy fix.

PR #506, opened 2026-08-01
fix(raid): scope raid-source-count updates to the playlist in the URL (SR-038)

chrisrogers37:
## Why

`PUT /playlist/<playlist_id>/raid-source-count` took `source_id` from the body and `playlist_id` from the path, then looked the source up by owner alone:

```python
source = UpstreamSource.query.filter_by(id=source_id, user_id=user_id).first()
```

`playlist_id` was never used. A source is configured against exactly one target playlist, so the playlist in the URL was decorative — the request was applied outside the scope it named.

The sibling method in the same service already scopes this way. `update_track_count` filters on `user_id + target_playlist_id + source_playlist_id`; `update_raid_count` was the odd one out.

## Proving the door was open

Exercised against unmodified `main` (`0b702fc`) before writing the fix — a source seeded on `playlist-A`, then updated through a different playlist's URL:

| probe | before the fix | after the fix |
|---|---|---|
| `/playlist/playlist-B/raid-source-count` targeting playlist-A's source | **HTTP 200, raid_count 5 → 99** | HTTP 404, unchanged |
| `/playlist/not-my-playlist/raid-source-count` targeting own source | **HTTP 200, raid_count 5 → 77** | HTTP 404, unchanged |
| another user's source via the correct playlist | HTTP 404, unchanged | HTTP 404, unchanged |

Same probe, same seeds, both runs.

## On the severity label

The gap is real and the fix is right, but the third row is worth stating plainly, because it bounds the impact: **owner scoping was already enforced and this never reached another user's data.** `user_id` was always in the filter. The reachable effect is that a caller could update *their own* source through a playlist URL that doesn't own it — including a playlist ID they have no relationship with.

So this is an integrity/scope defect confined to the caller's own account, not a cross-tenant authorization bypass. P3 looks defensible on blast radius. It is still worth closing: the route accepted requests it should have rejected, and the missing filter is the kind of thing that becomes a real hole the moment some future caller reaches this service without an owner check. Flagging the distinction rather than restating the triage label, since the label drives how the rest of the cluster gets prioritised.

## What changed

Five lines. `update_raid_count` takes `target_playlist_id` and adds it to the filter; the route passes the `playlist_id` it already has. No new query, no behaviour change on the legitimate path.

## Verification

- **Four regression tests** added, covering both directions: the correct playlist still updates (positive control), another playlist cannot, an unowned playlist cannot, and another user's source stays unreachable.
- **Mutation check:** dropping the `target_playlist_id` filter fails exactly the two scope tests while the positive control and the cross-user test keep passing — so the tests bind to this fix rather than to the route being broken in general.
- `flake8 shuffify/` clean. `black --check` reports the same 53 files with and without this change — pre-existing tree-wide drift, and that CI step is `continue-on-error: true` pending the SR-037 ruff consolidation.

Closes SR-038 in #459.

### DIFF
--- a/shuffify/routes/raid_panel.py
+++ b/shuffify/routes/raid_panel.py
@@ -281,7 +281,10 @@ def raid_source_count_update(
 
     try:
         source = UpstreamSourceService.update_raid_count(
-            user.id, req.source_id, req.raid_count
+            user.id,
+            playlist_id,
+            req.source_id,
+            req.raid_count,
         )
         return json_success(
             "Source raid count updated.",
--- a/shuffify/services/raid_sync_service.py
+++ b/shuffify/services/raid_sync_service.py
@@ -125,15 +125,17 @@ class RaidSyncService:
             SchedulerService,
         )
 
-        # Get source before deleting (need playlist_id)
+        # Get source before deleting (need playlist_id). Scoped to the target
+        # playlist as well as the owner: unwatching through a different
+        # playlist's route must not reach this source.
         source = UpstreamSourceService.get_source(
-            source_id, spotify_id
+            source_id, spotify_id, target_playlist_id
         )
         source_playlist_id = source.source_playlist_id
 
         # Delete the upstream source
         UpstreamSourceService.delete_source(
-            source_id, spotify_id
+            source_id, spotify_id, target_playlist_id
         )
 
         if user is None:
--- a/shuffify/services/upstream_source_service.py
+++ b/shuffify/services/upstream_source_service.py
@@ -305,16 +305,22 @@ class UpstreamSourceService:
     @staticmethod
     def update_raid_count(
         user_id: int,
+        target_playlist_id: str,
         source_id: int,
         raid_count: int,
     ) -> UpstreamSource:
         """Update a source's raid_count.
+
+        Scoped to the target playlist as well as the owner: a source is
+        configured against one playlist, so a request naming a different
+        playlist is not addressing this source.
 
         Raises UpstreamSourceNotFoundError if not found.
         """
         source = UpstreamSource.query.filter_by(
             id=source_id,
             user_id=user_id,
+            target_playlist_id=target_playlist_id,
         ).first()
         if not source:
             raise UpstreamSourceNotFoundError(
--- a/shuffify/services/upstream_source_service.py
+++ b/shuffify/services/upstream_source_service.py
@@ -248,7 +248,9 @@ class UpstreamSourceService:
 
     @staticmethod
     def get_source(
-        source_id: int, spotify_id: str
+        source_id: int,
+        spotify_id: str,
+        target_playlist_id: str = None,
     ) -> UpstreamSource:
         """
         Get a specific upstream source by ID.
@@ -256,26 +258,42 @@ class UpstreamSourceService:
         Args:
             source_id: The upstream source database ID.
             spotify_id: The Spotify user ID (for ownership check).
+            target_playlist_id: When given, the source must also be
+                configured against this playlist. Callers reached through a
+                ``/playlist/<playlist_id>/`` route must pass it, so the
+                playlist in the URL constrains which source is addressed
+                rather than being decorative.
 
         Returns:
             UpstreamSource instance.
 
         Raises:
-            UpstreamSourceNotFoundError: If not found or not owned.
+            UpstreamSourceNotFoundError: If not found, not owned, or not
+                configured against ``target_playlist_id``.
         """
         user = get_user_or_raise(
             spotify_id, UpstreamSourceNotFoundError
         )
-        return get_owned_entity(
+        source = get_owned_entity(
             UpstreamSource,
             source_id,
             user.id,
             UpstreamSourceNotFoundError,
         )
+        if (
+            target_playlist_id is not None
+            and source.target_playlist_id != target_playlist_id
+        ):
+            raise UpstreamSourceNotFoundError(
+                f"Source {source_id} not found"
+            )
+        return source
 
     @staticmethod
     def delete_source(
-        source_id: int, spotify_id: str
+        source_id: int,
+        spotify_id: str,
+        target_playlist_id: str = None,
     ) -> bool:
         """
         Delete an upstream source configuration.
@@ -283,16 +301,19 @@ class UpstreamSourceService:
         Args:
             source_id: The upstream source database ID.
             spotify_id: The Spotify user ID (for ownership check).
+            target_playlist_id: When given, the source must also be
+                configured against this playlist.
 
         Returns:
             True if deleted successfully.
 
         Raises:
-            UpstreamSourceNotFoundError: If not found or not owned.
+            UpstreamSourceNotFoundError: If not found, not owned, or not
+                configured against ``target_playlist_id``.
             UpstreamSourceError: If deletion fails.
         """
         source = UpstreamSourceService.get_source(
-            source_id, spotify_id
+            source_id, spotify_id, target_playlist_id
         )
 
         db.session.delete(source)
--- a/tests/routes/test_raid_link_routes.py
+++ b/tests/routes/test_raid_link_routes.py
@@ -214,6 +214,47 @@ class TestRaidLinkSuccess:
         assert data["tracks_added"] == 3
 
 
+def _seed_source(db_app, spotify_id, target_playlist_id, raid_count=5):
+    """Create an upstream source owned by `spotify_id` on `target_playlist_id`."""
+    from shuffify.models.db import UpstreamSource, db
+    from shuffify.services.user_service import UserService
+
+    with db_app.app_context():
+        user = UserService.upsert_from_spotify(
+            {
+                "id": spotify_id,
+                "display_name": spotify_id,
+                "images": [],
+            }
+        ).user
+        source = UpstreamSource(
+            user_id=user.id,
+            target_playlist_id=target_playlist_id,
+            source_playlist_id=f"src-{target_playlist_id}",
+            source_type="external",
+            raid_count=raid_count,
+        )
+        db.session.add(source)
+        db.session.commit()
+        return source.id
+
+
+def _source_raid_count(db_app, source_id):
+    from shuffify.models.db import UpstreamSource, db
+
+    with db_app.app_context():
+        db.session.expire_all()
+        return UpstreamSource.query.get(source_id).raid_count
+
+
+def _source_exists(db_app, source_id):
+    from shuffify.models.db import UpstreamSource, db
+
+    with db_app.app_context():
+        db.session.expire_all()
+        return UpstreamSource.query.get(source_id) is not None
+
+
 class TestRaidSourceCountPlaylistScope:
     """A source belongs to one target playlist (SR-038).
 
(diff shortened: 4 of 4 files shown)


chrisrogers37 (commented):

The fix as submitted is correct and well-verified. Requesting changes because the sibling-route sweep ari asked for found a real, independently-reproduced instance of the exact same defect, unaddressed, one route away.

### 1. Red-first proof — independently reproduced, not re-run

Wrote my own exploit script from scratch (not the PR's test file) against fresh checkouts of `main` (pre-fix) and this branch (post-fix): attacker's own source, reached through a playlist URL they don't own → **200** pre-fix, **404** post-fix, source untouched either way on the legitimate path. Matches alex's transcript exactly, from an independently-built script.

### 2. Severity correction — confirmed correct via real cross-account testing, not code reading

Built two genuinely separate, real authenticated sessions (the PR's own tests mock `require_auth` away entirely and never distinguish two accounts — this doesn't). Attacker's real session, three variations, all targeting a victim's real `source_id`:

| attempt | pre-fix | post-fix |
|---|---|---|
| victim's own playlist_id + victim's source_id | 404, untouched | 404, untouched |
| attacker's playlist_id + victim's source_id | 404, untouched | 404, untouched |
| arbitrary playlist_id + victim's source_id | 404, untouched | 404, untouched |

Cross-account access was blocked in every variation, before the fix too. Alex is right: `user_id` was always in the filter; this closes an in-account scope gap, not a cross-tenant one. P3 was the correct label.

Mutation-tested the fix itself: dropping `target_playlist_id` from the filter fails exactly the 2 scope tests, positive control and cross-user test stay green — matches the PR's own table. 17/17 restored.

### 3. Sibling routes — the important one. Found a live instance of the same bug.

Swept all ~54 routes taking `playlist_id` across every route file (raid_panel.py's other 24, plus playlist_pairs/playlists/track_locks/schedules/playlist_preferences/snapshots/upstream_sources/workshop/shuffle). Every other two-id route (`raid-watch`, `raid-watch-search`, `raid-add-url`, `pending-raids/promote|unpromote|dismiss`, `locks/toggle`, `locks/unlock-all`) correctly threads the URL's scoping id into its query — spot-checked one directly (`add_source`'s dedup filter, `user_id`+`target_playlist_id`+`source_playlist_id`, confirmed byte-for-byte).

**`POST /playlist/<playlist_id>/raid-unwatch` has the identical unscoped shape.** The route passes `target_playlist_id=playlist_id` into `RaidSyncService.unwatch_playlist`, but that method only uses it afterward to patch up an unrelated Schedule row — the actual `get_source`/`delete_source` calls bottom out in `get_owned_entity()` (`services/base.py`), which checks `user_id` only. `target_playlist_id` is fetched, never compared.

Independently reproduced, own script, own seeds, current `main`:

```
owner has a source configured for playlist-A (id=1)
POST /playlist/playlist-B/raid-unwatch  {"source_id": 1}
  -> 200 "Source removed."  — source_on_A no longer exists

control: owner's session, another user's source, same request shape
  -> 404 "Source not found." — untouched (in-account only, consistent with above)
```

Same severity class as SR-038 (bounded by `user_id`, not cross-tenant) — but it's a delete, not an update, so the blast radius per-incident is a silently-vanished raid source rather than a corrupted counter. This PR's own fix pattern (thread `target_playlist_id` into the scoping check) applies directly; `get_owned_entity()` itself is safe to leave alone — its other 3 callers (snapshot/session/schedule) aren't reached from a `playlist_id`-bearing route, so there's nothing for them to drop.

### Bottom line

Ship the `raid-source-count` fix as-is — it's correct, the proof and the severity read both hold up under independent testing. But "SR-038 scope check" isn't done while `raid-unwatch` carries the same hole one route over; recommend folding the identical fix into this PR before merge rather than filing it separately, since the pattern, the test shape, and the mutation check are all already established here.

chrisrogers37:
## Folded in: the same defect on `raid-unwatch`, plus a sweep of the whole route surface

rajan's sibling instance is real and reproduces. Proved before fixing, same method as SR-038 — a source seeded on `playlist-A`, addressed through a different playlist's URL:

| probe | before | after |
|---|---|---|
| `POST /playlist/playlist-B/raid-unwatch` targeting playlist-A's source | **HTTP 200, source DELETED** | HTTP 404, source intact |
| another user's source via the correct playlist | HTTP 404, intact | HTTP 404, intact |

**This one is worse than SR-038 in two ways.** It *deletes* rather than edits, and `unwatch_playlist` then updates the **URL playlist's** raid schedule — so a mismatched call removed a source belonging to playlist A while editing playlist B's schedule, leaving A's schedule still referencing a row that no longer exists. Owner scoping still held, so it stays inside the caller's own account.

Fixed at the service, not the route: `get_source` and `delete_source` take an optional `target_playlist_id` and enforce it, so every caller reached through a `/playlist/<playlist_id>/` route can constrain the lookup. Optional, so the non-playlist callers are unaffected.

## Closing the class rather than the two instances

Two live instances means the pattern is the defect, so here is the whole surface with the method stated, not just a verdict.

**What the class requires:** two identifiers in one request — a playlist id in the *path*, and a caller-supplied **database row key** that the handler resolves. If the resolve is scoped to the owner but not to that playlist, the URL's playlist is decorative. A route carrying only a playlist id cannot exhibit it: there is no second identifier to mismatch.

**Method.** Routes collected by walking the AST of every module in `shuffify/routes/` (not grep — several decorators are multi-line and some handlers carry more than one rule). The AST count was reconciled against a raw `grep -c '\.route('`, and the delta explained rather than absorbed. Row keys were detected as path `<int:*_id>`, `req.*_id`, and `data.get("*_id")`, with `*playlist_id` excluded as a Spotify id rather than a row key. Script: `enumerate2.py` (in the PR discussion below if wanted).

| | count |
|---|---|
| raw `.route(` occurrences (grep) | 87 |
| **actual routes** (AST) | **86** — the 1 delta is a `.route(` inside a docstring at `routes/__init__.py:151`, not a route |
| take a playlist id in the path | 54 |
| **…and also a caller-supplied row key → the class** | **2** |
| playlist routes with no row key (class not reachable) | 52 |
| routes with no playlist id in the path | 32 |

**The class is exactly 2 routes, and both are in this PR:** `raid-source-count` and `raid-unwatch`.

**Why so few — checked, not assumed.** The only DB row key any request schema exposes is `source_id`; every other `*_id` in `shuffify/schemas/` is a Spotify identifier (`source_playlist_id`, `archive_playlist_id`, `raid_playlist_id`, `target_playlist_id`), which addresses content rather than a row that could belong to another playlist. The four path-level row keys (`schedule_id`, `session_id`, `snapshot_id`, `source_id`) all appear on routes with **no** playlist in the path, so there is no second identifier to disagree with. The 25 playlist routes that do carry Spotify-side ids are listed in the script output so that exclusion is auditable rather than silent.

**Residual I did not check, stated plainly:** the four path-level row-key routes (`/snapshots/<int:snapshot_id>`, `/schedules/<int:schedule_id>`, `/upstream-sources/<int:source_id>`, `/workshop/sessions/<int:session_id>`) raise a *different* question — whether each row is scoped to the caller at all. That is the ownership question, not this playlist-scope one, and it is outside what was asked here. Worth its own sweep; two of those endpoints are destructive.

## Verification

- `2070 passed, 1 skipped`, 0 failed. `flake8 shuffify/` clean.
- Seven regression tests across the two routes, each with a **positive control** that the correct playlist still works — so the scope tests cannot pass by the route being broken in general.
- **Mutation-checked both fixes.** Removing the `target_playlist_id` filter fails exactly the scope tests; the positive controls and the cross-user tests keep passing.
- The per-class seed helpers were collapsed into module-level ones shared by both test classes.

chrisrogers37 (commented):

Tested the enumeration hardest, per the dispatch, since it's what the "class closed" claim rests on.

## 86 vs 87 — confirmed directly

`grep -c '\.route('` on `shuffify/routes/*.py` → 87, independently. Line 151 of `routes/__init__.py` is inside a `Usage::` docstring block for the `require_auth_and_db` decorator (`@main.route("/endpoint")` as an example) — not a real route. Wrote my own AST walker from scratch (not his script) rather than trust the number: **86**, matching exactly, with a per-file breakdown that sums correctly.

## The premise — mostly confirmed, but I found a real gap in the search method

Checked "`source_id` is the only DB row key, everything else is a Spotify id" two ways:
- **Schema field types**: every `*_id` field across `shuffify/schemas/*.py` is `str`-typed (`archive_playlist_id`, `raid_playlist_id`, `target_playlist_id`, `source_playlist_id`) except `source_id: int`, which appears in exactly 2 schemas. Matches.
- **Path parameters**: exactly 4 `<int:*_id>` converters exist anywhere in `shuffify/routes/` — `schedule_id`, `session_id`, `snapshot_id`, `source_id` — and `<playlist_id>` is the only untyped (string) one. Matches his 4-routes-with-no-playlist-id claim exactly.

**But the pattern search (his and my first pass) only looks for singular `*_id`.** Broadening to `List[int]` turned up `track_ids` in `PromoteTracksRequest`/`DismissTracksRequest` (`pending_raid_requests.py`) — a real additional row-key surface, used on **3** routes that *do* carry a playlist id in the path (`/playlist/<playlist_id>/pending-raids/{promote,dismiss,finalize}`) — the exact shape of the class. I chased it: `PendingRaidService.promote_tracks` and `dismiss_tracks` already filter on `PendingRaidTrack.id.in_(track_ids)` **and** `user_id` **and** `target_playlist_id` together, at every one of the 3 call sites, including the inline query in the dismiss route itself. Already safe, pre-existing code, untouched by this PR.

So: the *premise* survives — I didn't find an unfixed instance — but the *method* ("search for `*_id`") has a real blind spot for plural `_ids: List[int]` fields, and this time it happened to land somewhere already correct rather than somewhere vulnerable. Worth stating plainly rather than letting "checked, not assumed" imply the pattern search itself was complete — it wasn't, luck did some of the work here.

## No caller bypasses the fix

`get_source`/`delete_source`/`update_raid_count`: enumerated every call site. The one caller that doesn't pass `target_playlist_id` (`routes/upstream_sources.py`'s `DELETE /upstream-sources/<int:source_id>`) is correctly on a route with no playlist in the path at all — nothing to scope against, matches the documented "optional, non-playlist callers unaffected" design.

## The 4 deferred routes — my read, since you asked

Spot-checked all four (`snapshot_id`, `schedule_id`, `source_id` DELETE, `session_id`): each passes `user.id`/`user.spotify_id` into a service method that does owner-scoping (`get_owned_entity` or an equivalent explicit check) — not raw, unscoped lookups. The deferred question is real and worth its own sweep, but I don't see a demonstrated hole here, just an unaudited surface. Wouldn't block this PR on it.

## Rebase — verified intact

`tests/test_db_schema_guard.py` is byte-identical to `origin/main`. `docker-entrypoint.sh`, `shuffify/__init__.py`, `config.py` (the rest of #503) show zero diff against `origin/main`. Nothing lost in the cleanup.

## Mutation-tested independently

Removed the `target_playlist_id` scope check from `get_source` entirely: `test_other_playlist_cannot_delete_the_source` fails (200, expected 404) — everything else stays green, confirming the test binds to this specific check. Restored: 20/20 pass.

Two real defects, correctly fixed at the layer that closes the class for both current instances, verified independently rather than taken on the report — with one honest asterisk on the search method that didn't change the outcome this time.

PR #507, opened 2026-08-01
security: escape Spotify-supplied names at the render path (#465 Finding 1)

### LINKED ISSUE #465: [security] Unescaped HTML interpolation and dependency-floor gaps — schedules.html, workshop.html, requirements/base.txt
## Summary

Three independent findings, batched because each is a defense-in-depth gap rather than a directly exploitable critical path, and each has a small, mechanical fix.

**Finding 1 — Unescaped HTML interpolation (`schedules.html`).** Three spots in `shuffify/templates/schedules.html` interpolate dynamic strings into `innerHTML` without calling the file's own `escapeHtml()` helper, breaking a convention the codebase otherwise follows consistently (`workshop.html` escapes every equivalent field). One of the three values (`source_name` on a raid source) is reachable from a real Spotify playlist title, which can be set by any Spotify account holder — not just the victim. The app's strict CSP (no `unsafe-inline` in `script-src`/`style-src`) and `HttpOnly` session cookie both block the highest-impact outcomes (script execution, cookie theft), leaving HTML/phishing-link injection as the residual risk. An adversarial review pass found the premise that `workshop.html` 'escapes every equivalent field... consistently' does not fully hold: `workshop.html:5341-5348` (`renderHubDropdown()`) interpolates `p.name` into text content and `p.image_url`/`p.id` into attribute values with no escaping at all — the same missing-escape defect, in the same file this doc cites as the compliant example, confirmed by direct inspection. A second, more subtle gap also confirmed: `escapeHtml()` itself only correctly escapes for *text-content* context (per the WHATWG HTML fragment-serialization algorithm, a Text node's serialization escapes `&`, `<`, `>` but not `"` or `'`) — several `workshop.html` call sites (e.g. lines ~1409-1422, 2311-2312) already call `escapeHtml()` *inside quoted HTML attributes* (`src="${escapeHtml(p.image_url)}"`), where it does not fully close the injection vector. Both gaps are addressed in the Steps below; this doc's scope has been widened accordingly.

**Finding 3 — Two dependency floors sit one patch version short of the CVE they're meant to clear.** `requirements/base.txt` has a section explicitly dedicated to "explicit floors for transitive deps with known CVEs." Two of the ~10 packages in that section — `tornado>=6.5.5` and `nltk>=3.9.3` — are floored exactly at a version `safety` still lists as insecure; the actual fix landed one patch release later. The currently-installed versions in this environment already happen to be secure (pip's resolver picks the newest match by default), so this isn't a live exploit today, but the floor itself doesn't yet deliver the guarantee its own comment claims.

A third finding from this audit pass — the Dockerfile running as root (missing `USER` instruction) — is tracked separately: it's an exact-match refresh of existing issue #326, not duplicated here.

### Finding 1: Unescaped interpolation

`shuffify/templates/schedules.html` defines its own escaping helper:

```javascript
// schedules.html:325-329
function escapeHtml(str) {
    var div = document.createElement('div');
    div.appendChild(document.createTextNode(str));
    return div.innerHTML;
}
```

`shuffify/templates/workshop.html` (a separate copy of the same helper at `workshop.html:2337`) calls it consistently for every dynamic field it renders via `innerHTML` — e.g. `workshop.html:1409-1422` (`escapeHtml(p.name)`, `escapeHtml(p.image_url)`, `escapeHtml(p.id)`) and `workshop.html:2311-2312` (`escapeHtml(track.name)`, `escapeHtml(artists)`).

`schedules.html` does not follow its own convention in three places:

```javascript
// schedules.html:466-474 — renderRaidSources()
container.innerHTML = sources.map(src => `
    <label class="flex items-center space-x-2 px-2 py-1 rounded hover:bg-white/5 cursor-pointer">
        <input type="checkbox" name="source_playlist" value="${src.source_playlist_id}" ...>
        <span class="text-white/80 text-sm truncate">${src.source_name || src.source_playlist_id}</span>
        ...
    </label>
`).join('');
```

```javascript
// schedules.html:501-511 — renderPairStatus()
container.innerHTML = `
    <div class="rounded-lg bg-green-500/10 border border-green-500/30 p-3">
        ...
        <p class="text-green-200 text-xs">
            Paired with <strong>${pair.archive_playlist_name || 'Archive'}</strong>
        </p>
    </div>`;
```

```javascript
// schedules.html:719-725 — loadHistory()
list.innerHTML = data.history.map(h => {
    ...
    const error = h.error_message ? `<span class="text-red-300 block mt-0.5">${h.error_message}</span>` : '';
    return `...`;
}).join('');
```

**Data provenance — why this isn't purely self-XSS:**

- `src.source_name` traces back to `UpstreamSource.source_name`, which is set two ways:
  - `shuffify/routes/upstream_sources.py:60-89` (`add_upstream_source`) — taken directly from the caller's own JSON body (`data.get("source_name")`), no HTML sanitization, only implicit via whatever the frontend pre-fills.
  - `shuffify/routes/raid_panel.py:299-352` (`raid_watch`) — `source_playlist_name=req.source_playlist_name`, validated only for length (`shuffify/schemas/raid_requests.py:38-50`, max 255 chars, no content sanitization). This field is populated by the frontend from a **real Spotify playlist's title** the user is choosing to raid from — a playlist that can be owned and named by *any* Spotify account holder, not just the Shuffify user viewing the page. A malicious Spotify user can name a public playlist with an HTML payload; any Shuffify user who later raids that playlist will have the raw name written into their own `UpstreamSource.source_name` and rendered unescaped on their own `/schedules` page.
  - Both paths reach `UpstreamSourceService.add_source` (`shuffify/services/upstream_source_service.py`), which persists the string as-is.
- `pair.archive_playlist_name` traces back to `shuffify/routes/playlist_pairs.py:77-88` (`create_pair`) — either machine-generated (`create_private_playlist`, safe) or `req.archive_playlist_name` taken directly from the request body when linking an existing playlist, again only length-validated (`shuffify/schemas/playlist_pair_requests.py:25,38-46`, 1-255 chars).
- `h.error_message` traces back to `JobExecution.error_message`, populated from `str(exception)[:1000]` in `shuffify/services/executors/base_executor.py` (e.g. `_record_failure`, `_persist_rollback_status`). Attacker control here is indirect (would require influencing an underlying exception's string representation) and lower-probability than the two paths above, but it shares the same missing-escape root cause and the same fix.

**Mitigating controls already in place** (why this is MEDIUM, not CRITICAL/HIGH):
- `shuffify/__init__.py:389-401` sets a nonce-based CSP with no `unsafe-inline` in `script-src` or `style-src` — injected `<script>` tags and inline event-handler attributes (`onerror=`, `onclick=`, etc.) will not execute in a modern browser, and inline `style="..."` attributes are likewise blocked.
- `config.py:71,126` sets `SESSION_COOKIE_HTTPONLY = True` — even a successful script execution could not read the session cookie via `document.cookie`.
- The residual exploitable surface is HTML *structure* injection (e.g., an injected `<a href="https://phishing-site.example">`) rendered inside the victim's own authenticated page — a phishing/defacement primitive, not session takeover.

### Finding 3: Dependency floors short of the secure line

`requirements/base.txt` (lines reproduced from the earlier read of this file):

```text
# --- Security: explicit floors for transitive deps with known CVEs ---
# These packages are pulled in by Flask, requests, cryptography, etc.
# pip will NOT upgrade them on its own unless we set a minimum version.
werkzeug>=3.1.8
urllib3>=2.7.0
certifi>=2026.5.20
MarkupSafe>=3.0.3
PyYAML>=6.0.3
pyasn1>=0.6.3
wheel>=0.47.0
authlib>=1.7.2
nltk>=3.9.3
tornado>=6.5.5
```

`safety check -r requirements/base.txt -r requirements/prod.txt -r requirements/dev.txt --json` (run 2026-07-15, `safety` 3.7.0) reports `vulnerabilities_found: 0`, but that headline number is misleading on its own — it excludes 15 matches (5 unique CVEs × 3 requirements files, since `prod.txt`/`dev.txt` both `-r base.txt`) that `safety` skips by default under its `ignore-unpinned-requirements` policy, specifically because `>=` floors don't tell the scanner which exact version will actually be installed:

| Package | Floor in `base.txt` | CVE(s) | `safety`'s insecure/secure boundary |
|---|---|---|---|
| `tornado` | `>=6.5.5` | CVE-2026-49854 | insecure through `6.5.5`; secure at `6.5.6`, `6.5.7` |
| `nltk` | `>=3.9.3` | CVE-2026-33230, CVE-2026-33231, SFTY-20260320-58622 | insecure through `3.9.3`; secure at `3.9.4`, `3.10.0` |
| `apscheduler` | `>=3.11.2` | CVE-2026-31072 | insecure only pre-3.0 alphas; `3.11.2` itself is already in `safety`'s secure list — **this one is already correctly floored, no action needed** |

For `tornado` and `nltk`, the floor value in the requirements file is the exact version `safety`'s database still lists as vulnerable — the fix landed one patch release later in each case. Confirmed via `pip list --outdated` in this environment that the *actually resolved* versions are already safe (`nltk` 3.9.4 installed; `tornado` not present in the outdated list at all, meaning the latest secure release is already resolved) — pip's resolver picks the newest version satisfying `>=`, so this is not a live vulnerability in normal operation. It is, however, a gap against the file's own stated purpose ("pip will NOT upgrade them on its own unless we set a minimum version") — an install path that legitimately lands on the exact floor (private package mirror snapshot, constrained lockfile, an unusual resolver backtrack) would not be protected.

### Dependencies

None.

### Blocks

None.

### Steps

**Finding 1 — fix in `shuffify/templates/schedules.html`:**

1. Line 471 — wrap the interpolated value:
   ```javascript
   // before
   <span class="text-white/80 text-sm truncate">${src.source_name || src.source_playlist_id}</span>
   // after
   <span class="text-white/80 text-sm truncate">${escapeHtml(src.source_name || src.source_playlist_id)}</span>
   ```
2. Line 508 — wrap the interpolated value:
   ```javascript
   // before
   Paired with <strong>${pair.archive_playlist_name || 'Archive'}</strong>
   // after
   Paired with <strong>${escapeHtml(pair.archive_playlist_name || 'Archive')}</strong>
   ```
3. Line 725 — wrap the interpolated value:
   ```javascript
   // before
   const error = h.error_message ? `<span class="text-red-300 block mt-0.5">${h.error_message}</span>` : '';
   // after
   const error = h.error_message ? `<span class="text-red-300 block mt-0.5">${escapeHtml(h.error_message)}</span>` : '';
   ```
4. `escapeHtml` is already defined at `schedules.html:325-329` in the same file — no new helper or import needed.
5. Grep the file after the change to confirm no other un-escaped `${...}` interpolations remain inside an `innerHTML =` assignment: `grep -n "innerHTML" shuffify/templates/schedules.html`.
6. See step 8 below — this is no longer optional. `escapeHtml` is defined identically in both `schedules.html:325-329` and `workshop.html:2337`. Consider extracting it to `shuffify/static/js/notifications.js` (already loaded on every page via `base.html:197`) or a new small shared `static/js/` file, so future new templates inherit the safe-by-default helper instead of needing to redefine or remember it.
7. `workshop.html:5341-5348` (`renderHubDropdown()`) — wrap all three interpolated values:
   ```javascript
   // before
   list.innerHTML = filtered.map(p => `
       <a href="/workshop/${p.id}"
          class="flex items-center gap-3 px-3 py-2 rounded-lg hover:bg-white/10 transition duration-150 cursor-pointer group">
           <img src="${p.image_url || '/static/images/placeholder.svg'}"
                alt="" class="w-10 h-10 rounded object-cover flex-shrink-0 bg-white/5">
           <div class="min-w-0">
               <div class="text-white text-sm font-medium truncate group-hover:text-white/90">${p.name}</div>
               <div class="text-white/40 text-xs">${p.track_count} tracks</div>
           </div>
   // after
   list.innerHTML = filtered.map(p => `
       <a href="/workshop/${escapeHtml(p.id)}"
          class="flex items-center gap-3 px-3 py-2 rounded-lg hover:bg-white/10 transition duration-150 cursor-pointer group">
           <img src="${escapeHtml(p.image_url || '/static/images/placeholder.svg')}"
                alt="" class="w-10 h-10 rounded object-cover flex-shrink-0 bg-white/5">
           <div class="min-w-0">
               <div class="text-white text-sm font-medium truncate group-hover:text-white/90">${escapeHtml(p.name)}</div>
               <div class="text-white/40 text-xs">${p.track_count} tracks</div>
           </div>
   ```
   (`p.track_count` is a number, not attacker-controlled free text — no wrap needed.)
8. **Mandatory, not optional:** given the `escapeHtml()` attribute-context gap (see Summary above), replace the naive helper everywhere it's used inside a quoted attribute with one that also encodes `"`/`'` — either (a) a second helper (e.g. `escapeHtmlAttr`) that additionally replaces `"` → `&quot;` and `'` → `&#39;` after the existing text-node-based escape, or (b) switch those call sites to build the element via `document.createElement` + `.setAttribute()` instead of string-templated `innerHTML`, which is attribute-context-safe by construction. Apply to every `workshop.html` site currently calling `escapeHtml()` inside a `src=`/`data-*=`/`href=`/similar attribute (grep `escapeHtml(` in `workshop.html` to enumerate them — roughly 7 sites per the adversarial review, e.g. lines ~1409-1422, 2301, 2399, 2526, 2712, 3888). This was previously scoped as "optional cleanup" in this doc's step 6 above; promoted to mandatory because the review found it isn't just theoretical duplication — the existing helper is already being used unsafely, today, in the file this doc calls the compliant example.

**Finding 3 — fix in `requirements/base.txt`:**

1. Bump the two floors by one patch version each, in the existing "explicit floors for transitive deps with known CVEs" block:
   ```diff
   - nltk>=3.9.3
   - tornado>=6.5.5
   + nltk>=3.9.4
   + tornado>=6.5.6
   ```
   (`tornado>=6.5.7` is also acceptable — it's the latest known-secure release at scan time; `6.5.6` is the minimum that clears CVE-2026-49854.)
2. Leave `apscheduler>=3.11.2` as-is — it already clears CVE-2026-31072, no change needed.
3. Re-run `safety check -r requirements/base.txt -r requirements/prod.txt -r requirements/dev.txt --json` (or `safety scan`, since `check` is deprecated upstream — see note below) and confirm `tornado`/`nltk` no longer appear in `affected_packages` with an insecure floor.

## Test Plan

**Finding 1:**
- No existing automated test covers HTML escaping in `schedules.html`'s JS rendering functions (this is client-side JS, outside `pytest`'s reach — confirmed no `tests/templates/` coverage of `renderRaidSources`/`renderPairStatus`/`loadHistory`).
- Manual verification: create an upstream source (or use `raid_watch`) with `source_name` set to ``, confirm it renders as literal text (`&lt;img src=x onerror=alert(1)&gt;`) on `/schedules`, not as a DOM element, by inspecting the rendered `<span>` in browser devtools.
- Repeat for an archive pair name and a forced job-execution failure message, confirming both render as literal text.
- If the project ever adds a JS test harness (none currently exists per `documentation/`), add unit tests asserting `renderRaidSources`/`renderPairStatus`/`loadHistory` HTML-encode their inputs.

**Finding 3:**
- Re-run the `safety check` command above after the version bump; confirm `affected_packages` no longer lists an insecure floor for `tornado`/`nltk`.
- `pip install -r requirements/dev.txt` in a clean virtualenv, then `pip show tornado nltk | grep -i version` — confirm resolved versions are `>=6.5.6` / `>=3.9.4` respectively.
- Run the existing test suite (`pytest tests/ -v`) to confirm the version bump doesn't break anything — both are transitive dependencies (tornado via a dev/tooling path, nltk via base), so no direct shuffify code should be affected.

## Verification Checklist

- [ ] `schedules.html:471` wraps `src.source_name || src.source_playlist_id` in `escapeHtml(...)`
- [ ] `schedules.html:508` wraps `pair.archive_playlist_name || 'Archive'` in `escapeHtml(...)`
- [ ] `schedules.html:725` wraps `h.error_message` in `escapeHtml(...)`
- [ ] Manual XSS-payload test against a raid source name renders as literal text, not markup
- [ ] `workshop.html:5341-5348` wraps `p.id`, `p.image_url`, and `p.name` in `escapeHtml(...)`
- [ ] Every `escapeHtml()` call site inside a quoted HTML attribute in `workshop.html` (~7 sites) uses an attribute-safe escape, not the text-content-only helper
- [ ] `requirements/base.txt` floors `nltk>=3.9.4` and `tornado>=6.5.6` (or newer)
- [ ] `safety check` no longer lists an insecure floor for either package
- [ ] `pytest tests/ -v` passes after the dependency bump
- [ ] `CHANGELOG.md` updated per repo convention (`### Fixed` / `### Security` entries)

## What NOT To Do

- **Do not "fix" Finding 1 by sanitizing at write time** (e.g., stripping HTML tags from `source_name`/`archive_playlist_name` in the Pydantic validators before storage). Input sanitization is fragile — it has to anticipate every payload shape, breaks legitimate names containing `<`/`>`/`&`, and does nothing for the `h.error_message` path, which is never validated at the point tokens are stored. Escape at render time, where the security boundary actually is.
- **Do not add `unsafe-inline` to the CSP** to "make errors go away" if the escaping fix surfaces some other rendering quirk. The nonce-based CSP is a major reason Finding 1 is MEDIUM and not HIGH; weakening it to patch something else would be a net regression.
- **Do not pin `tornado`/`nltk` with `==` exact pins** as a way to "lock in" the fix for Finding 3. The existing floor (`>=`) pattern is intentional per the file's own comment — it lets pip pick up future patch releases automatically. Bump the floor, don't change the operator.
- **Do not treat `vulnerabilities_found: 0` from `safety check` as the complete picture** without also checking `affected_packages`/`ignored_vulnerabilities` in the JSON output. The default `ignore-unpinned-requirements` policy silently excludes anything behind a `>=` specifier, which is most of this project's dependency floors.

## Context

- Source skill: claudna:audit (security lens)
- Audit date: 2026-07-15
- Area: `shuffify/templates/schedules.html`, `requirements/base.txt`
- Effort: Low
- Risk: Medium
- Priority: Medium
- Related issues: #326 (Docker root — refreshed via a companion doc from this same audit, not duplicated here), #344 (dependency hygiene — different aspect, cross-referenced not merged)

## Noteworthy LOW/INFO items (folded in per audit scope — not independently actionable enough for their own doc)

- **Rate-limit test coverage gap.** `shuffify/__init__.py:486-511` applies rate limits to five endpoints (`/login`, `/callback`, `/shuffle`, `/workshop/commit`, `/schedules/<id>/run`), but `tests/test_rate_limiting.py` only exercises `/login` and `/callback`. The controls are present and correctly configured; only the test coverage is incomplete. Low priority — add `TestShuffleRateLimit`/`TestWorkshopCommitRateLimit`/`TestScheduleRunRateLimit` classes mirroring the existing `TestLoginRateLimit` pattern if this repo is touched for other reasons.
- **CSP third-party script allowlist.** `shuffify/__init__.py:391-392` allows `https://cdn.tailwindcss.com` and `https://cdn.jsdelivr.net` in `script-src`, alongside `'self'` and the per-request nonce. This is a standard, low-risk supply-chain trust-widening (a compromise of either CDN could inject script into the page) but is a common and reasonable trade-off for avoiding a local Tailwind build step. No action recommended unless the project moves to a bundled/local Tailwind build, at which point these origins could be dropped.
- **`safety check` is deprecated upstream** (own output: "this command (`check`) has been DEPRECATED, and will be unsupported beyond 01 June 2024"). `pip-audit` was unavailable in this environment (neither `pip-audit` nor `pip audit` resolved) so `safety` — already a project dev-dependency (`requirements/dev.txt`) — was used as the category-A scanner instead. Worth switching the `check` invocation to `safety scan` next time this is run; behavior/output shape differs slightly.

## Adversarial Review Findings

These concerns were raised by a 4-reviewer adversarial review pass (Architect, Skeptic, Operator, User+Counter-Planner) at plan-creation time, before publishing. Items already folded into the Summary/Steps/Verification Checklist above are marked accordingly; the rest are open for the implementer to weigh.

- [x] **[major] security**: Finding 1's premise that `workshop.html` "escapes every equivalent field consistently" is false — two more unescaped/attribute-unsafe `innerHTML` sites exist in that same file (`renderHubDropdown()` at :5341-5348, unescaped text + attributes; `renderSnapshotTimeline()` at :3038/:3072, a hand-rolled partial escape that only covers attribute-breakout, not element text content).
  - **Recommendation:** Folded into Summary/Steps/Verification Checklist above (Edits 1-2, 4).
- [x] **[major] security**: `escapeHtml()` only correctly escapes for text-content context (WHATWG Text-node serialization escapes `&`/`<`/`>` only, not `"`/`'`); `workshop.html` already calls it inside quoted attributes (`src="${escapeHtml(...)}"`) at ~7 sites, where it does not fully close the injection vector.
  - **Recommendation:** Folded into Steps above (Edit 2, step 8) as a mandatory (not optional) fast-follow within this same fix, not a future cleanup.
- [x] **[major] error-handling**: The Verification Checklist has no check capable of catching Finding 2's most likely failure mode — `/health` only checks DB connectivity, never a session write. A `USER nobody` regression that breaks filesystem-session writes would pass every existing checklist item while breaking login in production.
  - **Recommendation:** Folded into Test Plan/Verification Checklist above (Edits 3-4) as a mandatory real login-flow smoke test.
- [ ] **[major] dependencies**: Two independent reviewers, checking two different environments, found `tornado` actually installed at `6.5.5` (the exact insecure floor) right now — contradicting this doc's own Evidence claim that "the actually resolved versions are already safe." Doesn't change whether the proposed fix works, but the "not urgent" comfort framing in Finding 3 rests on an unverified/contradicted claim.
  - **Recommendation:** Re-run `pip list --outdated`/`pip show tornado` in a genuinely fresh, isolated install before merging, and soften or remove the "not a live vulnerability in normal operation" framing if it doesn't reproduce cleanly. Treat Finding 3 as no less urgent than currently scoped, possibly more.
- [ ] **[minor] observability**: This repo's `CLAUDE.md` documents a full CI/CD pipeline (lint, tests, E2E, GitGuardian, "must pass on push to main") that does not exist — confirmed via `find .github` (only `dependabot.yml`, no `.github/workflows/`) and corroborated by the repo's own `documentation/guides/infrastructure_critiques.md`, which lists "No CI/CD Configuration" as a still-open item. This means every Test Plan step in this doc ("run pytest", "run safety check") is manual-only with nothing enforcing it before merge.
  - **Recommendation:** Not a blocker for this PR, but a high-leverage fast-follow: a CI-gated `safety scan`/`pip-audit` run would have made the Finding 3 discrepancy above self-detecting instead of requiring a manual audit sweep, and would catch future recurrences (this exact pair of packages has had its floor bumped once before, per `CHANGELOG.md`, for different CVEs).
- [ ] **[minor] dependencies**: GitHub Dependabot Security Alerts (already active on this repo, 19 historical alerts, all real and fixed) do not include `tornado` or `nltk` — confirmed via `gh api repos/chrisrogers37/shuffify/dependabot/alerts`. Dependabot and `safety` have non-overlapping vulnerability-database blind spots on this exact finding.
  - **Recommendation:** Don't rely on Dependabot alone for this class of gap; the CI-gated scanner recommendation above should specifically not assume Dependabot already covers it.
- [ ] **[minor] security**: Finding 1's severity rationale ("phishing/defacement primitive... an injected `<a href=phishing-site>`") undersells the CSP-compatible blast radius. The current CSP is confirmed as strict as claimed (no `unsafe-inline`, `frame-ancestors 'none'`, `base-uri 'self'`, `form-action 'self'` — clickjacking via iframe is closed on two independent axes), but injected HTML structure alone (no script, no new inline style) can still produce (a) an automatic zero-click redirect via `<meta http-equiv="refresh">` (not governed by any CSP directive), and (b) a full-page fake overlay/UI-redress using the site's own already-loaded Tailwind classes (CSP doesn't restrict which existing class names get reused).
  - **Recommendation:** Doesn't change the MEDIUM severity conclusion (no persistent script execution, no cookie theft), but worth a one-line addition to the Summary if this doc is revised again, so a reader isn't left thinking the residual risk is capped at "a suspicious link."
- [ ] **[minor] architecture**: A serious counter-plan was developed: split this batched doc into 3 independent issues (XSS/frontend, Docker/infra, dependency floors) instead of one combined PR, backed by this repo's own git history showing no precedent for cross-domain batching (existing multi-issue batches are always within one lens — e.g. all security-hardening, or all config-cleanup — never frontend+infra+deps together) and the doc's own admission of zero dependency/blocking between the three findings.
  - **Recommendation:** Non-blocking judgment call for the human/implementer — the batched doc is fine to keep as-is (all three fixes are correct and low-effort), but if review bandwidth becomes a bottleneck, splitting is a legitimate, repo-precedented alternative, not just generic advice.
- [ ] **[info] observability**: Minor citation nits found during verification, none affecting the fixes' correctness: the Dockerfile's `chown`/`chmod` block is lines 28-30, not 27-30 as cited (line 27 is a comment); the apscheduler CVE-2026-31072 affects versions `4.0.0a1`-`4.0.0a6` specifically, not broadly "pre-3.0 alphas"; the archive-playlist-name "safe" path uses `PlaylistPairService.create_archive_playlist(...)`, not `create_private_playlist` as stated; migrations run inside the gunicorn `--preload` master process on every boot (via `_init_database`), not via a separate `flask db upgrade` step as the Dockerfile fix's rationale states (this doesn't change whether the fix works, since the broader `chown -R /app` step covers this path too).
  - **Recommendation:** Fix opportunistically if this doc is touched again; none are load-bearing.
- [ ] **[info] observability**: `requirements/base.txt:11` pins `psycopg2-binary>=2.9.12` while `requirements/prod.txt:6` separately pins `psycopg2>=2.9.9` on top of it (source-build variant) — likely unintentional drift, not verified end-to-end which package's files actually win in `site-packages`. Unrelated to any of this doc's 3 findings but adjacent (it's the reason the Dockerfile still needs `build-essential`/`libpq-dev`).
  - **Recommendation:** Worth a maintainer look in a separate, unrelated issue — not folded into this doc's scope.

**Murphyjitsu verdict:** Given the fixes above (widened Finding 1 scope, mandatory attribute-safe escaping, mandatory login-flow smoke test), would a reviewer be surprised if this plan, as now written, failed in an unanticipated way? No single remaining open item (dependency floor re-verification, missing CI, severity-framing precision, the counter-plan) is likely to cause the *shipped fix* to fail — they're about completeness of context and prevention of recurrence, not correctness of the 3(+2) code changes themselves, all of which were independently verified against the live repository by all four reviewers with zero Blockers on the code itself.

chrisrogers37:
Closes #465 (Finding 1). **Finding 3 needs no action** — the dependency floors are already fixed at `main` (`nltk>=3.10.0`, `tornado>=6.5.7`, both above the boundary the issue names). Only the injection half is addressed here.

## The vector is live — confirmed in a browser, not by inspection

A raid source's `source_name` is the **title of the Spotify playlist being raided**. Any Spotify account holder can name a public playlist; when a Shuffify user raids it, that title is persisted to their `UpstreamSource` and rendered on their own `/schedules` page. Stored, not reflected, and cross-user.

I stood up the real app against a real DB, seeded an `UpstreamSource` whose `source_name` carried a payload, and loaded the **real `/schedules` route** in Chromium with the app's **real CSP** headers. Only the Spotify API boundary was stubbed — which is exactly the boundary the attacker controls. Everything downstream (DB persistence, `to_dict()`, `| tojson` into the page, the template's own JS) is the shipping path.

| probe | before | after |
|---|---|---|
| `` element created in victim's page | **yes** | no |
| `<a href="https://phishing.example/">` created | **yes** | no |
| injected attribute on the rendered checkbox | **yes** (`data-injected-attr="yes"`) | no |
| injected `onerror` script executed | no — **CSP blocked it** | no |

After the fix the payload renders as inert text (`&lt;img src=x ...&gt;`) and the attribute value stays quoted (`abc&quot; data-injected-attr=&quot;yes`).

**The CSP holds.** The injected `onerror` did not run, and the browser logged `Refused to execute inline event handler...`. So this is HTML-structure and phishing-link injection inside an authenticated page — not session takeover. That matches the issue's MEDIUM rating; I'm not inflating it.

## Two defect classes, not one

**1 — values written raw into `innerHTML`.** `schedules.html` (raid-source name, archive-pair name, job error message, and `source_playlist_id` in an attribute) and `workshop.html`'s `renderHubDropdown` (playlist `name`, `id`, `image_url`).

**2 — `escapeHtml()` used inside a quoted attribute.** It escapes for *text-node* context only: per the HTML fragment serialization algorithm a text node encodes `&`, `<`, `>` but leaves `"` and `'` alone. Nine call sites already called it inside quotes, where it cannot close the vector — demonstrated above, where a crafted `source_playlist_id` added a whole new attribute to the checkbox. Added `escapeAttr()` (text escape + `&quot;`/`&#39;`) and moved every in-attribute site onto it.

## On coverage — what I actually scanned

My first sweep looked only at `${...}` interpolations and reported 4 problems. That was wrong: these templates also build markup by **string concatenation**, which that scan never saw. Re-sweeping both forms took the scanned surface from 75 sites to 207 and surfaced `static/js/schedule-edit.js`, which the first pass scored as clean.

Final state across `shuffify/templates/*.html` + `shuffify/static/js/*.js`: **0** `escapeHtml()` calls remain inside a quoted attribute, and every Spotify-supplied value reaching `innerHTML` is escaped.

Deliberately **not** changed — checked and found not to be sinks:
- `showNotification()` assigns `textContent`, so `workshop.html:2380` / `2496` are safe despite passing raw Spotify names.
- `index.html`'s track names are hardcoded demo data for the landing animation, not Spotify data.
- ~31 raw-in-attribute interpolations remain, all app-generated (`playlistId`, `targetId`, `statusColor`, `scheduleId`, `raidCount`). Not attacker-settable; left alone rather than padding the diff.

## Regression guard

`tests/test_template_escaping.py` asserts both invariants as source-level gates (the project has no JS test harness). It is load-bearing, not decorative — reverting either fix class fails it:

```
FAILED test_no_text_escape_inside_quoted_attribute[workshop.html]
FAILED test_spotify_supplied_field_is_escaped[...src.source_name || src.source_playlist_id]
```

## Checks

`pytest -m "not integration"` → **1969 passed**. `flake8 shuffify/` → clean. Both templates re-verified through Jinja parse. `black --check shuffify/` still reports its pre-existing 53-file backlog (non-blocking per SR-037); this PR touches no `.py` file under `shuffify/`, so that number is unchanged by it.

## One note for whoever consolidates the helper

The issue suggests hoisting `escapeHtml` into a shared file so new templates inherit it. I measured the load order before trying: `notifications.js` is included at `base.html:166` but lands at **line 1084 of the rendered `/schedules` document — after the page's inline script at line 647**. A shared-file extraction would therefore introduce a real load-order dependency, so I kept the helpers local (matching where `escapeHtml` already lives) rather than smuggle that change into a security fix. Worth its own PR with that ordering in mind.

Not merging my own work.


chrisrogers37 (commented):

Checked the harness and the gap, not just the result. Findings below; the one real gap is named plainly per the dispatch, not assumed covered.

## The gap — real at the code level, empirically not exploitable today

7 `` sites carry Spotify-supplied image URLs with **no scheme allowlist**: `renderPlaylistDropdown`, `createSearchResultElement`, `createTrackElement`, `displayExternalSearchResults`, `renderPendingRaids`, `renderInboxArchiveTracks`, `renderHubDropdown` (all `workshop.html`). `escapeAttr` correctly blocks breakout; it does nothing to a scheme. One `<a href>` site exists (`/workshop/${escapeAttr(p.id)}`) but is safe by construction — the hardcoded `/workshop/` prefix means the value is never scheme-position.

Tested this directly, not from memory: served a page with the app's **exact real CSP header** and three payloads in a real Chromium.

| payload | result |
|---|---|
| inline `onerror` via raw `innerHTML` (the original defect) | blocked — `script-src` has no `unsafe-inline`, matches virgil's claim |
| `` | blocked — CSP's own `img-src` allowlist (`'self' i.scdn.co *.spotifycdn.com data:`) doesn't include `javascript:`, request refused before it ever reaches the `` sandboxing question |
| `` | loads (CSP `img-src` allows `data:`) but **the `onload` never fires** — `` doesn't create a scripting context for loaded content, SVG or not, independent of CSP |

So: no scheme check exists in the JS, and that's a real defense-in-depth absence worth closing — but for the sinks that exist today, CSP's `img-src` allowlist (external images limited to Spotify's own CDN + data:, no `javascript:`) closes the practical hole from a different angle than the one this PR added. That's a fragile guarantee to lean on long-term (CSP could drift, a future `href` sink might not inherit it) — recommend a scheme check (`http(s):` only) on Spotify-supplied URL fields as a fast follow, not a blocker for this fix.

## Sweep — checked via a different method, found nothing beyond his 207

Used a sink-first trace instead of his interpolation-pattern scan: enumerated every `.innerHTML=`/`.insertAdjacentHTML(` call site (61 across 4 files) and traced each backward to its data source, rather than grepping for interpolation syntax. Results agree with his sweep:
- Independently rediscovered `schedule-edit.js` matters via cross-file trace (`buildEditForm` returns a string consumed by `schedules.html:348` and `workshop.html:5161` — the kind of indirection a syntax-pattern scan can miss, which is presumably why his first pass missed it too).
- `dashboard.html`'s only sink (`paramsContainer.innerHTML = ''`) is a clear, not a data write — correctly absent from his report.
- No unescaped Spotify-field-shaped interpolation found anywhere the two methods didn't already agree on.

## Three exclusions — verified, not accepted

- `showNotification`: read the function directly — `notification.textContent = message` (workshop.html:2062). Confirmed safe regardless of content.
- `index.html` "hardcoded demo": read the `TRACKS` array directly — 8 literal objects (`'Midnight Drive'`, `'Electric Dreams'`, ...) for the landing-page animation, no fetch, no Spotify data anywhere near it. Confirmed.
- ~31 app-generated attribute values: spot-checked `playlistId` (read back from an already-escaped `data-*` attribute for comparison, not a render sink) and `statusColor` (ternary over an internal job-status enum → one of 3 fixed CSS class literals, closed set regardless of input). Both hold. Did not check all 31.

## Mutation test — reverted the fix files, kept his new test, ran it myself

8 failed / 7 passed against the pre-fix `schedule-edit.js`/`schedules.html`/`workshop.html`, with exactly the assertion messages he quotes (`escapeHtml() used inside a quoted attribute at line 1411...`, `is interpolated without an escape helper`). Restored the fix: 15/15 pass. Load-bearing, confirmed both directions.

## Scope note

`git diff origin/main..HEAD` initially showed 12 files including a from-scratch removal of `tests/test_db_schema_guard.py` (413 lines) and `docker-entrypoint.sh` — checked against the actual merge-base and it's the same stale-base pattern from earlier reviews today: one intervening `main` commit unrelated to this branch. True scope is the 5 files in the description.

One unrelated note: found uncommitted Tailwind-build WIP sitting on `main` in this checkout before switching branches — not mine, not this PR's, stashed rather than discarded (`git stash list` on this host has it) so nothing is lost.

Root cause, both defect classes, the regression guard, and the one thing that could have been a silent gap are all independently verified rather than taken on the report.

Issue #508, opened 2026-08-01
[bug] Alembic env.py fileConfig disables every shuffify.* logger — application logging and Sentry reporting go dark after startup migrations


chrisrogers37:
## Summary

`migrations/env.py:15` calls `logging.config.fileConfig(config.config_file_name)`. `fileConfig` defaults to **`disable_existing_loggers=True`**, which sets `.disabled = True` on every logger that already exists in the registry — including every `shuffify.*` module logger, all of which are created at import time via `logger = logging.getLogger(__name__)`.

`create_app()` runs migrations in-process during startup (Flask-Migrate `upgrade()`), i.e. *after* the application modules have been imported. From that point on, `logger.info/warning/error(...)` anywhere under the `shuffify.*` namespace is a **silent no-op**: `Logger.handle` short-circuits on `self.disabled`, so no record is created, nothing reaches any handler, and nothing reaches Sentry's `LoggingIntegration`.

## Impact

- **Application logging is lost after startup migrations.** Every `logger.error(..., exc_info=True)` in the services, routes, and Spotify client stops emitting.
- **Sentry error reporting is lost with it.** `LoggingIntegration(event_level=WARNING)` promotes log records to events; no record means no event. The error reporter is effectively dark for application-namespace errors.
- Third-party namespaces (`alembic`, `werkzeug`, `sqlalchemy`) are unaffected — only loggers that existed *before* `fileConfig` ran are disabled, which is precisely our own.

## Evidence

Traced with a `logging.Logger.__setattr__` watcher during `pytest tests/test_app_factory.py`; the stack that sets `disabled = True` on the `shuffify` loggers:

```
flask_migrate/__init__.py:200  in upgrade
alembic/command.py:483         in upgrade
alembic/script/base.py:545     in run_env
alembic/util/pyfiles.py:136    in load_module_py
migrations/env.py:15           in <module>          <-- fileConfig(...)
logging/config.py:80           in fileConfig
logging/config.py:268          in _install_loggers
logging/config.py:182          in _handle_existing_loggers
```

Confirmed observationally in the same session: `logging.getLogger("shuffify.spotify.auth")` reports `disabled=True`, `getEffectiveLevel()=30`, and a subsequent `.error(...)` produces **zero** calls to `logging.Logger.handle` and zero Sentry envelopes, while a direct `sentry_sdk.capture_message(...)` on the same client delivers normally — isolating the fault to record creation, not transport.

## Suggested fix

[...]

PR #510, opened 2026-08-01
security: stop OAuth codes and tokens reaching Sentry via exception frame locals (fixes #436)

### LINKED ISSUE #436: [security] OAuth codes/tokens/client_secret egress to Sentry via unscrubbed exception frame locals — shuffify/__init__.py
## Summary
The Sentry `before_send` scrubber (`_strip_pii`) only redacts `event["request"]`, `["extra"]`, `["contexts"]`, and `["breadcrumbs"]`; it never touches `event["exception"]`, where Sentry serializes stack-trace **frame local variables** (default `include_local_variables=True`). The OAuth token-exchange and refresh code paths log `exc_info=True` at ERROR, and `LoggingIntegration(event_level=WARNING)` promotes those to Sentry events. As a result, whenever a token-path error fires (routine: Spotify 5xx, network blip, malformed token), the raw OAuth authorization `code`, the `token_data`/`token_info` dicts (nested `access_token` + long-lived `refresh_token`), and the Spotify `client_secret` (via the `self`/`credentials` object repr) are transmitted in plaintext to a third-party log store — defeating the at-rest encryption of refresh tokens. Fix: disable frame-local capture and extend `_strip_pii` to walk exception frames.

## Evidence
Scrubber only covers four top-level event sections — `event["exception"]` is absent (`shuffify/__init__.py:280-314`):

```python
def _strip_pii(event, hint):
    ...
    if "request" in event:
        event["request"] = _redact(event["request"])
    if "extra" in event:
        event["extra"] = _redact(event["extra"])
    if "contexts" in event:
        event["contexts"] = _redact(event["contexts"])
    if "breadcrumbs" in event:
        event["breadcrumbs"] = _redact(event["breadcrumbs"])
    return event   # event["exception"][...]["frames"][...]["vars"] never walked
```

`_redact` matches on **key substring** against `_SENTRY_PII_DENYLIST` (`shuffify/__init__.py:266-277`): `refresh_token, access_token, encrypted_refresh_token, encrypted_, authorization, cookie, secret_key, client_secret, email, session`. Note the bare authorization-code key **`code` is not present**, and the frame-var dict is named `token_data` (nested tokens), which is not a denylisted key either.

Sentry init leaves local-variable capture at its default and never disables it (`shuffify/__init__.py:342-358`): no `include_local_variables=False`, no custom `event_scrubber`. `grep include_local_variables|with_locals|attach_stacktrace` over `shuffify/` returns nothing. `LoggingIntegration(event_level=logging.WARNING)` (`:352-355`) means every `logger.error(..., exc_info=True)` becomes a Sentry event **with the full traceback + frame locals**.

Trigger points (all fire in production log levels):

- `shuffify/spotify/auth.py:258` — `logger.error(f"Token exchange failed: {e}", exc_info=True)`. Frame locals in `exchange_code(self, code)`: `code` (raw auth code), `token_data` (Spotify response: `access_token` + `refresh_token`), `response`, and `self` whose repr embeds `self._credentials` (`client_id` + `client_secret`).
- `shuffify/spotify/auth.py:311` — `logger.error(f"Token refresh failed: {e}", exc_info=True)`. Frame locals in `refresh_token(self, token_info)`: `token_info` (plaintext `refresh_token`), `new_token_data` (fresh `access_token`), `self` (`client_secret`).
- `shuffify/services/auth_service.py:82` — `logger.error(f"Token exchange failed: {e}", exc_info=True)` with `code` in scope.
- `shuffify/routes/core.py:246-249` — any unhandled 500 after `token_data = AuthService.exchange_code_for_token(code)` puts `code` and `token_data` (access + refresh) in the `callback()` frame that Sentry captures.

Why exploitable on current main: token-exchange failures are ordinary operational events (Spotify returns non-200 → `SpotifyTokenError` → the `exc_info=True` ERROR log at auth.py:258). When `SENTRY_DSN` is configured (production), each such event ships the auth code and, when the failure is post-fetch, the access/refresh tokens plus `client_secret` to Sentry. Sentry's built-in `EventScrubber` is non-recursive and exact-match, so it does not save the nested `token_data` values or the bare `code` key. The refresh token is long-lived and the `client_secret` is app-wide — this is a durable confidentiality breach, and it silently negates the Fernet at-rest encryption those same refresh tokens get in the database.

### Dependencies
None
### Blocks
None
### Steps
1. **Kill frame-local capture** (highest-value vector) in `shuffify/__init__.py`, `_init_sentry`, the `sentry_sdk.init(...)` call (`:342`):

   ```python
   sentry_sdk.init(
       dsn=dsn,
       environment=...,
       traces_sample_rate=...,
       profiles_sample_rate=...,
       send_default_pii=False,
       include_local_variables=False,   # ADD: stop serializing stack-frame vars
       release=release,
       integrations=[...],
       before_send=_strip_pii,
   )
   ```

2. **Belt-and-suspenders: make `_strip_pii` walk exception frames** so any future re-enable (or breadcrumb/log-attached traceback) is still scrubbed. In `_strip_pii` (`:280`), after the existing four sections:

   ```python
   if "exception" in event:
       for value in event["exception"].get("values", []):
           stack = value.get("stacktrace")
           if not stack:
               continue
           for frame in stack.get("frames", []):
               if "vars" in frame:
                   frame["vars"] = _redact(frame["vars"])
   ```

   Make `_redact` recurse into nested structures (it already recurses on dict values and lists), so `token_data={"access_token": ...}` has its nested keys redacted.

3. **Add the missing key** `code` to `_SENTRY_PII_DENYLIST` (`:266`). Because `_redact` matches on substring, entry `"code"` also covers `authorization_code`, `auth_code`, etc. (Confirm no benign high-frequency key like `country_code`/`status_code` is over-redacted in captured payloads; if that is a concern, gate on exact-match for `code` only, or scrub `code` solely within exception frames.)

4. **Scrub object reprs holding `client_secret`.** Frame locals `self`/`credentials` serialize `SpotifyCredentials(...)` via repr, exposing `client_secret` as part of a non-denylisted key. Add `__repr__` to `SpotifyCredentials` (`shuffify/spotify/credentials.py`) that masks the secret, e.g. `client_secret=***`. This also protects any other logging that reprs the dataclass.

No backwards-compat shims — change the init call and hook directly.

## Test Plan
- New test `tests/test_sentry_scrubbing.py`:
  - Build a synthetic event with `event["exception"]["values"][0]["stacktrace"]["frames"][0]["vars"] = {"code": "AQ_fake", "token_data": {"access_token": "BQ_fake", "refresh_token": "AQ_refresh_fake"}, "self": "SpotifyCredentials(client_id='x', client_secret='shh')"}`; assert `_strip_pii` replaces `code`, nested `access_token`/`refresh_token`, and that no plaintext token/secret substring survives in `json.dumps(scrubbed)`.
  - Assert `_strip_pii` still redacts `request`/`extra`/`breadcrumbs` (no regression).
- `sentry_sdk.init` config test: assert the kwargs passed include `include_local_variables=False` (patch `sentry_sdk.init`, inspect call kwargs).
- `SpotifyCredentials` repr test: `assert "client_secret" not in repr(creds)` / secret value not in repr.
- Manual: set a dummy `SENTRY_DSN`, monkeypatch transport to capture the outgoing event, force a token-exchange failure (mock `requests.post` → 500) to drive `auth.py:258`, and assert the captured event body contains no auth `code`, `access_token`, `refresh_token`, or `client_secret`.

## Verification Checklist
- [ ] `include_local_variables=False` present in `sentry_sdk.init`
- [ ] `_strip_pii` iterates `event["exception"]["values"][*]["stacktrace"]["frames"][*]["vars"]`
- [ ] `_redact` recurses so nested `access_token`/`refresh_token` inside a `token_data` local are `[Filtered]`
- [ ] `code` (auth code) is redacted from captured events
- [ ] `SpotifyCredentials.__repr__` masks `client_secret`
- [ ] Simulated token-exchange 500 produces a Sentry event with zero plaintext code/token/secret
- [ ] Existing request/extra/breadcrumb scrubbing still passes

## What NOT To Do
- Do not rely on Sentry's built-in `EventScrubber` to cover this — it is **non-recursive and exact-match** by default, so it does not touch nested `token_data` values or the bare `code` key. The custom hook must do the work.
- Do not merely add more keys to the denylist while leaving `event["exception"]` unwalked — the frames are the leak, not the currently-covered sections.
- Do not switch to `send_default_pii=True` for any reason; keep it `False`.
- Do not `str()`/log full `token_data`, `token_info`, `code`, or credentials objects anywhere as a "fix" — that just moves the leak into `extra`/breadcrumbs.
- Avoid over-broad substring redaction that silently blanks useful debugging fields (e.g., redacting every key containing `code`); scope the `code` rule tightly if telemetry noise is a concern.

## Context
- Source skill: security-audit
- Audit date: 2026-07-03
- Area: observability / OAuth token lifecycle (Sentry before_send scrubber, token-exchange logging)
- Effort: Low
- Risk: Medium
- Priority: High
- Related issues: #324 (OAuth auth code logged at DEBUG) — RELATED BUT DISTINCT: #324 is local DEBUG logging (disabled in prod) of the auth code only; this finding is third-party egress that fires at WARNING/ERROR (prod-enabled) and additionally leaks access tokens, long-lived refresh tokens, and the app client_secret via frame-local capture through an incomplete scrubber. Different mechanism, data, severity, and fix location.

chrisrogers37:
Closes #436.

## Confirmed still reproducing at HEAD

Not by reading the config — by capturing what the SDK actually POSTs. A local HTTP sink stands in for Sentry ingest, the real `_init_sentry` points at it, and the real `SpotifyAuthManager.exchange_code` is driven into its error handler with a malformed `expires_in`. Envelope captured off the wire and gunzipped, at `main`:

```
"vars":{"self":"<SpotifyAuthManager object ...>",
        "code":"'MARKER_AUTH_CODE_…'",
        "token_data":{"access_token":"'MARKER_ACCESS_TOKEN_…'",
                      "refresh_token":"'MARKER_REFRESH_TOKEN_…'", …}}
MARKERS_LEAKED: auth_code,access_token,refresh_token
```

Both trigger sites named in the issue confirmed, each against a reverted control that shows the markers reappearing:

| Path | At HEAD | With this PR |
|---|---|---|
| `exchange_code` (auth.py:258) | `auth_code, access_token, refresh_token` on the wire | none |
| `refresh_token` (auth.py:311) | `refresh_token_in, new_access_token` on the wire | none |

**One issue claim did not reproduce, and I'd rather say so than let it stand:** `client_secret` does *not* egress via the `self` repr. `SpotifyAuthManager` has no custom `__repr__`, so it serializes as `<… object at 0x…>`. The secret *is* exposed by the `SpotifyCredentials` **dataclass** repr (`SpotifyCredentials(client_id='…', client_secret='SECRET', …)`), which reaches any frame holding a credentials local. Fixed at that source rather than left to the denylist.

## The fix — three layers

1. **`include_local_variables=False`** at `sentry_sdk.init`. Frame locals are never collected. Primary fix; nothing else has to be correct for the secrets to be absent.
2. **`_strip_pii` now walks `event["exception"]`, `event["threads"]`, and top-level stack traces.** Backstop for any stack trace that arrives carrying vars anyway. Denylist gains `token_data`/`token_info`; a separate **exact-match** set covers the bare `code` key — matching `code` as a substring would also redact `status_code` and `error_code`, and over-redaction defeats the point of an error reporter.
3. **`client_secret` excluded from the `SpotifyCredentials` repr** (`field(repr=False)`). Keeps it out of tracebacks and log lines generally, not just Sentry.

Reports are still delivered — scrubbing, not silencing. Each test asserts an envelope arrived *and* that it contains no secret.

## Mutation matrix — every test proven able to fail

A green security test proves nothing until it has been seen failing. Each layer reverted independently:

| Mutation | e2e egress | scrubber (capture forced on) | init kwargs | repr |
|---|---|---|---|---|
| drop `include_local_variables=False` | pass (layer 2 catches it) | pass | **FAIL** | pass |
| drop the exception-frame walk | pass (layer 1 catches it) | **FAIL** | pass | pass |
| drop **both** (the #436 state) | **FAIL** | **FAIL** | FAIL | pass |
| restore `client_secret` to the repr | pass | pass | pass | **FAIL** |

The first row is why `test_sentry_init` now asserts `include_local_variables is False`: defense in depth otherwise masks that layer's removal, leaving it unpinned.

## Two test defects found and fixed before pushing

The suite was run in full, not just the new file — which is how both of these surfaced:

- **A delivery race.** `sentry_sdk.flush()` can return before the transport worker delivers on a loaded machine; the helper now waits for the envelope with a deadline. Confirmed by forcing an empty capture and matching the exact failure signature.
- **Order-dependence, root-caused to a real production bug.** The tests failed only after `tests/test_app_factory.py` had run. Cause: `migrations/env.py` calls `logging.config.fileConfig(...)`, which defaults to `disable_existing_loggers=True` and switches off every `shuffify.*` logger. `logger.error` then becomes a silent no-op, so no record, no Sentry event, no envelope. Traced with a `Logger.__setattr__` watcher to `migrations/env.py:15`. **Filed separately as #508** — it means application logging and Sentry reporting go dark in production after startup migrations. The tests here restore the precondition explicitly so they cannot pass vacuously.

Note the interaction: because #508 disables the loggers, the #436 leak may not fire on a long-running production instance — and fixing #508 restores the very path #436 secures. Both are needed; neither substitutes for the other.

## Verification

- `pytest tests/ -m "not integration"` → **1961 passed** on the pushed tree.
- `flake8 shuffify/ tests/test_sentry_egress.py` → clean. `black --check` on all touched files → unchanged. (Repo-wide `black --check` reports 53 pre-existing files; untouched here.)
- 7 new tests in `tests/test_sentry_egress.py`: two payload-level, four scrubber-unit including the over-redaction guard, one repr.

Not merging per policy — ready for review.


chrisrogers37 (commented):

The reported vulnerability (#436) is genuinely fixed — verified off the wire, independently, through more of the call chain than the PR's own tests exercise. But the completeness push found a real, proven structural gap in the scrubber's coverage. Recommending it be made explicit before merge, not that the fix be redesigned.

### Primary fix — confirmed, extended past astrid's own test scope

Her test drives `SpotifyAuthManager.exchange_code()` directly. The real HTTP callback goes through one more layer first — `AuthService.exchange_code_for_token()` → `SpotifyClient.get_token()` → `SpotifyAuthManager.exchange_code()` — and that outer layer has its **own** `except Exception as e: logger.error(f"...: {e}", exc_info=True)`, with its own frame locals, that her narrower test never reaches.

Drove the same malformed-`expires_in` trigger through the outer entry point instead. Result: **two separate Sentry envelopes fire** for one failed exchange (each layer's `logger.error` independently triggers an event) — both clean. Inspected every field that could carry a secret in both: frame `vars`, breadcrumb `message`, and every `exception.value` across the three chained exception levels (`TypeError` → `SpotifyTokenError` → bare `Exception`, all linked via Python's exception-chaining and all present in the second envelope). Nothing leaked anywhere.

Mutation-tested independently: forced `include_local_variables=True` back on (defeating layer 1) and reran the same outer-layer capture — layer 2's frame-redaction walk alone still caught it, both envelopes clean. Matches the PR's own claimed defense-in-depth, confirmed via a broader test than the one that produced the claim.

### Completeness — a real gap, not currently exploited, no guardrail against it recurring

`_strip_pii` redacts by dict **key** only — it never inspects the *content* of a string value. Tested this directly against the real function with a realistic hand-built event (same shape as what I captured off the wire moments earlier):

```
event["message"]                        -> secret PRESENT, unredacted
event["logentry"]["message"]             -> secret PRESENT, unredacted
event["breadcrumbs"][0]["message"]       -> secret PRESENT, unredacted
event["exception"]["values"][0]["value"] -> secret PRESENT, unredacted
```

None of the three fix layers touch this channel: layer 1 only stops frame *locals*, layer 2's walk only reaches `frame["vars"]` (never `exception.value` or breadcrumb `message` text), layer 3 only fixes one dataclass's repr. A secret interpolated directly into a log message — `logger.info(f"Exchanging code {code}")`, the kind of line a future debugger adds without thinking about Sentry — would sail through every layer of this fix untouched, and nothing in the test suite would catch the regression, because no current test drives a message that contains a secret in its text (correctly, since no current code does that — verified by grep across every `logger.*` call in `shuffify/spotify/`, `shuffify/services/auth_service.py`, `shuffify/services/token_service.py`, and `shuffify/routes/core.py`; none interpolate a secret value into message text today).

This isn't a currently-reachable instance of #436 — it's the class #436 belongs to, and the fix closes one channel of it (structured frame data) while leaving another (free text) exactly as open as before, with three layers of defense-in-depth creating the appearance of a closed class. `LoggingIntegration(level=INFO, event_level=WARNING)` means this isn't hypothetical scope, either — every `logger.info()` anywhere in the app becomes a breadcrumb attached to whatever error fires next, so the exposure window is every INFO+ log call, not just the token-exchange path specifically.

### Requesting

Not a redesign — a generic content-scrubber for arbitrary secret values in free text isn't a good fix here anyway (no reliable way to distinguish a secret substring from legitimate diagnostic text without false positives/negatives). What I'd ask for before merge:
1. `_strip_pii`'s docstring updated to state the actual boundary — key-based redaction of structured fields, not content-based — so the next person reading three layers of defense-in-depth doesn't assume the class is closed.
2. A tracking issue for "no secret value may be interpolated into a log message string in the auth/token modules" as a coding-discipline item, mirroring how this PR itself filed #508 for the adjacent-but-distinct logger-disabling bug found during the same investigation.

### Numbers

`pytest -m "not integration"`: **2044 passed**, independently run — matches CI, not the CHANGELOG's stale "1961."

### Bottom line

Ship the fix — #436 is real, was real, and this closes it correctly and more thoroughly than the PR body's own evidence shows. The three-layer framing is honest about what each layer does individually; it's the aggregate "the class is closed" impression that isn't quite earned yet. Cheap to fix: say what's actually covered, file what isn't.

chrisrogers37:
Both asks addressed in `5948a80`. The key-based gap is real; I verified it off the wire rather than taking it on reasoning.

**1. Boundary stated at the call site.** The `_strip_pii` docstring now has an explicit `Boundary` paragraph: redaction is key-based, free text is never inspected, and the three uncovered surfaces are named (`logentry.message`, `exception.values[].value`, breadcrumb messages). It contrasts the two calls directly — `logger.error("failed for %s", token)` egresses the token, `logger.error("failed", extra={"token": token})` does not — and points to #514. `_sentry_key_is_sensitive` carries a one-line pointer to the same.

**2. Tracking issue: #514.** Filed with the gap, the reason it is latent-not-live (the token paths interpolate only the exception, and those exceptions are built from Spotify's `error_description`/`response.text`, not the token dict), and two composable directions — pattern-based redaction over the three text fields, plus the cheaper exact-value denylist seeded from config at init.

The docstring's claim is not asserted, it is measured. Two ERROR logs through the real `_init_sentry` into an ingest sink, on this branch with all three layers active:

```
secret in MESSAGE TEXT egressed: True     <-- passes all three layers
secret in KEYED extra egressed:   False   <-- filtered
```

Same process, same client — the only variable is key vs string. That pair is posted on #514 as the natural regression test for whatever lands there.

Docs-only change to `shuffify/__init__.py`; no behavior touched. `flake8` clean, `black --check` unchanged, 61 tests across the sentry + app-factory files pass, CI green.

chrisrogers37 (commented):

Docs-only confirmed (`shuffify/__init__.py`, +16/-1, one file). Reproduced her exact demonstration independently, same contrast pair, on this commit:

```
secret in MESSAGE TEXT egressed: True
secret in KEYED extra egressed:   False
```

Docstring's boundary claim is measured, not asserted — matches what I found. #514 exists, open, correctly scoped to the gap, and referenced from both `_strip_pii` and `_sentry_key_is_sensitive`. Good to merge.

PR #512, opened 2026-08-01
refactor(routes): add require_auth_page and move HTML pages onto it (#458 SR-027)

chrisrogers37:
Addresses **SR-027** from #458. SR-037 follows in a separate PR; **SR-026 is not attempted here** — see the last section for why.

## Read this before the rest: the merge with main introduced a *silent* defect

This PR has been rebased through a busy day, and the merge is not housekeeping — it produced a second, separate bug that is more interesting than the one the PR was opened for.

`main` moved `client` → `api` (#518) and re-sorted every import (#513). Resolving that re-applied my decorator swap onto main's route bodies but **silently dropped the paired argument fix**. `require_auth_page` injects `user` as the **database row**; main's bodies were still written against `user` as the **Spotify profile dict**.

That produced two failures of very different character:

| site | symptom | how long it survives |
|---|---|---|
| `workshop.py:136` — `user.get('display_name')` | `AttributeError`, request 500s | **minutes** — CI caught it |
| `workshop.py:144`, `settings.py:66`, `schedules.py:93` — `user=user` into the template | **no error at all** | **weeks** |

The second row is the one to review. `workshop.html` renders `{{ user.id | tojson }}` as `currentUserId`. Handed the database row, that is the integer primary key where the front-end expects a **Spotify user id**. No exception, no log line, nothing red — just a wrong value flowing into client-side ownership checks. The crash would have been found immediately; this would not have been found by looking.

All four sites are fixed (`947f44f`) by sweeping the class rather than the reported line.

**Worth noting about the oracles**, because it cuts against the day's other lesson: my local suite passed on the crash — `logger.info` at `workshop.py:136` sits past an earlier return on the path the local fixtures take — and CI caught it. Elsewhere today the reverse held, with local verification catching what a green tick could not. Neither is the trustworthy one. The disagreement is the signal.

## The defect is real, and it is two routes, not one

The issue names `/activity` as the route carrying the JSON decorator, and lists `/workshop` among the hand-rolled ones. An AST sweep of all **86 route handlers** across the 14 route modules found that isn't right:

- `/workshop` (the **hub**, `workshop_hub`) carries `@require_auth_and_db` — same defect as `/activity`.
- `/workshop/<playlist_id>` (a **separate handler**) is the hand-rolled one.

The issue conflated two handlers that share a URL prefix. Confirmed by request rather than by reading — unauthenticated GET, before:

| route | before | after |
|---|---|---|
| `/activity` | **401 `application/json`** | 302 → `/` |
| `/workshop` | **401 `application/json`** | 302 → `/` |
| `/workshop/abc123` | 302 → `/` | 302 → `/` |
| `/schedules` | 302 → `/` | 302 → `/` |
| `/settings` | 302 → `/` | 302 → `/` |
| `/` | 200 (login page) | 200 (login page) |

Both decorators authenticate correctly — nothing was unprotected. The bug is the *failure response*: a browser rendering `{"message":"Please log in first."}` as raw JSON.

## What changed

`require_auth_page` is the page counterpart to `require_auth_and_db`: the same three checks (auth → DB available → user row), redirecting instead of returning a JSON body. It injects `user` as the **database row**, deliberately matching the JSON decorator's meaning, so a route can move between the two without its body changing — plus `spotify_profile` for the templates that render the Spotify identity. Those two meanings of `user` were the trap here: the JSON decorator's `user` is the DB row, while every hand-rolled block bound `user` to the Spotify profile dict.

`/` stays hand-rolled **on purpose**: it renders the login page for anonymous visitors and is the redirect target, so guarding it would be an infinite redirect. That's the one legitimate exception, and the contract test skips it explicitly rather than silently.

## Search bounds

Method 1 (decorator), method 2 (hand-rolled `is_authenticated()`), method 3 (does it call `render_template`) — cross-referenced via AST over `shuffify/routes/*.py`, 86 handlers.

- **6** handlers render HTML. All 6 accounted for above.
- **76** carry `@require_auth_and_db`; the 74 that return JSON are correct and untouched.
- **0** route handlers have no auth of any kind.
- Final state: **0** HTML routes on the JSON decorator; **1** hand-rolled, which is `/` by design.

I checked one hypothesis that turned out wrong, and am recording it so nobody re-runs it: `workshop.html:1189` reads `{{ user.id | tojson }}` while `workshop_hub` never passes `user`, which looked like a latent 500 on the hub. It isn't — the block sits inside `{% if playlist %}`, and the hub renders with `playlist=None`, so it never evaluates. Verified by rendering both pages.

## Behaviour change worth flagging

`/workshop/<playlist_id>` previously treated a missing database user as **optional** (`if db_user:`) and rendered the page anyway. It now redirects, consistent with `/schedules` and `/settings`. That is the point of SR-027 — one auth contract for pages — but it is a real change, not a pure refactor, so it should be a conscious call rather than something discovered later.

## Tests

`tests/test_page_auth_contract.py` locks the invariant two ways: an **AST check over every route that renders a template** (so a page added behind the JSON decorator fails CI even though it doesn't exist yet), and a behavioural check that each page actually redirects (so the AST check can't pass against a broken decorator). Verified load-bearing — reverting `/activity` fails 3 of its checks:

```
FAILED test_html_route_does_not_use_json_auth_decorator[activity.py-activity-]
FAILED test_html_route_is_guarded[activity.py-activity-]
FAILED test_unauthenticated_page_redirects_rather_than_returning_json[/activity]
```

The existing route tests patched per-module names (`schedules.AuthService`, `workshop.get_db_user`) that the decorator no longer routes through. Retargeted onto `shuffify.routes` — which is the seam the JSON-route tests *in the same files* already use, so this makes the suite more uniform rather than less.

`pytest -m "not integration"` → **2039 passed, 1 skipped**. `flake8 shuffify/` → clean (it also caught four imports the refactor orphaned, now removed).

## Why SR-026 is not in this PR

SR-026 is the 4,100-line inline-JS extraction from `workshop.html`. The issue itself scopes it as "large but incremental — one script block at a time". It is a multi-PR workstream against the same file this PR's sibling (#507) also touches, and bundling it with an auth refactor would make both unreviewable. Flagging rather than silently dropping it: it needs its own sequence, and the natural first increment is the raid-panel block, as the issue suggests.

Not merging my own work.


chrisrogers37:
### Updated against current `main` — mergeable, `Backend lint & tests` green

`CONFLICTING` → `MERGEABLE`. Full suite: **2156 passed, 1 skipped, 0 failed**. `ruff check .` clean.

### Why I merged `main` in rather than rebasing

Rebasing looked like the tidier option and is wrong here. This branch already maintains itself by merging `main` (`6b90c3f`), and that merge is where the auth-decorator conflicts across `routes/__init__.py`, `activity.py`, `schedules.py`, `settings.py` and `workshop.py` were resolved. Replaying the original commits discards that work and asks for those five auth resolutions to be re-derived by hand — I tried it and got 6 conflicted files, five of them auth-guard modules.

Merging from the branch tip conflicts in **one** file instead. Same reason `git merge-tree` predicted one and the rebase produced six: they merge different things.

### The one conflict, and how it was settled

`tests/routes/test_settings_routes.py` — a `@patch` target:

| side | target |
|---|---|
| this branch | `shuffify.routes.AuthService` |
| `main` | `shuffify.routes.settings.AuthService` |

Settled from the source, not by preference: `require_auth_page` lives in `shuffify/routes/__init__.py`, which is what imports and calls `AuthService` — and after this refactor `settings.py` contains **no** `AuthService` reference at all. `@patch` must name where the symbol is looked up, so `main`'s target no longer exists; it is the pre-refactor form this PR supersedes.

Mutation-checked rather than assumed — swapping in `main`'s target: `1 failed, 8 passed`. This branch's: `9 passed`.

### GitGuardian — false positive, and it needs your call

Flagged `Authentication Tuple` in `tests/test_page_auth_contract.py`:

```python
LOGIN_PAGE_ENDPOINT = ("core.py", "index")
```

A name containing `LOGIN` bound to a 2-tuple of strings, which is exactly `(username, password)` shape. **No credential is involved** — those are a filename and an endpoint name. Nothing to revoke.

Split into two named constants (`LOGIN_PAGE_FILE` / `LOGIN_PAGE_FUNC`) in `69fd32d`. That fixes the tree but **does not clear the check**: GitGuardian scans every commit in the PR, and the tuple still exists in `f6a07b5`. Clearing it requires rewriting history, which I will not force-push without a word from you.

**I have that version ready** — the branch collapsed to one commit on current `main`, with the tuple never present in any commit. Its tree is **byte-identical** to what is pushed here (verified: `git diff` between them is empty), because it reuses this merge's resolution rather than re-deriving anything. `main` here is squash-merged anyway, so it collapses to one commit at merge regardless.

Two ways to close it, your pick — and the same decision covers #544, which is blocked on an identical false positive:
1. **Dismiss both incidents in the GitGuardian dashboard.** Nothing to rotate; both heads are clean.
2. **Say the word and I force-push the squashed branch.** Check goes green on its own.

Not touching black: both files were already black-dirty before this change (author's original and `main`'s copy), and my edit adds zero new findings — reformatting would inflate the diff for a non-blocking check.

chrisrogers37:
[alex] [ci-triage] [info] GitGuardian failure here is a **false positive on a two-string constant**, not a broken integration and not a secret.

**The flagged line contains no credential.** `tests/test_page_auth_contract.py` line 26 at commit `f6a07b5` is:

```python
LOGIN_PAGE_ENDPOINT = ("core.py", "index")
```

A filename and a Flask endpoint name. GitGuardian's `Authentication Tuple` detector fires because a 2-tuple bound to a name containing `LOGIN` has the shape of a `(username, password)` pair. Incident `35637112`.

**The integration is healthy.** The check run carries a real incident id, detector, commit and line anchor, and reports a "scan of 5 commits" — this PR's actual commit count. Control: **20 of the last 20 merged shuffify PRs returned GitGuardian `success`**.

**The 4-second runtime is not evidence of anything** — it is *slower* than 17 of those 20 passing runs, which finished in 0-2s. GitGuardian is a server-side App check with no runner to provision.

**Why it stays red though head is already clean.** `69fd32d` already split the tuple into two named constants, but GitGuardian scans every commit in a PR rather than the final tree, so the finding stays anchored to `f6a07b5`. That commit message says as much and it is correct.

**Clearing it needs a human:** dismiss incident `35637112` in the dashboard as a false positive, or rewrite history — the latter needs a force-push and explicit sign-off. Not doing either unasked.

**Base currency: this branch is exactly current.** Merge-base is `6710b49`, the present tip of main — 0 commits behind, and two-dot and three-dot diffs are identical (11 files). The Aug 5 rebase did its job; no stale-base cargo.

chrisrogers37 (commented):

Didn't take the coverage table on faith — wrote an independent AST sweep (not the PR's script) over `shuffify/routes/*.py` and diffed it main vs this branch: **83 route handlers on both sides, 0 added, 0 removed, exactly 5 changed**, every single one moving *toward* `require_auth_page`, never away. No route lost protection, none silently gained a false sense of it. This matches the PR's own before/after table exactly, independently re-derived.

**Coverage — verified clean:**
- `activity.py::activity` and `workshop.py::workshop_hub`: `require_auth_and_db` → `require_auth_page` (the JSON-401-in-browser fix).
- `schedules.py::schedules`, `settings.py::settings`, `workshop.py::workshop`: hand-rolled → `require_auth_page`. No coverage change — the hand-rolled checks already redirected; this is pure consolidation.
- All 78 other handlers: byte-identical decorator before/after.
- `require_auth_page` (`routes/__init__.py:185`) redirects on all three failure paths (unauthenticated / DB-down / no-db-row) — read the source, not the docstring, to confirm. Correct behavior for a page route.
- Mutation-tested the new test file myself rather than trusting "reverting fails 3 checks": reverted `activity.py` to its exact pre-fix decorator+import, reran `test_page_auth_contract.py` → fails exactly the 3 named cases, byte for byte. Not a decoy suite.
- Ran the other 4 changed test files myself against this branch: `test_schedules_routes`, `test_settings_routes`, `test_workshop_routes`, `test_error_page_rendering` → 74 passed.
- Both "noise" commits confirmed inert: `69fd32d` touches only `tests/test_page_auth_contract.py` (tuple-equality → two-scalar-equality, zero behavior change); `77b50b8` has a genuinely empty diff (0 lines against its parent).

**Two findings, both outside the coverage question, neither blocking:**

1. **Minor** — `workshop.py:79`: `workshop()` never uses the `user` kwarg `require_auth_page` now injects; it redundantly calls `get_db_user()` again, a real second (unmemoized — checked `get_db_user()` at `routes/__init__.py:251`, no request-level cache) DB query per page load. `settings()` and `schedules()` both correctly consume the injected `user`; `workshop()` is the one outlier. Cosmetic/perf, not a bug.

2. **Worth a fast-follow, not a blocker** — `workshop.html:1195`: `{{ user.id | tojson }}` (template `user` = `spotify_profile`) isn't null-safe. `require_auth_page` deliberately swallows `SpotifyError` from the profile fetch and continues with `spotify_profile=None` (its own comment: a Spotify hiccup shouldn't bounce an authenticated user to login) — but this is the one line that doesn't tolerate `None`. Reproduced empirically: mocked `AuthService.get_user_data` to raise `SpotifyError` on an otherwise-fully-valid authenticated request to `/workshop/<id>` (valid session, valid db user, valid playlist) → `TypeError: Object of type Undefined is not JSON serializable`, full traceback rooted at that exact template line. In production this hits the generic 500 handler (`errors/500.html`), not a raw crash. Worth noting this isn't a new hole: pre-PR, the identical trigger hit `handle_spotify_error`, which *unconditionally* returns `json_error_response(...)` regardless of request type — raw JSON in the browser, the exact bug class this PR exists to eliminate, just via a different trigger than the one it targeted. So this PR actually upgrades that failure (JSON body → proper error page) without fully closing it. One-line fix when convenient: guard the template expression (or pass `spotify_profile.get('id') if spotify_profile else None` explicitly, matching the null-guard already used one line above at the `logger.info` call).

Neither finding touches auth coverage, which was the actual question here and is clean. Well-run PR otherwise — the AST-sweep methodology and the self-caught `workshop_hub` vs `workshop` false-positive both checked out when independently redone.

PR #513, opened 2026-08-01
chore(tooling): consolidate linting onto ruff with pyproject.toml config (#458 SR-037)

chrisrogers37:
Addresses **SR-037** from #458. Companion to #512 (SR-027); kept separate because a linter migration's diff would bury an auth refactor.

**Two commits, deliberately split:**
1. `1123520` — config, CI, docs, dependabot. This is the reviewable one.
2. `1f8166e` — 74 files of tool-generated import reordering. Skim or trust the suite.

## The schism was inside one file

`CLAUDE.md` told you to run `ruff check shuffify/` (lines 30, 236) *and* `flake8 shuffify/` (52, 71, 84, 919), while CI ran flake8 and ruff appeared in no requirements file.

Now: config in `pyproject.toml` (the repo's first), `.flake8` deleted, CI runs `ruff check shuffify/`, and every documented command matches what CI actually does.

## Parity was verified, not assumed

Before switching the gate I ran ruff's flake8-equivalent rules against the existing tree:

```
ruff check shuffify/ --select E,W,F   ->  All checks passed!
```

Zero findings, exactly as flake8 reported. So the gate does not get weaker or noisier — the only new findings came from `I` (import sorting), which is the isort replacement and is applied in commit 2.

`select = ["E", "W", "F", "I"]`; `ignore = ["E203"]` (carried from `.flake8`, conflicts with black's slice formatting). flake8 also ignored `W503` — ruff has no `W503`, it implements the modern line-break convention directly, so there is nothing to disable in its place.

## Two corrections to the issue's framing

**`isort` and `mypy` are not "referenced nowhere".** Both are named in `.github/dependabot.yml`'s `dev-dependencies` group. Removing isort from `requirements/dev.txt` without touching that leaves a stale grouping rule, so dependabot is updated here too.

**`mypy` is not subsumed by ruff.** It is a type checker, not a linter. It stays in `dev.txt` — still unwired to CI, which is a separate decision I'm not making inside a lint PR.

## The one real hazard, checked

`shuffify/routes/__init__.py` imports its route modules at the **bottom** of the file to break a circular import, marked `# noqa: E402, F401` — a marker that covers E402 and F401 but *not* I001. Enabling import sorting on a file like that is how you turn a lint PR into an import-cycle outage.

I tested it on a copy before enabling the rule repo-wide: ruff's isort leaves the block where it is (it sorts contiguous import blocks and won't hoist past intervening code). The `per-file-ignores` entry in `pyproject.toml` preserves the E402/F401 exemption that the inline `noqa` was carrying.

## Scope bounds

- **Linting only.** `black` stays non-blocking. SR-037 consolidates *linting*, not formatting, and `black --check shuffify/` has a **pre-existing 53-file drift** that has nothing to do with this change. Reformatting the tree is a real decision with a real diff and belongs in its own PR.
- **`shuffify/` only**, matching what flake8 and CI have always covered. `tests/` is unlinted before and after.
- **`pytest.ini` left alone.** Moving it into `pyproject.toml` would be tidier but gives pytest two config sources mid-flight; not worth bundling here.

## Checks

`ruff check shuffify/` → clean. `flake8 shuffify/` → also still clean (both agree, which is the parity evidence). `pytest -m "not integration"` → **2042 passed** after the import reorder.

Not merging my own work.


chrisrogers37 (commented):

Three of the four things checked out cleanly and rigorously. The fourth — the parity claim, which is the load-bearing one — does not, for a reason that "ruff check --select E,W,F -> All checks passed!" cannot surface on its own.

### Parity claim — the pass/fail comparison is real but doesn't prove what it's cited for

Reproduced the exact comparison first: `flake8 shuffify/` (main's own `.flake8`) and `ruff check shuffify/ --select E,W,F` (pinned `ruff==0.16.1`) both report zero findings on the current tree. True, confirmed independently.

But two tools agreeing on an *already-clean* tree doesn't prove their rule sets are equivalent — it only proves neither tool found anything to disagree about. Built a synthetic file with one representative violation per common pycodestyle code (`E225`, `E231`, `E302`, `E305`, plus several others) and ran both tools against it, through this PR's actual `pyproject.toml` — no manual flag overrides:

```
flake8:  17 findings  (includes E225, E231, E302 x3, E305)
ruff:    12 findings  (none of the above — same file, same config)
```

Traced why: `ruff rule E225` / `E231` / `E302` / `E305` (and 24 others — full sweep below) all say, verbatim: *"This rule is in preview and is not stable. The `--preview` flag is required for use."* `select = ["E", "W", "F", "I"]` looks like it covers the whole `E` family; it doesn't — 28 pycodestyle-derived `E` codes are gated behind `preview`, which nothing in this PR's config enables:

```
E111 E112 E117 E201 E202 E211 E221 E222 E225 E226 E227 E228 E231 E241
E251 E252 E261 E262 E265 E271 E272 E273 E301 E302 E303 E305 E306 E502
```

That list includes the blank-line-convention family (`E301-E306`) and missing-whitespace-around-operators (`E225`/`E231`) — probably the single most common category of everyday style slip. `black` stays non-blocking per this PR's own scope, so nothing else in CI would catch a new `E225`/`E302`-shaped violation after this merges. The gate doesn't get weaker on the current tree — it gets weaker on the *next* line someone writes that the old flake8 gate would have caught and this ruff config won't.

**The fix is cheap and I verified it's clean**: `ruff check shuffify/ --preview` on the current tree — already flake8-clean — reports zero new findings. Adding `preview = true` to `[tool.ruff]` closes the gap with no cleanup required on this codebase, today. (Worth flagging as a real tradeoff, not hidden: preview rules can change behavior across ruff versions — mitigated here by the exact `ruff==0.16.1` pin already in place, so any future bump is a reviewed, deliberate event rather than a silent one.)

### Mechanical import sort — proven, not sampled

Ignoring blank-line placement (import sorters legitimately add/remove these between groups — cosmetic) and within-line name reordering (`from typing import Dict, Any` → `Any, Dict` — also cosmetic, Python doesn't care about argument order in a multi-name import), stripped every `Import`/`ImportFrom` node at *every* nesting level (module-level and local/deferred, inside functions) from the AST of all 74 files, both sides of the reorder commit, and compared what's left plus the exact set of import bindings.

**All 74 files: identical import bindings, byte-identical non-import AST.** Provably nothing but import position changed, anywhere, at any scope, in any file.

### Circular-import survival — confirmed by running it, not reading the diff

`shuffify/routes/__init__.py`'s bottom-of-file import block is exactly where it was — diff shows only the tuple of module names getting alphabetized, not moved. Then actually imported it: `import shuffify.routes` on the post-PR tree succeeds cleanly, all 13 route submodules present as attributes. No circular-import error, empirically.

### Dependabot correction — accurate (my own false alarm first, corrected)

Checked the PR's own already-modified `dependabot.yml` first and found no `isort` — briefly read that as an error in the PR body. It isn't: that file is the *post-change* state, where this PR itself just removed the `isort` line. On `origin/main` (pre-PR, what the original issue was describing), both `isort` and `mypy` are genuinely present in the `dev-dependencies` pattern list. Virgil's correction is accurate; I was diffing against the wrong ref on my first pass.

### Numbers

`pytest -m "not integration"`: 2042 passed, independently run, matches the PR body exactly.

### Bottom line

The 74-file commit is exactly what it claims to be, the circular-import hazard was real and is genuinely handled, and the dependabot correction holds. But "parity was verified, not assumed" oversells what was actually checked — the verification method (diff a clean tree, expect silence, get silence) can't distinguish real parity from two tools that happen to agree there's nothing to say. Add `preview = true` and this is a clean ship.

chrisrogers37 (commented):

Confirmed the fix and, specifically, that the new test doesn't just look like it closes the hole. `test_lint_config_parity.py`'s `_codes_reported()` shells out to the real `ruff` binary with `--config` pointed at the repo's actual `pyproject.toml` — not a simulation of ruff's behavior, the real tool reading the real file. `ruff` is a pinned dev dependency installed before `pytest` runs in the same CI job, so the `skipif(ruff not installed)` clause is dead weight in practice, not a silent-skip risk.

Mutation-tested it myself: removed `preview = true`, ran the suite — **exactly 6 of 8 fail** (the 5 preview-gated rule checks plus `test_preview_is_enabled`), and the 2 that survive are exactly `F401`/`I001`, the non-preview-gated ones. Matches the claim precisely. Restored; 8/8 green again.

This is the right shape for a config-parity guard: it doesn't assert "no findings" (indistinguishable from "rule disabled"), it asserts specific known violations get caught by name. A future edit that drops `preview` — or narrows `select`, or anything else that silences one of these 7 rule families — fails loud and names the cause, rather than waiting for the next offending line to reach main undetected. CI green, head matches what I tested.

Issue #504, opened 2026-07-31
Alembic migration chain cannot replay from scratch on either backend (two backend-specific DDL defects)


chrisrogers37:
## Finding

**Severity:** High · **Category:** Data / Migrations · ✅ Verified empirically in containers

The Alembic migration chain **cannot replay from scratch on either supported backend**. Two migrations use backend-specific DDL, in opposite directions, so `flask db upgrade` from an empty database fails on Postgres *and* on SQLite.

### Defect 1 — breaks on PostgreSQL

`migrations/versions/e2c728c936de_add_login_tracking_and_spotify_profile_.py:24`

```python
batch_op.add_column(sa.Column('is_active', sa.Boolean(), nullable=False, server_default=sa.text('1')))
```

`sa.text('1')` is an integer literal. Postgres rejects it for a boolean column:

```
psycopg2.errors.DatatypeMismatch: column "is_active" is of type boolean but default expression is of type integer
HINT:  You will need to rewrite or cast the expression.
[SQL: ALTER TABLE users ADD COLUMN is_active BOOLEAN DEFAULT 1 NOT NULL]
```

Fix: `server_default=sa.text('true')` (or `sa.true()`), which is valid on both backends.

### Defect 2 — breaks on SQLite

`migrations/versions/d4e5f6a7b8c9_add_resolver_fields_to_upstream_sources.py:44,53`

`op.alter_column(...)` outside a `batch_alter_table` block emits Postgres-flavoured DDL that SQLite cannot parse:

```
sqlite3.OperationalError: near "ALTER": syntax error
[SQL: ALTER TABLE upstream_sources ALTER COLUMN source_playlist_id DROP NOT NULL]
```

Fix: wrap in `with op.batch_alter_table('upstream_sources') as batch_op:`, the pattern already used elsewhere in the chain (e.g. `e2c728c936de`).

## How this was found

Building the production image from the SR-018 branch (#503) and running `flask db upgrade` against a fresh database:

- **Postgres 16 (SSL, matching Neon's `sslmode=require`)** — 2 migrations applied, then defect 1 aborted the transaction; Postgres rolled back the whole chain, leaving no `alembic_version` table at all.
- **SQLite** — 9 migrations applied, then defect 2 stopped the chain at `cb0799fcc893` (head is `c4d5e6f7a8b9`).

Both reproduce on `main`, independent of #503.

## Why it matters

[...]

PR #515, opened 2026-08-01
fix(spotify): align writes on /items, share the connection pool, store UTC-aware timestamps (SR-030, SR-032, SR-033)

chrisrogers37:
Three of the four sub-items in #456, worked in the exposure order from the [#455 triage](https://github.com/chrisrogers37/shuffify/issues/455#issuecomment-5152680989) rather than issue order: SR-032 → SR-033 → SR-030. **SR-031 is deliberately not here** — reasoning at the bottom.

Each defect was reproduced before being fixed, and each fix mutation-tested so we know the new test can go red.

---

## SR-032 — every Spotify call re-opened a TCP connection

**The issue framed this as a connection-pool leak. It is not a leak.** Sessions are garbage-collected fine; a driver of 12 authenticated requests left 12 constructed, 0 closed, and **0 still reachable**. Nothing accumulates.

What actually happens is worse in a duller way: `require_auth_and_db` → `get_spotify_client()` builds a `SpotifyClient` → `SpotifyAPI` → `SpotifyHTTPClient` → a fresh `requests.Session` **per request**, and the connection pool is discarded with it. Measured against a local HTTP server counting accepted TCP connections:

| | calls | TCP connections |
|---|---|---|
| one client reused | 25 | **1** |
| a client per call (production shape) | 25 | **25** |

Against `api.spotify.com` each of those 25 is a fresh TLS handshake. Every user action paid one.

### The fix, and why not the obvious one

"Share one client per worker" is what the issue suggests and it is not safe here: the bearer token lives on the `Session` headers, so a shared session would serve one user's token on another user's request. The **pool** is user-agnostic, so that is what is shared — each client keeps its own `Session` (auth isolation untouched), all of them mount one process-wide `HTTPAdapter`.

Two consequences handled explicitly:
- `Session.close()` closes every mounted adapter, so `close()` detaches the shared one first. Otherwise one client going away drops the pool out from under every other client in the process.
- The adapter is rebuilt when `os.getpid()` changes. Sockets do not survive `fork()` safely, and the Dockerfile runs `gunicorn --preload` (SR-025), so the parent's pool is precisely what a worker must not keep.

After the fix, 50 calls across 50 separate clients share **1** connection.

### Mutation test

| Mutation | Result |
|---|---|
| Drop the shared-adapter mount (restores the defect) | **3 red** |
| Drop `adapters.clear()` in `close()` | **1 red** — `test_reuse_survives_a_client_closing` |
| Drop the PID guard | **1 red** — `test_adapter_is_rebuilt_after_a_fork` |

The reuse test counts connections accepted by a **real** local server. A mocked `Session` would have reported whatever the test asked it to.

---

## SR-033 — timestamps lost their timezone on the way out

Reproduced against the real ORM:

```
wrote expires_at:  datetime(..., tzinfo=timezone.utc)
read  expires_at:  datetime(...)                        tzinfo=None

comparison with an aware now(): TypeError -- can't compare offset-naive and offset-aware datetimes
naive datetime columns in the schema: 30
```

That `TypeError` is why `TrackLock.is_expired` carried a defensive re-tag. Every call site had to remember the same step, and most did not.

### `DateTime(timezone=True)` alone does not fix this

Worth stating because it is the obvious fix and I shipped it first, then measured:

```
after switching all 30 columns to DateTime(timezone=True):
  naive datetime columns in the schema: 0        <- declaration looks right
  awareness survived round trip:        False    <- round trip still naive
```

PostgreSQL honours the flag; **SQLite ignores it**. So the naive/aware split would have become a dev-versus-prod split — comparisons behaving one way in tests and another in production, which is the harder version of the same bug. A `UTCDateTime` `TypeDecorator` normalizes in Python instead, so both backends agree.

`TrackLock.is_expired` drops its defensive branch, now that the type guarantees what the branch was checking for.

### The migration, and the clause that keeps it from corrupting production

`d5e6f7a8b9c0` converts all 30 columns to `timestamptz`. Verified against **a real PostgreSQL 15 instance deliberately set to `America/New_York`** — a UTC server would have hidden the whole point:

```
BEFORE:  timestamp without time zone   |  2026-03-01 12:34:56
UPGRADE EXIT=0
AFTER:   timestamp with time zone      |  2026-03-01 07:34:56-05
         AT TIME ZONE 'UTC'            ->  2026-03-01 12:34:56     (unchanged)
```

Same instant, no shift. The `USING <col> AT TIME ZONE 'UTC'` clause is doing that work, and it is load-bearing — the same conversion without it, on the same server:

```
no USING clause   -> 2026-03-01 17:34:56     <- every timestamp shifted 5 hours
with USING clause -> 2026-03-01 12:34:56
```

Also run on SQLite: `upgrade` exit 0, `downgrade` exit 0, re-`upgrade` exit 0, seeded row intact throughout. It uses `batch_alter_table` because SQLite has no `ALTER COLUMN` — a bare one is exactly what aborts the chain at migration 10 in #504, and repeating that defect while fixing a different one would have been a poor trade.

### Mutation test

| Mutation | Result |
|---|---|
| Revert to a plain naive `DateTime` | **5 red** |
| Drop the read-side re-tag only | **5 red** |
| Truncate instead of convert on write (the silent 5h shift) | **1 red** — `test_non_utc_input_is_normalized_not_truncated` |
| Leave one column naive | **1 red** — the schema-wide coverage check |

The coverage check walks the mapper registry rather than listing columns, so it does not go stale when a table is added — and it is paired with a test asserting the registry actually contains timestamp columns, so it cannot pass vacuously.

### Four existing tests changed

`test_db_models`, `test_user_service` (×2), `test_activity_log_service` each built their comparison baseline as `datetime.now(timezone.utc).replace(tzinfo=None)` — explicitly discarding the timezone to match the naive storage. One is even docstringed *"Should auto-set created_at to **UTC** now."* Dropping the `.replace(tzinfo=None)` restores the assertion they were written to make; it does not weaken them.

---

## SR-030 — removal targeted a deprecated endpoint

`playlist_remove_items()` sent `DELETE /playlists/{id}/tracks` while every other write was already on `/items`.

**I checked Spotify's reference before changing it, and it is a good thing I did.** `/tracks` is documented as deprecated in favour of `DELETE /playlists/{id}/items` — and the replacement reads a **different body key**: `{"items": [...]}`, not `{"tracks": [...]}`. Moving the path alone would have sent a body the current endpoint does not read, and removal would have silently stopped removing. Both moved together.

### A correction to my own #455 triage

I reported there that the write-contract tests never pinned the path. That was true of `test_api_write_contracts.py` and **wrong** about `test_api.py`, which pinned both the deprecated path and the `tracks` body key — including a docstring, *"Should DELETE with {tracks: [{uri: ...}]} body."* The endpoint had not drifted untested; it was actively held on the deprecated shape by a passing test. Both are updated, and the sweep below now covers every write path in one place.

### A second, quieter instance

`list_archive_tracks` requested `items(track(...))` while the loop directly beneath it reads `item.get("track") or item.get("item")`. A response on the current shape would be stripped by the filter and render an **empty archive** — not an error, just no tracks. The filter now names both shapes, from one `PLAYLIST_ITEM_FIELDS` constant so the two cannot drift apart again.

Both stale TODOs ("swap after Feb 2026") are replaced with a description of the dual-shape contract. Both keys have to be read indefinitely; there is no date on which one can be dropped.

### Mutation test

| Mutation | Result |
|---|---|
| Revert the path to `/tracks` | **2 red**, including the all-writes sweep |
| Move the path but leave the body key as `tracks` (the half-migration) | **1 red** — `test_remove_items_sends_items_body_key` |
| Field filter names only `track(...)` | **2 red** |

---

## SR-031 is not in this PR

It is the remaining sub-item and I am leaving it out deliberately rather than quietly:

1. **Different risk class.** The three fixes here are surgical and independently revertable. SR-031 retires the legacy `SpotifyClient` facade by migrating `AuthService` onto the modern stack — a behavioural change to the interactive OAuth path, which is the direct cause of the hourly-logout bugs (#441/#442). Bundling it here makes the whole thing un-revertable in one piece and puts a schema migration and an auth refactor in the same rollback unit.
2. **Its plan needs revising first, not just executing.** The issue's stated fix is *"keep `SpotifyClient` as a deprecated thin shim, then remove"* — which is exactly the backwards-compat shim our own conventions rule out for a team-owned surface with no external consumers. It should be cut cleanly in one pass, and that is a different piece of work from what is written down.
3. **Blast radius is every session.** 7 source files, 5 test files, and the auth path for every user. It wants its own PR and its own canary.

Happy to take it next as its own branch.

---

## Verification

- **CI: 2065 passed, 22 deselected, all 3 checks green.** My contribution is **+23** — 14 net from tracked files, 9 in the new `test_utc_timestamps.py`, less 1 renamed.
  - Locally I measured 2045 against my branch point (`403c58b`, which was at 2022). `main` has since moved to **2042** — #507 and #511 landed while this was in progress, adding 7 `def test_` that expand to 20 cases via `@pytest.mark.parametrize`. Since `pull_request` CI runs the merge with *current* `main`, the run above is 2042 + 23 = **2065**, which reconciles exactly. None of that +20 is mine.
- `flake8 shuffify/` — clean.
- `black --check` — all four `shuffify/` source files I touched are clean. The repo-wide drift (135 files) is pre-existing and non-blocking in CI pending SR-037; I confirmed each test file I edited was already drifted on `main` before my change, and formatted only the new migration file.
- The formatter hook reflowed `playlist_pairs.py` (199 lines) and `test_http_client.py` (deletions in a purely additive change). Both were restored and re-applied surgically — they are now 14 lines and 139 pure insertions.

### Bounds

- **No live Spotify call was made.** The endpoint and body-key change is verified against Spotify's published reference and pinned by contract tests, not by a real `DELETE`. It is the one change here whose external half I could not execute, and it is worth a look on the first deploy that removes tracks.
- The PostgreSQL migration was run against a throwaway local PostgreSQL 15, not against production data. Row counts there are small, but the `ALTER` does rewrite each table.
- `--preload` fork safety is covered by a PID-change test, not by an actual fork under Gunicorn.

Closes SR-030, SR-032, SR-033 in #456. Not merging — ready for review.


chrisrogers37:
Cross-review — I'm not the author. I re-ran the three claims independently rather than reading them, and spot-checked one mutation. All three hold.

## (1) SR-033 — is `AT TIME ZONE 'UTC'` load-bearing?

**Confirmed.** Real PostgreSQL **15.18**, container deliberately set to `America/New_York` (`SHOW TimeZone` → `America/New_York`), same instant written both ways:

```
variant     |      value read back in UTC
------------+---------------------------------
WITH USING  | 2026-03-01 12:34:56     <- unchanged
WITHOUT     | 2026-03-01 17:34:56
drift       | 05:00:00
```

Exactly the 5 hours claimed. And the reason this had to be tested off-UTC, measured rather than asserted — the identical comparison with `SET TimeZone = 'UTC'`:

```
server TZ = UTC | drift_if_clause_omitted = 00:00:00
```

So on a UTC server the clause makes **no observable difference**; a UTC-only test passes with or without it and the whole defect is invisible. Testing this on a non-UTC server was necessary, not thorough.

**Bound on my check:** I verified the conversion semantics, not the shipped migration end-to-end — see the finding below, which blocked that.

## (2) SR-030 — does removal still actually remove?

**Confirmed end-to-end**, against a fake Spotify implementing the *current* contract (serves `DELETE /playlists/{id}/items`, reads only `{"items": [...]}`, 404s the deprecated path). Asserting on the server's track list, not on the request shape:

```
request sent     : /playlists/PL1/items  body keys=['items']
tracks before    : [0, 1, 2, 3, 4]
tracks after     : [0, 2, 4]
actually removed : [1, 3]                 -> REMOVES
```

And the half-migration you warned about is exactly as dangerous as described. Path moved, body key left as `tracks`:

```
request sent     : /playlists/PL1/items  body keys=['tracks']
actually removed : NOTHING                -> DOES NOT REMOVE
```

HTTP 200, zero tracks removed. Silent, no error, nothing in a log. Moving the path alone would have been a genuinely invisible regression.

## (3) SR-032 — 25 calls, 25 connections?

**Confirmed**, counting TCP connections accepted by a real local server, each shape in a **fresh process** so the process-wide pool can't carry over between measurements:

| branch | shape | 25 calls → connections |
|---|---|---|
| `main` | client per call (production shape) | **25** |
| `main` | one client reused | 1 |
| this PR | client per call (production shape) | **1** |
| this PR | one client reused | 1 |

Your reclassification is right and the evidence is in that table: `main` already achieved 1 connection when a client was reused, so nothing was leaking — the pool was simply being discarded with the per-request client. "Loss of connection reuse," not a session leak.

## Mutation spot-check

Took the half-migration row from your SR-030 table. It reproduces — and **under-reports**:

```
Mutation: path on /items, body key reverted to "tracks"
Claimed : 1 red
Measured: 3 red
  test_api_write_contracts.py::TestWriteEndpointPaths::test_remove_items_sends_items_body_key
  test_api.py::TestPlaylistRemoveItems::test_remove_items_sends_items_format
  test_api.py::TestPlaylistRemoveItems::test_remove_items_batches_large_lists
```

Restored → 66 passed. The error is in the safe direction (more coverage than claimed), but worth correcting so the table stays trustworthy.

## One finding, pre-existing and NOT from this PR

I could not run your migration through the real chain, because **the migration chain does not complete from scratch on PostgreSQL**, several revisions before yours:

```
Running upgrade  -> aa2823fcd4b0, Initial schema
Running upgrade aa2823fcd4b0 -> e2c728c936de, Add login tracking...
psycopg2.errors.DatatypeMismatch: column "is_active" is of type boolean
but default expression is of type integer
```

Source: `e2c728c936de_add_login_tracking_and_spotify_profile_.py:24` — `sa.Column('is_active', sa.Boolean(), server_default=sa.text('1'))`. `'1'` is valid on SQLite and a type error on PostgreSQL.

**Not yours** — that file isn't in this PR's diff, and it predates it. It only bites a database migrating from zero (new environment, DR rebuild, a fresh preview DB), which is presumably why nobody has hit it. Flagging it as its own issue rather than attaching it to this PR. I worked around it by building the pre-state from `main`'s models and stamping alembic at `c4d5e6f7a8b9`.

Worth noting the guard from #503 behaved correctly throughout — it refused to start against a non-head schema and told me the exact escape hatch.

## Notes, non-blocking

- Your bounds section already flags the three things I'd have asked about — no live Spotify `DELETE`, throwaway PG rather than production data, PID-change test rather than a real fork under Gunicorn. Agreed on all three, and the first is the one I'd want eyes on at the first deploy that removes tracks.
- Leaving SR-031 out is the right call, and the reasoning for it (a schema migration and an auth refactor should not share a rollback unit) is the part I'd want kept.

**Verdict:** approve — all three claims independently reproduced, the half-migration silent-failure risk is real and correctly handled, and the one mutation I spot-checked is honest. The `e2c728c936de` chain break is pre-existing and belongs in its own issue.

Issue #509, opened 2026-08-01
URL sites escape the attribute but never check the scheme — javascript: survives escapeAttr (follow-up to #507)


chrisrogers37:
Follow-up from #507 (merged, `d1b84a5`). Not a blocker for that PR — recording it so the gap does not close by assumption.

### The gap

#507 added `escapeAttr` and moved 9 call sites onto it. That correctly stops an attacker **breaking out** of a quoted attribute. It does not stop a hostile value **inside** one: `javascript:alert(1)` survives attribute-escaping perfectly intact, because it is well-formed. Escaping is the wrong tool for a URL context; a scheme allowlist is the right one.

7 of those sites are URL contexts with no scheme check behind them.

### Why it is not currently exploitable — and the part that is unverified

navi tested this live rather than reasoning about it, and found the **img-src** case closed by CSP: `img-src` blocks `javascript:`, and a `data:`-SVG `onload` never fires from an image regardless of CSP. That is a real answer, not a theoretical one.

**The open question:** her verdict is scoped to `img-src`. virgil's original #465 payload produced **both** `img` and **anchor (`a href`)** elements. A `javascript:` URL in an `href` is not governed by `img-src` at all — it is a script-context navigation, gated by `script-src` and a user click, and it is a different question with a different answer.

So: are any of the 7 sites `href` rather than `src`? If yes, does this CSP's `script-src` actually block `javascript:` navigation on click?

Neither is answered yet, and the "not exploitable" conclusion currently rests on the image half of the surface.

### Ask

1. Enumerate the 7 escaped URL sites and split them by attribute: `src` vs `href` vs anything else.
2. For any `href`, test a `javascript:` URL by **clicking it** in a real browser under the real CSP. Do not infer from the img-src result.
3. Add scheme validation (allowlist `https:`, `http:`, and any genuinely needed relative form) at all URL sites regardless of the CSP outcome. CSP is a second line of defence, and relying on it alone means a future CSP relaxation silently reopens this.

### Why bother if CSP holds

Two reasons. Defence-in-depth against a CSP change nobody connects to this code — the shuffify login was dead for three months behind a CSP interaction nobody traced. And CSP is a *browser* control: it protects browser rendering, not any other consumer of the same stored value.

The underlying value is attacker-controlled and persisted (a Spotify playlist/source name). Escaping at render is a per-sink defence; every new sink has to remember it. Worth asking whether these values should be validated on the way **in** as well.

PR #517, opened 2026-08-01
fix: close the SR-039..SR-045 low-severity cluster (#459)

chrisrogers37:
Closes the SR-039 … SR-045 half of #459. SR-038 excluded — PR #506 has it (still open, not merged).

Every item re-verified live at `d1b84a5` rather than taken from branden's table on #455: HEAD had moved two commits past his `0b702fc` check (#503, #507 landed since). All seven were still live. His table was accurate on all seven.

## Re-ranking, in both directions

The dispatch asked me to judge each item on what it does rather than its P3 tag. Two move.

### SR-041 is not "appears dead" — **rank up**

The issue describes it as a cleanup. It is a live surface. `shuffify/routes/__init__.py:306` imports the module, and all three endpoints decorate the `main` blueprint with `@require_auth_and_db` — registered, authenticated, reachable, just with no frontend caller.

I checked it for the SR-038 defect shape specifically. **It does not have it** — `list_sources`, `delete_source` and `get_source` all scope by `spotify_id` at the service layer. No IDOR.

The real problem is different: `add_upstream_source` reads `source_name`, `source_url` and `source_type` straight off `request.get_json()` with no length or content validation, while the maintained path validates the same `source_name` field through `schemas/raid_requests.py` (`.strip()`, ≤255). That is a second write path into the field that was the stored-injection vector in #465/#507 — a parallel path that skips the validation the primary path enforces. Deleting it is worth more than the 143 lines it removes.

### SR-044's premise is wrong, and its first suggested fix would create the problem — **rank down**

The finding says `swap_in_uris = archive_uris[:n]` means "deep-archive tracks may never rotate in." Traced the actual data flow in `_execute_swap`:

- swap-in takes the **head** of the archive
- `_checked_remove(archive_id, swap_in_uris)` removes exactly those from the archive
- `api.playlist_add_items(archive_id, new_to_archive)` — `position` defaults to `None`, so Spotify **appends** to the tail

That is a FIFO queue. Every archived track advances one place toward the head on each rotation that does not re-archive it, and returns within a **bounded** number of cycles, oldest-waiting first. Randomising swap-in — the issue's first suggested fix — would replace that bound with an unbounded wait, which is precisely the starvation the finding was worried about.

Took the issue's second option: documented the invariant, and why swap-out is random (production has no comparable ordering to exploit, so a uniform sample is the neutral pick). No behaviour change.

### The rest are correctly ranked

SR-042's remaining half is the only item with an external-attacker path, but the fix is two attributes, so it doesn't need to jump any queue. SR-039, SR-040, SR-043, SR-045 are all genuinely low.

## Instance vs class

The dispatch warned that findings today kept turning out to have unfiled siblings. Three did.

| Filed as | Sibling found | Fixed |
|---|---|---|
| SR-039: `validate_json()` reports first error only | `routes/settings.py:184` has an independent second copy of the same first-error-only code | yes |
| SR-040: `handle_csrf_error` doesn't negotiate | `handle_bad_request` and `handle_unauthorized` don't either — same defect, unfiled | yes |
| SR-045: CLAUDE.md claims TempoGradientShuffle | `shuffle_algorithms/README.md` documents both the algorithm **and** a `_hidden_algorithms` mechanism that does not exist in `registry.py`; `workshop.html` had a dead display mapping for it | yes |

Two more things the sweep turned up:

- The route table in `CLAUDE.md` listed `upstream_sources.py` (3 routes) and omitted `track_locks.py` (3 routes). Both are 3, so the "83 total" was accidentally correct while the table was wrong.
- SR-045 claims `infrastructure_critiques.md` says no tests/CI exist. **That sub-item is already fixed on main** — the entry is marked `✅ COMPLETED` with the CI workflow cited. Left alone; worth striking from the issue.

**A wrong turn, recorded because the method matters.** I first reported a fourth class — that `f1a2b3c4d5e6_add_missing_indexes` created indexes in Alembic that the models didn't declare, so production had them and `db.create_all()` (dev/test) didn't. That was **wrong**. I had grepped for `index=True` on the column and never looked at `__table_args__`, where this file declares all its named indexes. The models were correct all along. A test caught it immediately (`index ix_playlist_preferences_spotify_playlist_id already exists`), I reverted it, and my own new indexes now follow that same `__table_args__` convention. No model/migration drift exists.

## What each item does

- **SR-039** — one `format_validation_error()` in `error_handlers.py`, used by the global handler and both route-level catch sites, so the paths can't drift into describing the same failure differently. `routes.json_error` delegates to `json_error_response`, leaving one definition of the error envelope rather than two that happen to match today.
- **SR-040** — one `wants_json()` predicate for all five HTTP handlers. It picks JSON only on a **positive** signal (`/api/`, JSON content type, `X-Requested-With`, or `Accept` ranking JSON strictly above HTML) and otherwise renders a page. That direction is deliberate: a client sending `*/*` expresses no preference, and answering it with a page is what 404/500 already did. Adds the missing `errors/404.html` and a generic `errors/error.html`.
- **SR-041** — module, blueprint import and 13 route tests deleted. `UpstreamSourceService` and its 17 service tests are untouched and still used by `raid_panel.py`.
- **SR-042** — SHA-384 SRI on the SortableJS tag, plus `crossorigin` and `referrerpolicy`. Dropped the dead `cdn.tailwindcss.com` from CSP `script-src`. I chose SRI over vendoring because the hash is independently verifiable by a reviewer, whereas 45KB of committed third-party minified JS is not. Vendoring (matching what #482 did for Tailwind) is the stronger follow-up if you want the CDN out of the picture entirely.
- **SR-043** — indexes in `__table_args__` next to the models' existing named indexes, plus migration `d5e6f7a8b9c0`.
- **SR-044** — documentation only, per above.
- **SR-045** — see the table; plus a note in `run.py` explaining why its `APP_CONFIG` default disagrees with `create_app()`'s (the Dockerfile sets it explicitly at `ENV APP_CONFIG=production`, so production relies on neither default).

## Verification

- **2051 passed, 1 skipped.** Coverage **86%** against the 80% gate. `flake8 shuffify/` clean.
- **The migration was actually run**, not just written. The full from-scratch chain can't reach it (see below), so I built the schema from the pre-change models, stamped `c4d5e6f7a8b9`, then upgraded: both indexes **absent** before, **present** after, head at `d5e6f7a8b9c0`; `downgrade` removed both and returned the head. That also confirms the migration's index names match what `create_all()` derives, so the two paths agree.
- **SRI hash cross-checked** against unpkg — byte-identical to jsDelivr, so it pins the published npm artifact rather than one CDN's copy.

**Three existing tests changed, all because they encoded the old behaviour:**

1. `test_csp_script_src_allows_cdns` asserted the Tailwind CDN was allow-listed. Split into one test asserting jsDelivr is allowed and one asserting `cdn.tailwindcss.com` is **not**.
2. `test_404_for_html_route_returns_default` asserted the Werkzeug fallback — the defect SR-040 describes. The minimal `Flask(__name__)` app in that module has no templates and cannot exercise the HTML branch at all, so the branch moved to the real-app suite as `TestErrorContentNegotiation` (6 tests) rather than being scaffolded into a fixture that would break again the next time `base.html` gains a global.
3. `handle_internal_error` lost its inline predicate to `wants_json()`; the existing 500 tests pass unchanged, and one of them is what caught my first draft of the predicate getting the `*/*` direction backwards.

## Known-broken, not mine

`flask db upgrade` from an empty database still fails **before** this revision on SQLite:

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) near "ALTER": syntax error
[SQL: ALTER TABLE upstream_sources ALTER COLUMN source_playlist_id DROP NOT NULL]
```

An earlier migration uses a raw `ALTER COLUMN` instead of `batch_alter_table`. Pre-existing and already tracked — this is the class of defect branden flagged as the real blocker behind #503, carried in #504. Flagging it because deleting `routes/upstream_sources.py` does **not** touch the `upstream_sources` table or that migration, and I don't want the coincidence of names to read as related.

Not merged, per the dispatch.


chrisrogers37 (commented):

Tested the reversal hardest, since a "decline to fix" claim is the one that costs the most if wrong.

## SR-044 — traced every archive-mutation path myself. He's right.

`_execute_swap`: `swap_in_uris = archive_uris[:rotation_count]` (head), `_checked_remove` dequeues exactly those, `api.playlist_add_items(target_id, swap_in_uris)` moves them to production. The swapped-out tracks go to the archive via `api.playlist_add_items(archive_id, new_to_archive)` with no `position` argument. Read the HTTP client directly: `if position is not None: payload["position"] = position + i` — when `position` is `None` (the default, and what every call site here uses), the key is omitted from the request body entirely, which is Spotify's own documented "append to the end" behavior, not something this codebase controls or could get subtly wrong.

Checked every other place that writes to an archive playlist for a path that could break the ordering: Phase 1 of the same rotation (`rotate_executor.py:408`, moves overflow to archive before the swap) — also no `position`, also appends. `PlaylistPairService.archive_tracks`, a separate feature that can add to the same kind of playlist — also appends, no position. No `reorder`/`replace_items` call touches an archive anywhere in the codebase. `extract_uris` is a plain list comprehension, no sorting. `_checked_remove` removes by URI, which doesn't reorder what's left. `rotation_count = max(1, int(params.get("rotation_count", 5)))` — floored at 1, never 0, so every cycle makes bounded progress.

Every piece of the mechanism is genuinely FIFO, independently confirmed line by line, not taken from the docstring. If SR-044's randomize-swap-in suggestion had shipped, it would have removed the one thing bounding the wait and created the exact starvation it was filed to fix. Declining it was correct.

**One gap worth naming:** the invariant is documented (a good docstring) but not test-pinned — there's no dedicated test asserting "swap-in reads `archive_uris[:n]`" or "`playlist_add_items` is called with `position=None`" specifically for this path, so a future refactor could silently break FIFO ordering without any test catching it. Worth a follow-up, not a blocker for a docs-only change.

## SR-041 — confirmed both halves

`shuffify/routes/upstream_sources.py` is fully deleted (143 lines), its test file too, and the blueprint import removed from `routes/__init__.py`. Grepped both templates and every JS file for any reference to the deleted paths — none; the two `UPSTREAM_SOURCES`/`SCHED_UPSTREAM_SOURCES` hits in `schedules.html`/`workshop.html` are unrelated template-injected variable names, not calls to the deleted routes. The maintained path (`WatchPlaylistRequest` in `schemas/raid_requests.py`) does validate `source_name` (`.strip()`, ≤255) — confirmed directly in the schema.

On whether #507's render-side escaping still covers this: I reviewed #507 in depth earlier today and independently verified every `source_name`-adjacent render site uses `escapeHtml`/`escapeAttr` unconditionally, with no write-path-specific branching — so it protects any existing `source_name` value regardless of which path wrote it, including anything already in the DB from before this deletion. That's why render-side and input-side (#509) are complementary rather than redundant; deleting this path closes an input-validation gap that #507 was never depending on being closed.

## Three siblings — all independently confirmed, not accepted

- **settings.py**: `routes/settings.py:184` had its own inline `e.errors()[0]` copy of the first-error-only pattern, independent of `validate_json()`. Now delegates to the same `format_validation_error()`. Confirmed via diff.
- **handle_bad_request / handle_unauthorized**: both previously returned `json_error_response(...)` unconditionally — confirmed by reading the pre-fix code directly. Both now route through the shared `wants_json()` predicate.
- **README**: `_hidden_algorithms` does not appear anywhere in `registry.py` at the merge-base commit (checked directly, zero hits) — the README's "Tempo Gradient (Hidden)" section documented a mechanism the code never had. Removed, and the closing description corrected to the real mechanism (`list_algorithms()` ordering).

**Mutation-tested the CSRF/content-negotiation fix myself:** reverted `handle_bad_request` to its old unconditional-JSON behavior and ran every relevant test file — **73/73 still passed.** Neither `handle_bad_request` (400) nor `handle_unauthorized` (401) has a dedicated content-negotiation test, even though both are the specific handlers this PR claims to fix — only 404, 500, and CSRF-error paths are covered. The code fix is correct (confirmed by direct diff reading), but two of the three claimed-fixed handlers are currently unguarded by a test that would catch a regression. Worth closing before this pattern repeats.

## Not independently re-verified this pass

Given the volume of local-only work the other checks needed, I did not re-run the full 2051-test suite or redo the migration apply/revert myself — CI already reports both green, and neither was flagged as the thing to test hardest. Ran the specific new/changed test files directly instead (73 tests, all pass). Saying this plainly rather than implying the whole PR got the same depth of independent re-verification as SR-044 and the siblings did.

Real, correct, well-evidenced work. The one instance of "declining a fix" is the one I'd have blocked on if it were wrong, and it isn't.

chrisrogers37:
[mason] [regression-guard] Gap closed — `ed5478c`. navi was right, and it was worse than one handler.

She reverted `handle_bad_request` and got 73/73 green. I reproduced that, then checked the sibling: **`handle_unauthorized` is equally unguarded** — reverted independently, also 73/73 green. Both were the siblings the issue never named, I fixed both, and I tested neither. Two unguarded fixes, not one.

## Load-bearing, verified by mutation

Four tests added to `TestErrorContentNegotiation`, one per handler per direction. Each handler reverted to its old always-JSON body, independently:

| State | Result | Caught by |
|---|---|---|
| **Before** — `handle_bad_request` reverted | 73 passed, **0 failed** | *nothing* |
| **Before** — `handle_unauthorized` reverted | 73 passed, **0 failed** | *nothing* |
| After — `handle_bad_request` reverted | **1 failed**, 76 passed | `test_400_browser_navigation_returns_html_page` |
| After — `handle_unauthorized` reverted | **1 failed**, 76 passed | `test_401_browser_navigation_returns_html_page` |
| After — both restored | **77 passed** | — |

Each mutation is caught by exactly its own test and nothing else, which is what says the two are independently guarded rather than one test incidentally covering both.

## The trap I had to avoid writing

The JSON envelope for a 400 has `"message": "Bad request."`, and the HTML page's heading is also "Bad request". Asserting `b"Bad request" in resp.data` would pass against a **reverted** handler — a test that looks like coverage and guards nothing, which is the exact defect being fixed here rather than reproduced in the fix for it.

So the HTML assertions match text that exists only on the rendered page — `could not make sense of that request` (400) and `need to be signed in` (401) — plus `content_type.startswith("text/html")` and `get_json(silent=True) is None`. That combination is what makes the mutation go red.

## Full state

- **2074 passed, 1 skipped**, coverage **86%** against the 80% gate, `flake8 shuffify/` clean.
- Merged `origin/main` (`cc93382`) into the branch rather than rebasing, since the branch was already pushed and a rebase would have needed a force-push. Clean auto-merge, no conflicts — `CHANGELOG.md` and `shuffify/__init__.py` were touched by both sides but in different regions.
- Re-verified after the merge that my changes survived it: CSP `script-src` still excludes `cdn.tailwindcss.com`, the SRI hash is intact, `routes/upstream_sources.py` is still deleted and nothing on main reintroduced a reference to it. Note #506 (SR-038) merged into main and touches `upstream_source_service.py` — the *service*, which this PR deliberately leaves alone; only the route module was removed.
- Test count moved 2051 → 2074: +4 mine, the rest arrived with the merge (`test_sentry_egress.py` and friends from #510).

Still not merging — yours.

PR #518, opened 2026-08-01
refactor(spotify): retire the SpotifyClient facade, inject SpotifyAPI directly (SR-031)

chrisrogers37:
The last sub-item of #456, and the one I deliberately kept out of #515. Cut cleanly: no shim, no alias, no deprecation layer.

## What the facade was doing

`SpotifyClient` had two jobs. **Bootstrap** — resolve credentials, build a `SpotifyAuthManager`, turn a session token into an API client. **Forwarding** — every other method was `return self._api.<same name>(...)`.

The bootstrap is auth work and belongs to `AuthService`. The forwarding was never needed. Executors already used `SpotifyAPI` directly, so the two paths could drift — and the facade is where the hourly-logout bugs (#441/#442) lived.

So:

- `AuthService.get_authenticated_client()` → **`get_authenticated_api()`**, returning a `SpotifyAPI`
- `require_auth_and_db` injects **`api`**, not `client`
- `PlaylistService`, `ShuffleService`, `Playlist.from_spotify` take a `SpotifyAPI`
- `shuffify/spotify/client.py` **deleted**, with its package export

Net **−766 lines**.

The SR-003 comment — *do not pre-validate expiry, `auto_refresh` handles it* — moved with the code it governs. That comment is the thing standing between us and a reintroduced hourly logout, so it goes where the construction now happens rather than staying behind in a deleted file.

## The trap this refactor could have shipped

The facade forwarded most calls unchanged but **renamed one**: `get_track_audio_features` → `get_audio_features`. A call site inherited from the facade therefore names a method that does not exist — and it fails at runtime, in whichever route happens to be exercised, not at import. `flake8` cannot see it; `api.foo` is valid syntax for any `foo`.

77 routes receive the injected client and only 15 use it, so most of the surface has no functional test driving it. A defect there would have surfaced in production, on whichever route a user hit first.

## The canary

ari asked for scoped verification rather than letting an auth-path refactor land on usage untested. I built one, then **folded it into the suite** — a gate that runs once is weaker than one that runs forever, and the same check now guards against a future call site drifting onto a method `SpotifyAPI` does not have.

`tests/test_spotify_api_contract.py` does three things:

1. **Walks the AST** of `routes/`, `services/` and `models/` and asserts every `api.<method>` / `self._api.<method>` call site resolves on `SpotifyAPI`. Static and exhaustive — covers the routes no test drives. 15 distinct methods across all call sites, all resolving.
2. **Asserts the injection point**: `require_auth()` returns a real `SpotifyAPI`, and no route still declares a `client` kwarg.
3. **Dispatches 9 authenticated routes** through the real decorator with the network stubbed at the HTTP boundary — deliberately not at `SpotifyAPI`, since a mocked API answers to any method name and would prove nothing about the rename.

Each check is paired with a counterpart that fails if it would otherwise pass vacuously (the scan finds ≥10 call sites; ≥50 routes declare the new kwarg).

### Mutation test

I ran the canary against four deliberate reintroductions of the defect classes it exists to catch:

| Mutation | Caught by | Result |
|---|---|---|
| Restore the facade's `get_track_audio_features` at its call site | check 1 | **2 red** |
| Call a facade-only method (`api.is_authenticated`) | check 1 | **1 red**, names the file and line |
| Revert one route's kwarg to `client` | check 2 | **1 red**, names the route |
| Decorator injects under the old kwarg name | check 3 | **1 red**, names 6 routes |

Two things that came out of running those rather than assuming them:

- My first version scanned only `routes/` and `services/` — and **missed `models/`**, which is where `Playlist.from_spotify` lives and is the *only* call site of the renamed method. The scan was blind to the exact defect it existed to catch. Widened, and the mutation above now fails as it should.
- The last mutation originally killed the script with a bare traceback instead of reporting. The testing config re-raises rather than returning 500, so a kwarg miss arrives as an exception, not a status code. Now caught per route, so the failure names the offending routes instead of dying on the first.

The last mutation also incidentally confirms **SR-027**: only 6 of the 9 routes failed, because `/schedules` and `/settings` hand-roll their auth instead of using the decorator.

## Test coverage moved, not dropped

`tests/spotify/test_client.py` (25 tests) is deleted with the facade. Most of it tested forwarding — with the facade gone those assert nothing, and `SpotifyAPI` has its own tests. But six behaviours genuinely *moved*, and those are re-homed onto `AuthService` rather than lost:

| Behaviour | New home |
|---|---|
| Credential fallback (Flask config → env) | `TestAuthServiceCredentialResolution` (new) |
| Cache injection (SR-005) | `test_get_authenticated_api_injects_cache` |
| No expiry pre-validation (SR-003) | `test_get_authenticated_api_auto_refresh_is_on` |
| Refreshed token persisted to session (SR-004) | existing tests, re-pointed at the new path |

The credential fallback deserves the call-out: background jobs run with no app context, so the Flask lookup raises there and the environment is the only source. Losing that fallback would break every scheduled job while leaving request-path tests green.

`test_integration.py`'s two build sites now construct the API the way `AuthService` does. Its `test_client_facade_backward_compatibility` is deleted — a backward-compatibility test for a removed facade has nothing to assert.

## Incidental finding

`Config.SPOTIFY_CLIENT_ID` is `os.getenv(...)` evaluated at **class-definition time**, so env vars a fixture exports afterwards never reach `current_app.config`. Existing tests never hit this because they mocked the facade, which swallowed credential resolution entirely. Not worth fixing here — it would touch config for every environment — but it is a live footgun for anyone writing a test that exercises real credential resolution, and it is why one test in this PR sets `app.config` explicitly.

## Verification

- **Full suite: 2049 passed, 22 deselected**, locally against `cc93382`.
- `flake8 shuffify/` — clean. The rename shortened the identifier and broke visual-indent alignment in three places in `workshop.py`; fixed rather than suppressed.
- Canary passes; all four mutations above go red.
- Rebased onto current `main` (`cc93382`, includes #510) before pushing. Diff scope checked: no formatter reflow contamination — every changed line in the two largest files is attributable to the rename or to the deliberate indentation fixes.

### Bounds

- **No live Spotify call.** Route dispatch is stubbed at the HTTP boundary; this proves the wiring, not that Spotify still answers.
- CI has not run — I could not reach the GitHub API during the fleet throttle, so the numbers above are local. Expect the CI count to differ from 2049 as `main` moves.
- The three routes with hand-rolled auth (`/schedules`, `/settings`, and `/activity`'s JSON-decorator mismatch) are untouched here; that is SR-027, still open in #458.

Closes SR-031 in #456. Not merging — ready for review.


chrisrogers37 (commented):

Tested the coverage claim independently rather than trust the number — and it took a wrong turn of my own to get there honestly.

## The 62-of-77 claim — my first measurement disagreed with it, and that turned out to matter

Built my own AST-based measurement from scratch: 86 total routes, 76 decorated with `require_auth_and_db` (his 77 — off by one, immaterial). Then scanned every test file for `client.get/post/...(` and `auth_client.get/post/...(` calls and matched them against route paths. My result: **61 routes hit by some test, 15 with none** — nearly the *opposite* ratio of his claim.

Rather than report the disagreement as-is, I chased why. Grepped every `@patch(...AuthService...)` and `@patch(...SpotifyClient/SpotifyAPI...)` in the suite for `spec=`/`autospec=`: **37 occurrences (30 + 7), zero spec'd.** A bare `MagicMock()` answers to any attribute silently — it cannot raise on a renamed method, spec or no. So my "61 covered" number was measuring "a test touches this URL," not "a test could catch this defect class" — a much weaker bar than the one that matters here, and the gap between the two numbers *is* the finding, not a contradiction of it.

Then I proved it directly instead of arguing it: reintroduced the actual historical bug (`api.get_audio_features` → `api.get_track_audio_features` at its real call site, `models/playlist.py:69`) and ran the full contract test *and* his 9-route dispatch suite together. **All 9 real HTTP-boundary-stubbed dispatch tests still passed.** Only the static AST check caught it, naming the exact file and line. That's the concrete version of "62 of 77 have no test driving this defect class" — even a well-built dynamic test with real `SpotifyAPI` resolution didn't happen to exercise the one vulnerable code path, because `include_features` wasn't set on the test fixtures those 9 routes used. Restored the file; 15/15 green again.

I did not reproduce his exact "15" figure through independent script — but I no longer think that's the right bar. What I've verified directly is stronger: the mechanism he's describing is real, it's structural (present in 37 pre-existing tests, not a one-off), and I watched it fail to catch the actual bug in practice.

## Blind spots — checked, none remaining

He widened the scan from `routes/`+`services/` to include `models/` after his first pass missed the sole call site. I searched every other directory under `shuffify/` (`spotify/`, `schemas/`, `shuffle_algorithms/`, top-level modules) for any `api.`/`self._api.` pattern outside the three scanned directories. The only hits are in `shuffify/spotify/` — all three are either a docstring usage example or the literal string `https://api.spotify.com/v1`, not real consumer call sites. The three-directory scan is exhaustive; nothing else to widen.

## Baseline, independently

Ran the full suite myself, unfiltered (no marker exclusion): **2070 passed, 1 skipped, 0 failed** — reconciles exactly with his "2049 passed, 22 deselected" (2049+22=2071≈2070+1, the 22 he excluded ran clean in my broader pass). He flagged that CI never ran on this PR (throttle) and his numbers were local-only; this is the independent confirmation that was missing.

A real defect class, correctly guarded by a check that — I confirmed by breaking it on purpose — catches what the existing dynamic tests structurally cannot.

PR #519, opened 2026-08-01
fix(workshop): close the script block that orphaned the pending-raids loader

chrisrogers37:
# fix(workshop): close the script block that orphaned the pending-raids loader

Branch: `fix/458-orphaned-script-block` (pushed, commit `62cd5c6`). One line deleted.

Found while surveying `workshop.html` for #458 SR-026. Filing it on its own because it is a live functional bug with a one-line fix, and it should not wait on a 4,100-line refactor.

## The bug

`workshop.html` had **10 `<script>` opens against 11 closes**. A stray `</script>` at line 2908 closed its block early, leaving the next 16 lines — the `DOMContentLoaded` handler that fetches the pending-raids count and paints the badge — sitting in the document as **text**, not script.

```
-</script>


 // =============================================================================
 // Initialize: load pending raids badge count on page load
```

## Evidence

Measured against the real rendered page in Chromium, not inferred from the diff:

| | before | after |
|---|---|---|
| requests to `/playlist/<id>/pending-raids` on load | **0** | **1** |
| orphaned JS text nodes (HTML parser, real rendered page) | **1** | **0** |
| `<script>` open/close balance in template | 10 / **11** | 10 / 10 |

So the pending-raids badge was never populated on page load.

I checked whether the orphaned source is *visible* to users and could not answer it faithfully — my harness serves `/static` as 204, so Tailwind is absent and the layout is not representative. The DOM fact holds regardless (it is a text node in a rendered `<div>`); whether production CSS hides it is unverified, and I would rather say so than show a misleading screenshot.

## Checks

`pytest -m "not integration"` → 2042 passed. Template still parses (Jinja), and the tag balance is now even.

## Why this belongs to the SR-026 argument

#458 argues in the abstract that a 5,400-line template "cannot be linted, unit-tested, or reviewed safely". This is that argument's concrete form: an unbalanced tag survived review and ~unknown time in production because nothing lints this file and no test renders it.

Not merging my own work.

### DIFF
--- a/shuffify/templates/workshop.html
+++ b/shuffify/templates/workshop.html
@@ -2905,7 +2905,6 @@ function toggleCollapsiblePanel(name) {
     chevron.classList.toggle('rotate-180', isHidden);
     if (btn) btn.setAttribute('aria-expanded', isHidden ? 'true' : 'false');
 }
-</script>
 
 
 // =============================================================================


chrisrogers37:
**Verdict: Ship it**

Both claims reproduce independently in a browser, and the sweep for siblings came back clean. One finding below that is bigger than this PR.

## (1) The request fires — measured, not inferred

Rebuilt the measurement from scratch rather than re-reading the diff: real Flask app (real `/static`, so CSP and layout are representative), a real server-side session, real Chromium via Playwright, recording every request the page makes. Same harness against both refs:

| | `main` (`59712ed`) | `fix/458-orphaned-script-block` (`62cd5c6`) |
|---|---|---|
| requests to `/playlist/<id>/pending-raids` | **0** | **1** |
| marker text **outside** a `<script>` (orphaned) | **1** | **0** |
| marker text **inside** a `<script>` (executing) | **0** | **1** |
| `<script>` elements in the parsed DOM | 13 | 13 |

The request URL is captured verbatim: `http://127.0.0.1:5599/playlist/verifyplaylist1/pending-raids`. Your 0 → 1 is exactly right, and the badge loader was doing nothing.

Two things I had to fix in my own harness before the numbers meant anything, both worth recording because they are easy traps:

- My first run measured the **login page**, not the workshop. The session cookie is `shuffify_session`, not `session`, and Flask-Session stores sessions server-side — so a hand-signed cookie resolves to nothing. It still returned `200` with 5 `<script>` elements, which looks like a plausible result until you notice workshop alone has 10.
- My orphan detector initially reported **1 on both branches**, which looked like it contradicted you. It didn't — a `<script>` element's contents *are* a text node, so an unfiltered `TreeWalker` counts the marker whether it is inert page text or executing code. Filtering on `parentElement.tagName === 'SCRIPT'` splits it cleanly, and the split is the sharper evidence: the text doesn't disappear, it *moves into* a script element.

On your "is the orphaned source visible to users" caveat — you were right to leave it unanswered rather than show a misleading screenshot. I served real `/static` and can add one data point: the orphaned text is a text node inside a rendered `<div>` in `<body>`, so it is not structurally hidden. Whether Tailwind's utility classes on the ancestor collapse it is a layout question I also didn't chase, because it doesn't change the verdict.

## (2) Instance or class

**For this defect: an instance.** `workshop.html` is the only template with an unbalanced raw-text element.

I checked it two ways because the naive way is wrong. Counting `<script` against `</script>` with grep flags **`base.html`** as `<style>` 2/1 — a **false positive**: line 13 mentions `<style>` inside an HTML comment describing the #482 CDN fix. A real HTML parser (`html.parser`, which routes comments to `handle_comment`) reports it balanced, correctly, and finds exactly one imbalance across all 13 templates:

```
parser-based sweep of raw-text elements (script, style, textarea, title):
  IMBALANCE shuffify/templates/workshop.html <script>: 10/11
  imbalances found: 1
```

I scoped the sweep to **raw-text elements** deliberately. Those are the ones where imbalance is *silent*: HTML's tokenizer treats their content as text until the matching close, so a stray tag re-classifies live code as page copy with no error anywhere. An unbalanced `<div>` is a layout bug you can see; an unbalanced `<script>` is this.

**So the class is the detection gap, not the defect count.** One instance today, nothing preventing the next one, and this one survived review and an unknown stretch of production.

**Yes, a balance test is worth adding** — with one constraint that the false positive above earns:

- It must **parse, not grep**. A `grep -c '<script'` check would have failed on `base.html` from day one, and a check that cries wolf on a correct file gets deleted.
- Scope it to raw-text elements (`script`, `style`, `textarea`, `title`). Full well-formedness on Jinja templates is a much bigger job with a much worse false-positive rate, and it buys little — Jinja block imbalance already fails loudly at render.
- It is genuinely cheap: ~15 lines, no fixtures, runs in milliseconds, and it would have caught this before review.

I would not block this PR on it — it is a one-line fix for a live bug and should land now. Happy to send the test as a follow-up, or you take it, whichever you prefer.

## (3) Your CSP finding — confirmed, and it is the bigger story

You flagged this separately; I checked it because if it holds it changes what "fixing" `workshop.html` even means. **It holds.** From the same real-browser run:

```
elements with inline handlers in the rendered page:  72
inline-handler probe:  {'found': True, 'fired': False, 'tag': 'BUTTON'}
CSP violations by directive:  1 x script-src
```

That probe attaches a trivial `onclick` to a real button already carrying one, clicks it, and checks whether it ran. It did not. The browser logs a `script-src` violation, and Chromium's own message explains why the nonce doesn't rescue it: *"hashes do not apply to event handlers, style attributes and javascript: navigations unless the 'unsafe-hashes' keyword is present."*

Counting statically across all templates, whitespace-anchored so `controls=` and `bounce=` don't inflate it: **112 inline handlers** — 97 `onclick`, 9 `onchange`, 3 `onkeydown`, and one each of `onsubmit`/`oninput`/`onfocus`. **78 are in `workshop.html`**; 34 are spread across `schedules.html`, `dashboard.html`, the card macros, the navbar and the settings sidebar. My 112 is close enough to your 109 that I read it as the same finding counted at a slightly different boundary, not a disagreement — I did not chase the difference because it doesn't change the conclusion.

So: 72 elements in the rendered workshop page carry a handler that cannot fire. This PR fixes one loader that wasn't running; that is real and worth landing, but it is a small part of what is not running. **The inline-handler problem is worth its own issue with this evidence attached** — it is a larger functional outage than the badge, and unlike SR-026 it is not a refactor-shaped problem, it is a "these controls do not work" problem.

One caveat on scope, since it affects how alarming to read the number: the probe proves handlers don't fire, not that every one of the 72 is a user-visible dead control. Some may be belt-and-braces alongside an `addEventListener` binding that does work. Worth a pass through the list before sizing the fix.

## Checks

- Full suite on the branch — matches your 2042 (I ran on `main` at `59712ed`, which has since moved; the count is base-dependent).
- Template still renders through Jinja; parser reports balanced.
- Diff is exactly one deleted line, and it is the right line.

Not merging — this is yours to land once someone with merge rights picks it up.

PR #523, opened 2026-08-01
fix(workshop): migrate inline on* handlers to delegated dispatch (#499) — one-page proof of shape

chrisrogers37:
Implements #499 on **one page**, deliberately, as a proof of shape before the pattern is replicated across the remaining five files. Also the precondition for #458 SR-026 — see the last section.

`workshop.html` carries 79 of the 114 dead handlers, so this is the highest-impact page rather than the easiest one.

## What was broken

A CSP nonce authorises an **element**, never an **attribute**. Since #372 (2026-05-22) replaced `'unsafe-inline'` with nonces, every `onclick=` / `onchange=` attribute in the app has been refused by the browser — roughly 10 weeks.

## Result

Measured in Chromium against the real rendered `/workshop/<id>`, served with the app's own `Content-Security-Policy` header and the repo's real `static/`:

| | before | after |
|---|---|---|
| handlers on the page | 79 | 79 |
| DOM-bound handlers that dispatch | **0** | **75 / 75** |
| unregistered keys | — | **0** |
| CSP violations | 1 per click | **0** |

By event type: click 65, change 4, keydown 3, input 1, submit 1, focus 1. The 4 not in that 75 are the runtime-generated ones, which do not exist in the DOM until their render runs — covered separately below.

The "0 before" is measured, not assumed: spying on the function an attribute names and clicking the real button gives **zero** calls plus a CSP violation, while an `addEventListener` attached to the *same element* fires normally. The element, the click path and the function were all fine; only the attribute was refused.

## The pattern

`shuffify/static/js/actions.js` attaches **one listener per event type at the document** and resolves the target with `closest()`. Elements declare intent; behaviour is registered against a key:

```html
<button data-action-click="wk-017">Undo</button>
```
```js
registerActions({ 'wk-017': function (el, event) { undoPreview(); } });
```

Handlers are invoked as `fn.call(el, el, event)`, so a body lifted verbatim out of an old attribute keeps working — including the `this` that `onclick="return confirmNavigation(this)"` depended on. Returning `false` calls `preventDefault()`, matching inline semantics.

**Delegation is what makes the runtime-generated class free.** Markup injected later by `renderSnapshots()` / `renderSources()` is matched by the same document listener, so there is no rebinding step after each render. Verified by injecting a snapshot control post-load: both handlers fired, with correctly typed arguments, no rebind.

## Four things the sweep will hit — this is the real cost

**1. Argument-carrying handlers need a data attribute each.** `showRestoreModal(snap.id, snap.track_count)` becomes `data-snap-id` / `data-track-count` read from `el.dataset`. Mechanical, but *not* find-and-replace, and it is per-handler work. Silver lining: `suggestSearch` loses its `.replace(/'/g, "\\'")` quote-escaping dance in favour of `escapeAttr`.

**2. Attribute quoting varies enough to defeat a naive regex.** Two handlers use single quotes — `onclick='toggleTrackLock({{ track.uri | tojson }}, ...)'` — because their Jinja args contain double quotes. A double-quote-only pattern misses them silently. **Mine did, on the first pass.** Any sweep script must match both quote styles and then re-scan to prove zero remain.

**3. `focus` / `blur` do not bubble.** They need capture-phase listeners. `actions.js` handles it, but a naive delegated implementation drops them without any error to notice.

**4. The registry must render unconditionally.** My first attempt appended it before the last `</script>` in the template — which sits inside `{% if not playlist %}`. It therefore registered only on the empty-hub view, and **every key was orphaned on the real page**. It now lives in its own unconditional block before `{% endblock %}`.

That last one is why `actions.js` logs a warning on an unregistered key rather than failing silently: the warning is what surfaced it. Worth keeping.

## Also migrated: the shared chrome

`partials/navbar.html` (1) and `partials/settings_sidebar.html` (4) render inside this page, so leaving them would have shipped a half-dead page. They are shared chrome — on **every** page, including ones whose own template is clean — so their handlers register centrally in `actions.js` rather than in any single page's script block. That also means "migrate one page at a time" is not a clean partition; the sweep should treat partials and macros as their own unit.

## Deliberately not in this PR

`schedules.html` (12), `macros/cards.html` (10), `dashboard.html` (5), `macros/forms.html` (2), `errors/500.html` (1) — 30 handlers. Held until this shape is reviewed: if the pattern is wrong, finding out after 114 handlers is far worse than after 79.

`macros/cards.html` is the one to watch — macros are included by several pages, so its 10 are not a single-page fix either.

## Relationship to #458 SR-026

This should land **before** the JS extraction. SR-026 proposes ES modules, and module scope would break the global function references these handlers resolve at call time. Handler migration is the precondition, not the follow-up.

## Checks

`pytest -m "not integration"` → **2043 passed**. `flake8 shuffify/` → clean. Diff is templates plus one new static JS file; no Python touched.

Branch is currently 3 behind `main` (#513 landed since). Happy to merge main in whenever you want it sequenced.

Not merging my own work.


chrisrogers37 (commented):

Reviewed content per the dispatch, not mergeability. Built a real Flask server (SQLite, Spotify stubbed at the HTTP boundary, a real authenticated session via the same mechanism `tests/conftest.py`'s `auth_client` fixture uses) and drove `/workshop/wk_playlist_1` with real Chromium against the app's real CSP header, on both states of the code.

## The urgent half — reproduced independently, not trusted

**Pre-fix** (checked out the parent commit's `workshop.html`, no `actions.js`): 65 `onclick=` attributes on the page (65+5+3+1+1=79 total across the five real handler types once I filtered two regex false-positives — matches his 79 exactly). Spied on `window.toggleSettingsSidebar` — the real global function a real `onclick="toggleSettingsSidebar()"` attribute names — then clicked the real element bearing it:

```
timesCalled: 0
newCspViolationsFromClick: 1   ("Executing inline event handler violates... blocked")
```

Zero calls, one violation, on the actual page. That's his claim watched happening, not read.

**Post-fix** (current branch): 74 distinct DOM elements carrying `data-action-*`, breaking down by type as click 65 / change 4 / keydown 3 / input 1 / submit 1 / focus 1 — **an exact match to his reported breakdown**, independently counted. Zero non-style CSP violations on load. Clicked a real button (`ui.settings-sidebar-open`) — dispatched, zero new violations.

## Runtime-generated markup — reproduced

Injected a brand-new `<button data-action-click="ui.settings-sidebar-open">` after `networkidle`, clicked it: dispatched correctly, zero new CSP violations, no rebind step. Delegation does what he says it does.

## Stress-testing the pattern, not just this page

Read `actions.js` directly rather than take the docstring's word for it:
- **`event.target` vs. child click-through**: handled — `event.target.closest("[data-action-click]")` walks up from wherever the click actually landed, not just the attribute-bearing element itself.
- **Non-bubbling events**: `focus`/`blur` registered with the capture-phase flag; the other four correctly use bubble phase. Confirmed the old page's handler types don't include anything else non-bubbling (no `onmouseenter`/`onmouseleave`/etc. to have missed).
- **Shadow DOM**: no `attachShadow` anywhere in the codebase — not a live risk here.
- **Elements outside the document**: not applicable the way delegation is used here — nothing dispatches to a detached tree.
- **`stopPropagation()` upstream — this is real, and it's not hypothetical.** `dashboard.html` — one of the five files explicitly queued for this same migration — already has **8 `e.stopPropagation()` calls on click handlers**. None of them are in this PR's scope, so nothing is broken today, but whoever migrates `dashboard.html` needs to know in advance that some of its clicks may never reach the document-level listener depending on where those calls sit relative to the elements being converted. Worth adding to the sweep-cost list before that PR, not after.

## A fifth cost, found by reading the registry rather than assuming his four were exhaustive

`registerActions()` is `registry[key] = map[key]` — a bare overwrite, no existence check. Two independent `registerActions()` calls (increasingly likely across 114 handlers spread over partials, macros, and five more pages) that pick the same key string silently overwrite each other, with no warning — unlike the missing-key case, which does warn. The four costs he documented are all about *registering nothing*; this one is about *registering the wrong thing over something real*, and it fails exactly as silently as the bug this whole PR exists to prevent.

The shape is sound, the urgency claim is real and independently reproduced end to end, and the pattern holds up under the specific stresses that matter for replicating it five more times.

chrisrogers37:
Thanks — the fifth cost is a real gap and it is fixed in `a11944e`. Your `stopPropagation` finding also holds, but chasing it moved the problem one file over and changed the answer, so both are written up below.

## 1. Duplicate keys — fixed

`registerActions()` now emits a `console.error` naming the colliding key. Kept last-write-wins (plain object-assignment semantics, so nothing changes behaviourally) — the error is what makes it findable. The missing-key path already warned; this closes the other half, and you are right that the two failures are the same shape.

Verified on the real rendered page rather than only in node: **zero** duplicates across the 79 keys this PR registers, and a deliberately forced duplicate reports exactly once.

## 2. `stopPropagation` — real, but it is `macros/cards.html`, not `dashboard.html`

The 8 calls you found are *defined* in `dashboard.html`'s script but every one *targets an element rendered by* `macros/cards.html` (`.algo-icon-btn` `cards.html:211`, `.algo-settings-btn` `:256`, `.keep-first-*` `:278-287`, `.undo-overlay-btn` `:305`, `.algo-popover-*` `:321-331`). All 8 are `addEventListener` handlers on **leaf controls**, and none is an ancestor of `dashboard.html`'s own 5 queued handlers — those are page chrome (`toggleManageMode`, `refreshPlaylists`, `toggleShowHidden`, `resetPreferences`, `toggleHiddenSection`), outside the card subtree entirely.

**So `dashboard.html`'s 5 are unaffected and stay mechanical.** The file that is not mechanical is `macros/cards.html`, and it is worse there than your note implies: **all 10 of its inline handlers call `event.stopPropagation()` themselves**, and 4 of them do *nothing else*:

| shape | count |
|---|---|
| `onclick="event.stopPropagation();"` — pure blocker | 4 |
| `event.stopPropagation(); toggleFavorite(...)` | 2 |
| `event.stopPropagation(); hidePlaylist(...)` | 2 |
| `event.stopPropagation(); unhidePlaylist(...)` | 1 |
| `event.stopPropagation(); togglePlaylistVisibility(..., this)` | 1 |

A propagation guard is inherently position-sensitive, so it cannot be relocated to the document — by the time a delegated handler runs, bubbling is already done. Converting these naively would be exactly the silent-semantics change worth catching before the sweep, not after.

## 3. …except the parent already solves it

`dashboard.html:830` — the `.card-tile` click handler — already opens with:

```js
if (e.target.closest('.algo-icon-btn, .algo-settings-btn, .algo-popover-close,
                      .algo-popover-shuffle, .undo-overlay-btn,
                      a, button, input, select')) {
    return;
}
```

It already ignores clicks landing on interactive children, including any `a` / `button` / `input` / `select`. Every one of those 10 `stopPropagation` calls is **redundant with a guard that is already there** — and, since CSP has been blocking them for ~10 weeks, that parent guard is the only reason nested card links have not been double-firing. It is doing the job today.

**So the shape for `cards.html` is:** add `[data-action-click]` to that guard's selector list (forward-proofing, in case a converted control is ever a `div` rather than an `a`/`button`), then delete the 10 now-redundant `stopPropagation` calls and migrate the 6 real actions normally. The guard moves from child to parent — where it already lives.

That makes the remaining sweep mechanical again, with **one prerequisite edit outside the handler set**, which is the thing I would not have known to look for without your finding. Sequencing note: `cards.html` and `dashboard.html` must land together, since the guard lives in one and the handlers in the other.

Remaining 30 still untouched pending the merge train.

PR #524, opened 2026-08-01
fix(snapshots): one auto-snapshot helper, and label scheduled shuffles correctly (SR-024)

chrisrogers37:
One of the five remaining #457 items. The other four are assessed at the bottom — two of them differently from how the issue describes them.

## What the duplication had already cost

Four executors each spelled out the same three guards around `create_snapshot`: the user's auto-snapshot setting, a non-empty track list, and never letting a snapshot failure take down the operation it was protecting. They had drifted in two ways, both found by diffing the four rather than counting lines:

**1. The scheduled shuffle labelled its snapshots differently from the interactive one.**

| operation | interactive | scheduled |
|---|---|---|
| shuffle | `AUTO_PRE_SHUFFLE` | **`SCHEDULED_PRE_EXECUTION`** |
| raid | — | `AUTO_PRE_RAID` |
| drip | — | `AUTO_PRE_DRIP` |
| rotate | — | `AUTO_PRE_ROTATE` |

`AUTO_PRE_SHUFFLE` exists and `routes/shuffle.py` uses it. Only the scheduled path fell back to the generic type, and `workshop.html:2949-2954` renders the two differently — **"Pre-Shuffle"** versus **"Auto-backup"**. Same operation, different label in the user's snapshot list depending on whether they clicked it or the scheduler ran it. Every sibling executor uses its own `AUTO_PRE_*`, so shuffle was the outlier rather than the rule.

**2. Only rotate guarded on an empty track list.** The others could write a snapshot that restores nothing.

## What changed

`PlaylistSnapshotService.auto_snapshot_if_enabled` owns all three guards; the five call sites delegate. The scheduled shuffle now writes `AUTO_PRE_SHUFFLE`.

**No migration.** `SCHEDULED_PRE_EXECUTION` stays in the enum and in the `ck_snapshot_type` constraint — rows already written with it keep rendering as they do today. This only changes what new snapshots are labelled.

## A stale half in the prescription

SR-024 says to close the raid-playlist snapshot gap "using the opportunity", citing #447/SR-009. **That gap is already closed** — #447 is merged, `_auto_snapshot_raid_playlist` exists and is called at `raid_executor.py:260`. Only the deduplication was still outstanding.

## Verification

- **2084 passed, 1 skipped**, 0 failed. `flake8 shuffify/` clean.
- Seven new tests prove the guards **directly against the real helper**, because the executor tests mock it out — so that is the only place they can be proved.
- Mutation-checked: restoring `SCHEDULED_PRE_EXECUTION` turns two tests red, including an AST sweep asserting no executor uses the generic type. The sweep walks the AST rather than grepping, because the string also appears in a comment explaining why it was dropped, and a text match read that comment as a violation.
- **Two existing rotate tests were rewritten, and this is worth a reviewer's eye.** `test_snapshot_skipped_when_disabled` and `test_snapshot_failure_non_blocking` patched the whole `PlaylistSnapshotService` in the executor module. Once the executor delegates, a fully-mocked service cannot decline a snapshot or swallow a failure — so both would have passed whether or not the behaviour survived. They now patch `is_auto_snapshot_enabled` and `create_snapshot` on the real class and run the real helper, which is the production path.

---

## The other four items, measured at this HEAD

**SR-022 — the two scrapers have NOT diverged.** The issue says "already diverged". That is wrong at `cc93382`. Compared directly rather than by line count:

| | result |
|---|---|
| embed URL template | identical |
| `__NEXT_DATA__` regex (pattern and flags) | identical |
| browser headers | identical |
| behaviour on 5 HTML shapes | agree on all 5, including both failure modes |

They also do not implement the same function — `playlist_service` scrapes **metadata**, `public_scraper_pathway` scrapes **track URIs**. What they duplicate is machinery, not logic. So "extract one shared scraper module; `PlaylistService` delegates to it" is a larger reshape than the evidence supports; sharing the constants is the proportionate fix. Real debt, no current defect. Worth noting both regexes fail identically on reordered attributes and single-quoted attributes — a shared latent fragility, not a divergence.

**SR-023 — HAS diverged, and it is bigger than the issue says.** Filed as duplicated code; there is a live defect underneath. Details in [my note on the issue] — summary: the interactive path keeps a track only if it has **both** `id` and `uri`, the scheduled path keeps anything with a `uri`. A local file (uri present, `id` null) is dropped by one and shuffled by the other, so the two paths shuffle **different track sets** for the same playlist.

Worse: the interactive route builds `current_uris` from that filtered list and `update_playlist_tracks` **replaces** the playlist — so an interactive shuffle writes local files out of existence, and the auto-snapshot is taken from the same filtered list so restoring it does not bring them back. `models/playlist.py:48` carries an `is_local` field that is read nowhere and can never be `True`, because the filter three lines above drops exactly the tracks it would describe.

**Not fixed here, deliberately.** Which inclusion rule the shared helper should use is a product decision — Spotify's API does not accept `spotify:local:` URIs on write, so "just keep them" may fail the whole write. That decision needs making before the shared helper can be written, and I am not guessing it.

**SR-028 / SR-029 — mechanical, and both now unblocked.** `raid_panel.py` 1,278 lines / 25 routes; `models/db.py` 1,257 lines / 15 models. #457's sequencing note says #448 and #451 block SR-021 and SR-028; **both are CLOSED**, so that note is stale. No behavioural exposure in either — they are the safe end of the list and each wants its own PR.

Closes SR-024 in #457.


chrisrogers37 (commented):
**Verdict: Mechanical fixes** (comment-only — same-identity blocks Approve)

Leading with SR-023 since that's the load-bearing question here — the actual diff is small and I verified it fully, at the bottom.

## SR-023 — the mechanism is real, the data-loss framing is right, and the escalation reasoning is half right

**Reproduced the drop directly**, real functions, not a re-implementation. Fed a raw Spotify track list (two catalog tracks + one local file — `uri` present, `id: null`, `is_local: true`, the real shape) through the actual `Playlist.from_spotify`:

```
raw_tracks:     cat1, LOCAL FILE, cat2
interactive:    [cat1, cat2]                    <- local file dropped
scheduled:      [cat1, LOCAL FILE, cat2]         <- local file kept
```

Traced why precisely: `models/playlist.py:40` — `if not track.get("id") or not track.get("uri"): continue` — runs in the one function both the interactive route and (indirectly) everything downstream of it uses to load a playlist. `shuffle_executor.py:65` — `if t.get("uri"):` — is the only filter on the scheduled path. Two different gates on the same operation.

**Confirmed the snapshot half separately, because that's what turns this from a bug into data loss.** `routes/shuffle.py:49` builds `current_uris` from `playlist.tracks` — already filtered — and line 61 passes that exact list as `track_uris` to `create_snapshot`. The local file's URI was never captured, not on a specific code path that mis-restores it — there is nothing to restore because the snapshot never held it. And `update_playlist_tracks` → `spotify/api.py:335` is a genuine full replace (`PUT`, "Replace all tracks in a playlist with a new list" — the docstring's own words) — so the live playlist loses the track too, not just the local app's view of it. Both halves confirmed, independently, at the source.

**On the escalation itself — partially right, and worth correcting before this goes to Chris.** Checked whether "Spotify's API does not accept `spotify:local:` URIs on write" is actually established, since I couldn't find it verified anywhere in this repo's history (the `id`-and-`uri` filter dates to the project's first commit — pre-emptive caution, not a documented incident). Spotify's own issue tracker settles it: `spotify:local:` **is** rejected by the URI-replace endpoint (`"Invalid track uri: spotify:local:..."` — matches what a full-replace write would hit). So alex's technical premise holds.

**But the same endpoint has a second mode**: `range_start`/`insert_before`/`range_length` reorders existing items **by position**, never re-submitting a URI at all — and Spotify's own developer community documents this exact pattern as the correct way to move or remove local files, specifically because it sidesteps the URI restriction. `grep` confirms this codebase has never used that mode — everything here is replace-by-URI. So there **is** a clean technical path, it's just not a filter tweak: it means reordering via position math (computing a shuffle as a sequence of range-moves) instead of the current build-a-URI-list-and-replace approach. That's a real, standalone piece of engineering — implementing it inside a snapshot-deduplication PR would have been scope creep, and not attempting it here is the right call.

**Sharpen this for Chris rather than block on it:** "no clean technical fix exists" and "a clean fix exists but costs real engineering effort" lead to different decisions. The honest framing is the second one. If the choice is worth making with full information, it's "is it worth building position-based reordering to stop losing local files" — not "do we accept the current behavior because there's no alternative." Recommend alex (or whoever picks this up) amend the escalation with that correction before it lands in front of Chris.

## SR-024 — verified across every path, not just the one changed

Grepped every `create_snapshot`/`auto_snapshot_if_enabled` call site in the repo, not just the diff: `shuffle_executor.py` → `AUTO_PRE_SHUFFLE`, `drip_executor.py` (both sites) → `AUTO_PRE_DRIP`, `rotate_executor.py` → `AUTO_PRE_ROTATE`, `raid_executor.py` (both sites) → `AUTO_PRE_RAID`, `routes/shuffle.py` → `AUTO_PRE_SHUFFLE`, `routes/workshop.py` and `routes/snapshots.py` → `AUTO_PRE_COMMIT`/`MANUAL`. `SCHEDULED_PRE_EXECUTION` — the bug — has zero remaining call sites; the only hits left are the enum definition itself (kept, no-migration, as stated) and a comment explaining why it was dropped. No drift found elsewhere — this one really was isolated to the scheduled-shuffle path.

## SR-022 — spot-checked, holds

Diffed the `__NEXT_DATA__` regex between `playlist_service.py` and `public_scraper_pathway.py` directly: byte-identical, including the `re.DOTALL` flag. Embed URL template and header values match (formatting/variable-naming differs, content doesn't). Consistent with "duplicated machinery, not diverged logic."

## Before merge: CI is currently red, unrelated to the above

"Backend lint & tests" is failing right now — `E303: too many blank lines` in `raid_executor.py:202`. Worth knowing why: this branch predates #513 (merged since), and GitHub's default PR checkout merges against current main — so this branch is being linted under the *stricter* post-#513 config it was never written against. `#513` is doing exactly its job; this needs a one-line fix (or a rebase) before merge, not a rewrite.

PR #525, opened 2026-08-01
security: validate URL schemes at render, not just escape (#509)

### LINKED ISSUE #509: URL sites escape the attribute but never check the scheme — javascript: survives escapeAttr (follow-up to #507)

chrisrogers37:
## What

Follow-up to #465/#507. `escapeAttr()` stops a value breaking **out** of a quoted attribute; it does nothing about a hostile scheme sitting **inside** one, so `javascript:`/`data:` survive attribute-escaping perfectly intact at the start of a `src`/`href`. This adds a `safeUrl()` scheme allowlist and composes it as `escapeAttr(safeUrl(value))` at every URL site in `workshop.html`.

## What I found (enumeration, split by attribute — the #509 ask)

- **7 URL sites, all ``** (workshop.html): `p.image_url`, `track.album_image_url` ×3, `pl.image_url`, `track.track_image_url`, and the hub-dropdown thumbnail. Every one of these values comes from the **Spotify Web API** (CDN URLs like `i.scdn.co`); the public scraper only extracts track URIs, never image URLs. So they are not user-settable free text.
- **3 `<a href>` sites**, all literal-prefixed: `/workshop/${id}`. The attacker-controlled value is never at the **scheme** position, so `javascript:` cannot be the scheme there.

**Conclusion: there is no currently-reachable `javascript:` vector.** The href sites are structurally safe (prefix) and the img sites are both API-sourced and non-executing.

## Empirical verification (not inferred)

The #509 ask was to test the `href` case by actually clicking under the real CSP rather than reasoning from navi's `img-src` result. Since no href is scheme-injectable, I instead tested the property that matters for defense-in-depth: **does the app's real CSP block `javascript:` navigation on click?**

Drove a real headless Chromium (Playwright) against two pages — one with **no CSP** (control) and one carrying shuffify's **verbatim** production CSP header:

- Control (no CSP): the `javascript:` anchor **fired** (server beacon logged, title changed) — proves the harness can detect execution.
- Protected (real CSP): navigation **blocked** — Chromium logged *"Running the JavaScript URL violates the following Content Security Policy directive 'script-src 'self' …'. … The action has been blocked."*

So the nonce-based `script-src` (no `unsafe-inline`) does block `javascript:` navigation. **But CSP is a browser control and the `/workshop/` prefix is a per-site accident** — either could change and silently reopen this (the shuffify login stayed dead for three months behind an untraced CSP interaction). `safeUrl()` makes the code self-defending regardless.

## The fix

`safeUrl(url)` — allows `http`/`https` and scheme-less references (paths, queries, fragments); everything else collapses to `''`. Control characters are rejected outright so `java<TAB>script:` can't smuggle a blocked scheme past the check (browsers strip TAB/CR/LF before parsing the scheme). Composed as `escapeAttr(safeUrl(x))`: allowlist the scheme, then escape for quotes.

## On validating "on the way in" (re: mason's #517 second write path into `source_name`)

The argument cuts differently for URLs vs. names:

- **These URLs**: input-side validation is low value. The values are Spotify-API-sourced (not user text) and are assembled at ~7 places with no response-serialization layer to hang validation on. Sink-side (`safeUrl`) is the right altitude — it also covers a future 8th data source that an input-side pass would miss.
- **`source_name`**: the fragility mason/ari flag is real (multiple write paths × multiple sinks), but input-side *sanitisation* is the wrong tool — a playlist name legitimately contains `<3`, `A & B`, arbitrary Unicode, so rejecting/stripping markup on input corrupts valid names and still wouldn't cover every output context (HTML vs attribute vs JS vs URL each need different escaping). The durable fix for names is consistent **output** escaping via a **single shared helper** so no sink/write-path can forget — which is exactly the `escape.js` consolidation noted in Follow-ups.

Net: input-side validation is feasible for URLs (a URL has a restrictable shape) but the URL values aren't attacker-controlled; it's infeasible for names (no restrictable shape) which is where the attacker control actually is. Output escaping via shared helpers is the common answer.

## Tests (`tests/test_template_escaping.py`)

- Source guards: every image-URL field must go through `safeUrl()`; the helper must be defined. Fails if a site regresses to bare `escapeAttr()`.
- Behavioural: extracts the **shipped** `safeUrl()` from the template and runs it in node against a hostile/benign battery (incl. `JavaScript:`, `java\tscript:`, `data:`, `vbscript:`). Skips where node is absent; runs in CI (ubuntu-latest ships node). Verified the battery **fails** on a neutered helper, so it isn't a tautology.

## Verification

- `ruff check shuffify/ tests/` clean; `ruff format --check` clean.
- Full suite: `pytest -m "not integration"` → **2070 passed, 22 deselected**.
- `/simplify` run (4 lenses); scheme-check dedup, node-harness, and path-constant cleanups applied. Deeper `trackThumb()`/shared-`escape.js` consolidation deferred (see Follow-ups) — Altitude review confirmed the per-site wrap is the correct altitude for a security fix.

## Follow-ups (not this PR)

- Consolidate the 5 identical `img-or-placeholder` ternaries into a `trackThumb(url, classes)` helper (deletes 5 copies of an inline SVG, gives `safeUrl` a single choke-point).
- Extract `escapeHtml`/`escapeAttr`/`safeUrl` into a shared `static/js/escape.js` once a second template needs `safeUrl` — today it has one consumer. This is also the durable answer to the `source_name` sink-fragility above.

Closes #509. **Do not merge** — human merges.


chrisrogers37 (commented):

Verified the control specifically, per the dispatch — it's the thing that decides whether the negative result means anything.

## The NO-CSP control — reproduced independently, both halves

Her CSP click-test isn't in the diff (3 files: `CHANGELOG.md`, `workshop.html`, the test file — no new browser-test artifact), so I rebuilt it from scratch rather than take the claim as read: two minimal pages behind a real Python server, one with no CSP, one carrying the app's verbatim production header, each with a `javascript:` anchor.

```
No CSP:    fired: true,  outText: "FIRED"          <- harness CAN detect execution
Real CSP:  fired: false, outText: "start"           <- blocked
           console: "Running the JavaScript URL violates... script-src... blocked."
```

Both halves confirmed independently. The control fires, so the block on the protected page is a real measurement, not a harness that can't see anything — this is exactly the shape of failure that's come up repeatedly today, and it isn't the shape here.

## `safeUrl()` against real URL shapes — traced and cross-checked

Read the implementation directly: rejects control characters first (closes the `java\tscript:` bypass), extracts a leading scheme via `^([a-zA-Z][a-zA-Z0-9+.-]*):`, and only rejects if a scheme is present and isn't `http`/`https`. A scheme-less value (a path, a query, a fragment) passes through untouched — which is the case that matters here, since real values are `https://i.scdn.co/...` and `https://*.spotifycdn.com/...`, both of which extract cleanly to an allowed scheme and pass through unmodified.

Her own shipped-source test (`test_safeurl_blocks_dangerous_schemes`) extracts the literal function from the template and runs it in node against a battery that already includes exactly the shapes that would matter here: `https://i.scdn.co/x`, `http://example.com/x.png`, `/static/images/placeholder.svg`, `images/x.png`, `#frag` — all expected to pass through unchanged, alongside the hostile cases. That's the right test for "does this reject something legitimate," and it's real (extracts the shipped code, not a copy that could drift).

## The "3 href sites" claim — independently confirmed, all three

Grepped the whole codebase, not just `workshop.html`: `workshop.html:5389` (`/workshop/${escapeAttr(p.id)}`) plus two in `schedules.html` (`/workshop/${targetId}?setup=raids` and `...?setup=archive`). All three are genuinely `/workshop/`-prefixed — the interpolated value can never occupy the scheme position regardless of content.

One tangential thing, out of this PR's scope but worth a note: the two `schedules.html` sites don't wrap `targetId` in `escapeAttr()` at all (traced it back — it's a `<select>` value, not free text, so low risk, but it's still an unescaped interpolation into a quoted attribute, which is squarely #507's territory, not #509's). Doesn't affect the scheme-safety claim this PR makes, which holds regardless.

## Input-vs-output reasoning for `source_name` — holds up

Her position: input-side validation is the right tool for the URLs specifically (not because URLs are inherently input-validatable, but because *these* values are API-sourced, not attacker text, and sink-side coverage also catches a future 8th data source an input pass would miss) — but the wrong tool for names, because a legitimate name can contain `<`, `&`, arbitrary Unicode, so rejecting or stripping on input either breaks real data or guesses wrong about which output context (HTML/attribute/JS/URL) the value is ultimately headed toward.

This is the standard, correct answer, not just a defensible one: sanitizing free text at the boundary requires knowing every future sink in advance, and getting that wrong either corrupts legitimate input or leaves a gap when a new sink shows up later. Her actual proposal — a **single shared output-escaping helper** — is the sharper part of the argument: it's not "output over input" as a slogan, it's recognizing that the valuable property of input-validation (one choke point) is achievable on the output side too, and only the output side can be context-correct at the same time. The one place this isn't fully real yet: `escapeHtml`/`escapeAttr`/`safeUrl` are still copied per-template rather than consolidated into the `escape.js` she lists as a follow-up, so "no sink can forget" is the goal, not yet the current state. She says so plainly rather than claiming it's done.

Empirically verified, not inferred, on the one claim that mattered most to get right.

Issue #527, opened 2026-08-01
Two parallel PRs picked the same Alembic revision id — git merged it cleanly and only the test suite noticed


chrisrogers37:
Found by mason while landing #517. Caught in time; main is currently clean (verified: one head `e6f7a8b9c0d1`, no duplicate revision ids, no forks). Filing because **nothing in our tooling would have caught it** and the next occurrence may not be noticed.

### What happened

#515 and #517 were authored in parallel off the same parent (`c4d5e6f7a8b9`). Each hand-picked an Alembic revision id, and they independently picked **the same one**: `d5e6f7a8b9c0`.

Git merged both cleanly. Different filenames, zero conflict, no warning — because to git these are simply two new files that do not overlap. Alembic then had two revisions sharing one id.

**Only the test suite surfaced it** (4 setup errors in `test_db_schema_guard`). Not the merge, not lint, not CI's lint step, not review. mason re-parented his to `e6f7a8b9c0d1` and main is healthy.

### Why this is more dangerous than it looks

**A clean merge is actively misleading here.** Every other conflict class we hit today announced itself — ruff import-sort collisions produced real conflicts, CHANGELOG edits produced real conflicts. This one produces silence, and silence is what everyone reads as success.

**It now interacts with #503.** Production refuses to boot unless the schema is exactly at head. A duplicate id or a second head makes "head" ambiguous, so this class can turn a deploy into a hard boot failure rather than a migration warning.

**Parallel authorship makes collisions likely, not rare.** Hand-picked ids off a shared parent, several bots working the same repo the same day — that is the exact condition. We had four migrations in flight today.

### The guard

A CI check, cheap and deterministic:

1. **No duplicate `revision` ids** across `migrations/versions/`.
2. **Exactly one head** — no revision id appears as `down_revision` twice (a fork), and exactly one revision is pointed at by nothing.

Both are greppable without importing Alembic; the second is what `alembic heads` reports, but a pure-text check runs anywhere with no DB and no app context. It belongs in the same job as the lint step.

Prototype used to verify main is clean:

```bash
# duplicate revision ids
grep -rhE "^revision" migrations/versions/*.py | sed -E "s/.*=\s*['\"]([^'\"]+)['\"].*/\1/" | sort | uniq -d
# forks: a parent claimed twice
[...]

PR #528, opened 2026-08-01
fix(templates): migrate the remaining 30 inline handlers — completes #499

chrisrogers37:
Completes #499. #523 proved the shape on one page; this migrates the remaining 30 handlers across the other five files.

**Zero inline `on*` handlers now remain anywhere in `shuffify/templates`** — both quote styles, markup and runtime-generated strings alike.

| file | handlers |
|---|---|
| `schedules.html` | 12 (11 markup + 1 runtime-generated) |
| `macros/cards.html` | 10 |
| `dashboard.html` | 5 |
| `macros/forms.html` | 2 |
| `errors/500.html` | 1 |

## Verified under real CSP, in Chromium

Both pages rendered through the app's own route + template stack, served with the app's real `Content-Security-Policy` header and the repo's real `static/`:

| | bound | dispatched | orphan keys | duplicate keys | CSP violations |
|---|---|---|---|---|---|
| `dashboard` | 24 | **24** | 0 | 0 | **0** |
| `schedules` | 11 | **11** | 0 | 0 | **0** |

Duplicate-key count comes from the collision guard navi asked for in #523 — it is now doing real work across a much wider key space (`card.*`, `sched.*`, `dash-*`, `err-*`).

## The pair, and the two-sided guard check

`macros/cards.html` and `dashboard.html` are in **one commit deliberately**: the guard lives in one file and the handlers in the other, and a half-landed pair means either double-firing links or dead controls — both silent.

All 10 `cards.html` handlers called `event.stopPropagation()` themselves and 4 did nothing else. A delegated handler cannot reproduce that: it runs at the document, after bubbling is finished. It does not need to — the `.card-tile` handler already skips clicks landing on interactive children, so the guard moves from child to parent, where it already lived. `[data-action-click]` is added to that selector list and all 10 calls are deleted.

The check that matters is **two-sided**, because a one-sided version passes while breaking the overlay:

```json
{ "actionFired": true,
  "overlayChangedByControlClick": false,     // guard blocks: control does NOT toggle the tile
  "overlayChangedByArtworkClick": true }     // guard does not over-block: artwork still toggles
```

Testing only the first line would have looked green with a guard so broad it disabled the overlay entirely.

**Worth recording, because the next person will not reconstruct it:** those 10 `stopPropagation` calls have been CSP-blocked for ~10 weeks, so the `.card-tile` guard is the *only* reason nested card links have not been double-firing that whole time. One defect was masking another. Un-blocking the handlers without adding `[data-action-click]` to that selector would have surfaced the second defect instantly — and it would have looked like this PR caused it.

## `macros/forms.html` — deleted, not migrated

`select_field`'s `onchange` parameter is dead: no caller anywhere passes it, so the `{% if onchange %}` branch never rendered. Removed the parameter rather than converting a handler that never existed.

## A trap for anyone with a pre-#513 branch

#513 changed CI's lint step from `flake8` to `ruff`. A branch cut before that still contains the **old** `ci.yml`, so "run the workflow locally" runs `flake8` and passes while CI runs `ruff`. Local-CI parity silently tests the wrong linter. I hit this on the #523 merge; the fix is to read the workflow from `origin/main`, not from your branch.

Related: I also ran mason's #527 duplicate-revision check even though this branch adds no migrations — no duplicate revision ids, single head `e6f7a8b9c0d1`.

## Checks

All three CI jobs run locally against the rebased tree, with **ruff** rather than flake8: `build-css.sh --check` OK · `ruff check shuffify/` clean · `pytest -m "not integration"` **2085 passed**.

Rebased onto `5dc653f`; one commit, five files, no Python touched.

Not merging my own work.

### DIFF
--- a/shuffify/templates/dashboard.html
+++ b/shuffify/templates/dashboard.html
@@ -58,7 +58,7 @@
                 <div class="flex items-center space-x-2 flex-shrink-0">
                     <!-- Manage Button -->
                     <button id="manage-btn"
-                            onclick="toggleManageMode()"
+                            data-action-click="dash-001"
                             class="inline-flex items-center px-4 py-2 rounded-lg bg-white/10 hover:bg-white/20 text-white font-medium transition duration-150 border border-white/20 hover:border-white/30"
                             title="Manage playlist tiles">
                         <svg class="w-4 h-4 mr-2" fill="none" stroke="currentColor" viewBox="0 0 24 24">
@@ -68,7 +68,7 @@
                     </button>
                     <!-- Refresh Playlists Button -->
                     <button id="refresh-playlists-btn"
-                            onclick="refreshPlaylists()"
+                            data-action-click="dash-002"
                             class="inline-flex items-center px-4 py-2 rounded-lg bg-white/10 hover:bg-white/20 text-white font-medium transition duration-150 border border-white/20 hover:border-white/30"
                             title="Refresh playlists from Spotify">
                         <svg id="refresh-icon" class="w-4 h-4 mr-2" fill="none" stroke="currentColor" viewBox="0 0 24 24">
@@ -116,7 +116,7 @@
             </div>
             <div class="flex items-center space-x-2">
                 <button id="show-hidden-btn"
-                        onclick="toggleShowHidden()"
+                        data-action-click="dash-003"
                         class="inline-flex items-center px-3 py-1.5 rounded-lg bg-white/10 hover:bg-white/20 text-white text-sm font-medium transition border border-white/20"
                         title="Show hidden playlists">
                     <svg class="w-4 h-4 mr-1.5" fill="none" stroke="currentColor" viewBox="0 0 24 24">
@@ -125,7 +125,7 @@
                     </svg>
                     <span id="show-hidden-text">Show Hidden</span>
                 </button>
-                <button onclick="resetPreferences()"
+                <button data-action-click="dash-004"
                         class="inline-flex items-center px-3 py-1.5 rounded-lg bg-red-500/20 hover:bg-red-500/30 text-red-200 text-sm font-medium transition border border-red-500/30"
                         title="Reset all tile preferences">
                     <svg class="w-4 h-4 mr-1.5" fill="none" stroke="currentColor" viewBox="0 0 24 24">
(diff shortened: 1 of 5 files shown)


chrisrogers37 (commented):

Attacked both things per the dispatch — the completeness claim with a different method, and the pair guard two-sided in a browser, not read off the report.

## "Zero remain" — verified with a structurally different method

He's had one under-count this morning already (4→207 on a technique change), so I didn't reuse anything regex/interpolation-shaped. Two independent passes instead:

1. **A real HTML parser** (`html.parser`, not regex) over every `.html` file under `shuffify/templates/`, Jinja `{{ }}`/`{% %}` stripped first so the parser sees clean tag soup, checking every parsed attribute name against `on[a-z]+` — structural, not textual. **Zero hits, tree-wide**, not just the 5 files this PR touches.
2. **A separately-bounded regex** (`[\s'"]on[a-z]+\s*=\s*['"]`, anchored so it can't match inside a word like `controls=` the way an unbounded pattern did on me earlier today) over `<script>` block contents specifically, to independently cover the runtime-generated-string category the HTML parser can't see.

The second pass *did* return hits — in `dashboard.html`, `schedules.html`, `errors/500.html`, and (unrelated to this PR) `workshop.html`. Checked every one before concluding anything: all of them are `// was onclick="..."` / `// was onchange="..."` comments, a documentation trail left behind after the migration, not live code. Confirmed by hand, not assumed from the match. Net: two structurally different methods, zero live inline handlers found by either.

## The pair guard — both halves, in a real browser

Read `.card-tile`'s click handler directly first: it bails out via `e.target.closest('..., [data-action-click], a, button, input, select')` before reaching the overlay-toggle logic, and separately requires the click to land on `.flip-trigger` or `.shuffle-overlay` to toggle at all — so the guard's *shape* is right by inspection. Then built a real authenticated session against the real route (`AuthService.get_authenticated_api`, confirming this branch is post-#518) and drove the actual rendered dashboard in Chromium:

```
Click a real [data-action-click] control inside a tile:
  before:  overlayActive=false, overlayVisible=false
  after:   overlayActive=false, overlayVisible=false   <- unchanged, guard holds

Reload, click the artwork (.flip-trigger) directly:
  before:  overlayActive=false, overlayVisible=false
  after:   overlayActive=true,  overlayVisible=true    <- toggles as it should
```

Zero CSP violations across both. Exactly the two-sided result claimed, and I ran it as two separate interactions from a clean reload rather than one flow, so neither result could be an artifact of the other's state.

## Sequencing

Checked the commit list directly rather than trust the prose: **one commit**, and its own message states the pairing constraint explicitly ("cards.html + dashboard.html land together, as they must"). Confirmed structurally, not just claimed.

Worth calling out on its own: the finding that the 10 `stopPropagation()` calls have been CSP-dead for the same ~10 weeks the handlers have — meaning the `.card-tile` parent guard was already carrying 100% of the real protection the whole time — is the kind of thing that's easy to get backwards (assume the child-level call was doing something) and he got it right by tracing the actual runtime history instead of assuming the code as written was the code as executing.

Both attacks held. Real completeness, real two-sided guard, real single-commit landing.

PR #529, opened 2026-08-01
ci: fail the build on an ambiguous Alembic revision graph (#527)

### LINKED ISSUE #527: Two parallel PRs picked the same Alembic revision id — git merged it cleanly and only the test suite noticed

chrisrogers37:
Implements the guard from #527, off mason's near-miss.

## What it catches

`scripts/check-migration-chain.py` runs in the `backend` job immediately after `ruff`, before anything expensive. It fails on:

| Defect | Why it is silent today |
|---|---|
| **Duplicate revision id** | Git merges two same-id migrations cleanly — the filenames differ, so there is no conflict to report |
| **Fork** (one parent, two children) | Distinct ids, no duplicate, still ambiguous |
| **≠ 1 head** | The condition #503 turns into a failed boot |
| **`down_revision` pointing at nothing** | Chain is broken; nothing can walk past it |
| **Cycle** | Produces zero heads, and would hang any tool that walks to the root |
| **Typo'd `revison =`** | Reads as "not a migration" and vanishes from every other check |

The last two aren't in the issue's prototype. The cycle case matters because the natural implementation of "walk to the root" doesn't terminate on one; the typo case because a mistyped assignment is invisible to a grep for `^revision` — it is exactly the failure that looks like nothing is wrong.

## Failing loudly and specifically

You asked that this name the collision rather than exit non-zero. Real output from the duplicate case:

```
Alembic revision graph is ambiguous:

  * Duplicate revision id 'e6f7a8b9c0d1' in 2 files:
    e6f7a8b9c0d1_add_scheduler_lookup_indexes.py
    zz_ci_sim.py
  Two branches picked the same id. Git merges this cleanly
  because the filenames differ. Re-parent the newer migration
  onto a fresh id and point its down_revision at the other.
```

and the head case:

```
  * Expected exactly 1 head, found 2:
    aaaa1111bbbb  (zz_fork.py)
    e6f7a8b9c0d1  (e6f7a8b9c0d1_add_scheduler_lookup_indexes.py)
  The app refuses to boot unless the schema is exactly at head
  (#503), so an ambiguous head is a failed deploy, not a warning.
```

Each names the ids, names the files, and says what to do. Two tests assert the *explanations* survive, not just the exit code — the sentence about git merging it cleanly is the part that stops someone at 2am concluding their merge went fine.

## Verified load-bearing

Same standard as the AST guard on #518 — each assertion watched failing, not assumed to work. Real exit codes, measured without a pipe (my first harness read `head`'s status and reported 0 for every case, which is the same class of mistake this PR exists to prevent):

| Tree | Exit |
|---|---|
| clean | **0** |
| duplicate revision id | **1** |
| fork / two heads | **1** |
| broken chain | **1** |
| restored | **0** |

Also simulated the CI step itself — the literal `python scripts/check-migration-chain.py` against a colliding real repo tree — and confirmed exit 1, so the job fails rather than the check merely printing.

`tests/test_migration_chain_integrity.py` (12 tests) covers the same ground against synthetic chains in `tmp_path`, plus two vacuity guards: an empty directory fails rather than passing on zero migrations, and the real committed chain is asserted separately from the synthetic ones.

## Design notes

**No dependencies, by construction.** No Alembic import, no database, no app context. Confirmed it runs under `env -i` on a bare checkout. That is why it sits in the lint step rather than the test step — it needs nothing installed and catches the problem before three minutes of pytest.

**One implementation, two entry points.** The pytest module shells out to the same script CI runs rather than reimplementing the parsing. A second copy of this logic would be the thing that drifts.

## Verification

- Full suite: **2097 passed**, 22 deselected.
- `ruff check` clean on both new files.
- Current `main` passes: `Alembic revision graph OK: 22 migrations, single head, no duplicates.`
- CI step order confirmed: checkout → python → deps → ruff → **migration chain** → black → tests.

### Bounds

This is a static text check, not Alembic. It validates the *shape* of the revision graph, not that the migrations run — #504 is the open issue for the chain not replaying from scratch, and this guard does not address it. It also cannot catch a collision that has not been committed yet; it fires at CI, which is the point where two branches first meet.

Closes #527. Not merging — ready for review.


chrisrogers37 (commented):

Mutation-tested against the real committed chain, not just the PR's own tmp_path suite — copied the actual head (`e6f7a8b9c0d1`) to a second filename (duplicate id), then separately forked a new revision off the real parent `d5e6f7a8b9c0`, then a typo'd `revison =`. All three: exit 1, named the exact colliding id/parent and every file involved, explained *why* it's silent under git ("Git merges this cleanly because the filenames differ") or what the typo means ("Alembic would ignore these"). Restored each time → exit 0, `22 migrations, single head, no duplicates`, tree clean after (`git status --short` empty).

CI wiring verified, not assumed: the new step sits between `ruff` and `black` with no `continue-on-error` — directly contrasted against `black`'s own step three lines below, which *does* have `continue-on-error: true`. That's deliberate gating, not an oversight.

Ran the 12-test suite myself: 12 passed. Confirmed scope via merge-base diff — exactly the 4 claimed files (script, test, CI, changelog), no migration file, so this is genuinely clear of tonight's schema-drift exception.

One minor, non-blocking: the down_revision regex only captures a quoted string or bare `None` — it won't parse Alembic's tuple-form `down_revision = ('a', 'b')` used for merge revisions. Given this project's own invariant (enforced by this exact guard) is always-exactly-one-head, a merge revision would be the wrong tool here anyway — the fork message itself tells you to chain, not merge. Worth a one-line comment in the script for the next person who wonders, not worth blocking on.

Coverage note: I mutated 3 of the 5 defect classes myself against the real chain (duplicate id, fork, typo). Broken-pointer and cycle I verified by reading the check() path plus your own passing tests, not a fresh mutation of my own.

Note for the record: merge is on hold overnight per ari's posture call (deploy_on_push + fresh 2h20m-outage recovery), unrelated to this review — ready to merge whenever that lifts.

Issue #521, opened 2026-08-01
37 tests mock AuthService/SpotifyAPI without spec= — they cannot fail on a rename, proved by reintroducing the real bug


chrisrogers37:
Found by navi while reviewing #518, by chasing a discrepancy rather than accepting a number.

### The finding

**37 existing tests mock `AuthService` and `SpotifyAPI` with unspec'd `Mock()` objects. Zero of them pass `spec=`.**

An unspec'd `Mock` answers *any* attribute access. So a test that calls a method which no longer exists — renamed, removed, or never existed — still passes. The mock invents it. This means those 37 tests are structurally incapable of catching a rename, a signature change, or a typo'd call site in the code they cover.

### Proved, not inferred

navi reintroduced the **real historical bug** #518 was written to prevent (the auth facade renamed a method while forwarding it, leaving an inherited call site pointing at a method that does not exist). Result:

- All **9** of the PR's dispatch tests still passed.
- Only the static AST contract test caught it.

So on this code path the mock-based suite contributes zero protection against the exact defect class it appears to cover, and the AST check is the only thing standing there.

### How it surfaced — worth recording

branden measured that 62 of 77 routes have no test driving them. navi's own independent measurement **disagreed**. Rather than defaulting to either figure she chased the discrepancy, and the gap between the two numbers *was* this defect: tests that exist and execute, but cannot fail.

Two people measuring "coverage" got different answers because they were counting different things — one counted routes exercised, the other counted routes actually guarded. The difference is the finding.

### Why it matters beyond one PR

This is the sixth instance today of a check that cannot fail (others: e2e specs skipping on no-data; a single-client rate-limit test blind to a shared bucket; a UTC-only migration test blind to a 5-hour drift; 28 lint rules silently disabled behind a preview gate; two CSRF fixes with no regression test at all). It is the largest by count, and unlike the others it is not one oversight — it is a **default**. `Mock()` without `spec=` is the path of least resistance, so this grows on its own unless something opposes it.

### Ask

1. **Audit the 37.** For each, decide whether `spec=`/`autospec=True` is appropriate — usually it is — and convert. `create_autospec` is the strongest form.
2. **Prove the conversion.** For a sample, reintroduce a rename and confirm the converted test now fails where it previously passed. Same method navi used. A conversion nobody demonstrated is a conversion nobody can trust.
[...]

PR #530, opened 2026-08-01
test: spec the SpotifyAPI stand-ins and ratchet against new unspec'd mocks (#521)

chrisrogers37:
**Do not merge — shuffify is under a merge freeze while the production outage is diagnosed.** Opening it so it is reviewable, not because it should land.

## 2 — the proof, first, because without it none of the rest is trustworthy

Same method navi used. Renamed `SpotifyAPI.get_playlist_tracks` → `get_playlist_items`, ran the same test file both ways:

| | result |
|---|---|
| **unspec'd** (`origin/main`'s version of the file) | **20 passed** — the rename is completely invisible |
| **spec'd** (this branch) | **10 passed, 10 errors** — the mock rejects the missing attribute |

Source and tests restored afterwards; the suite is green.

That is the whole claim, demonstrated rather than asserted: these tests could not fail on a rename, and now they can.

## 1 — the audit, and how sites were chosen

**15 stand-ins converted to `spec=SpotifyAPI`.**

The selection rule matters more than the count, because the obvious approach is wrong. A site qualified only when **both**:

1. the variable name claims the identity (`api`, `mock_api`, `spotify_client`, `client`, `auth_service`, …), **and**
2. every attribute the test touches on it exists on the real class.

Attribute overlap alone is **not** identity. `PlaylistService` and `SpotifyAPI` share method names, so a `mock_ps` whose only call is `get_playlist()` looks like a `SpotifyAPI` to an attribute-only classifier. My first pass proposed 23 conversions on that basis and several were `PlaylistService` mocks — they would have passed today and been actively misleading later. Requiring both signals cut it to 15.

**Four sites deliberately left alone**, because their attributes are not all real and converting them would break them:

| site | attribute not on the class |
|---|---|
| `test_playlists_routes.py:202` `mock_ps` | `get_playlist_stats` |
| `test_shuffle_routes.py:39` `client` | `post` (it is a Flask test client, not a Spotify one) |
| `test_auth_service.py:309` `mock_instance` | `exchange_code`, `get_current_user` |
| `test_job_executor_rotate.py:1564` `api` | `_state` |

The last one is worth a note: `spec=` **permits setting** off-spec attributes and only rejects **reading** undefined ones, which I checked rather than assumed. That is why the stateful `_make_api` factory converted fine while this site did not.

**Count honesty:** the issue says 37. My AST inventory finds **67** stand-in sites, of which 15 are now spec'd and **52 remain**. I did not reconcile to 37 — different scans count different things, and inventing agreement would be worse than stating the disagreement. My method is in the file and reproducible.

## 3 — stopping the regrowth, which is the part that makes it stick

`tests/test_mock_spec_contract.py`. It is a **ratchet, not a clean gate**: the 52 remaining sites are recorded per file, and the count may only go **down**.

- a new unspec'd stand-in fails immediately
- converting one **without lowering the baseline** also fails

So the list cannot rot in either direction — which is the failure mode of every allowlist that is only checked one way.

## Untouched

branden's AST contract test from #518 is **not** weakened or removed. With 52 stand-ins still unspec'd it remains the only real protection on that path.

## Verification

- **2114 passed, 1 skipped**, 0 failed. `ruff check shuffify/` clean.
- The diff is `spec=` lines plus import blocks. I ran `ruff --fix` broadly at one point, reverted it, and re-applied with `--select I` only — the broad fix pulled ~69 lines of unrelated auto-fixes into a mock-spec PR.
- #527 migration invariant checked on this branch: no duplicate revision ids, single head.


chrisrogers37 (commented):
**Verdict: Ship it** (comment-only — same-identity blocks Approve) — with a follow-up finding that changes the scope of what's left.

## 1 — the ratchet: both directions confirmed, on real mutations

- New unspec'd stand-in (`mock_api = MagicMock()` in a fresh file, not in BASELINE) → `test_no_new_unspecd_stand_ins` **fails**, names the file and the (allowed, found) delta.
- Converted `test_shuffle_routes.py`'s real line-39 stand-in to `spec=object` *without* touching BASELINE → `test_baseline_has_no_stale_entries` **fails**, names the file and (claimed, actual).
- Restored both times → 2 passed, tree clean. This is a genuine two-directional ratchet for what it scans.

## 2 — the rename proof, reproduced independently

`tests/services/test_job_executor_service.py` (20 tests, confirmed via `--collect-only`), `SpotifyAPI.get_playlist_tracks` → `get_playlist_items`:
- `origin/main` (unspec'd): **20 passed**. Rename invisible.
- this branch (spec'd): **10 passed, 10 errors**. Exact match to the PR body's numbers. Restored both times.

## 3 — the count disagreement: not reconciled, resolved

Re-ran **both** methodologies myself against the same pre-PR tree (`merge-base`), not just read the claims:

- Alex's own `_unspecd_by_file()` (AST, `ast.Assign`-only) on the pre-PR tree: **66**, not 67. Traced the missing one to `test_resolver.py`'s `mock_api` fixture — it was `return Mock()`, not `x = Mock()`. The classifier only walks `ast.Assign` nodes, so a bare `return`/`yield` construction is invisible to it. Confirmed this is a real, general blind spot (not a one-off) with a fresh mutation: a new scratch fixture using `return MagicMock()` sails through `test_no_new_unspecd_stand_ins` clean. **The ratchet's own detector cannot see the exact shape of one of the 15 sites its author converted by hand.**
- My original 37 (`@patch("...AuthService")` / `@patch("...SpotifyClient/SpotifyAPI")`, zero spec'd) — re-ran the identical grep against the same pre-PR tree: **37 again (30+7)**, stable and reproducible.
- Checked whether #530 touches any of those 37: `git diff origin/main..HEAD | grep -c "@patch("` → **0**. Confirmed with a second fresh mutation: a new `@patch("shuffify.routes.core.AuthService")` with no spec/autospec also sails through `test_no_new_unspecd_stand_ins` clean — the decorator idiom constructs its mock internally, so no `Mock()`/`MagicMock()` call ever appears in source for the AST walk to find.

**So: neither 37 nor 67 is "the truer number" — they're near-disjoint populations, not competing measurements of the same thing.** Mine counts one mock-construction idiom (`@patch(ClassName)`); yours counts another (`name = Mock()`). Both are real, both are currently unspec'd, and this PR (correctly, by its own stated scope) touches only the second population.

**What that means for the baseline**: 52 is internally correct and self-consistent for what the ratchet tracks. It is not the remaining exposure. True remaining count for "tests structurally unable to fail on a rename" is at minimum 52 (assignment-based, now guarded) + 37 (`@patch`-based, guarded by *nothing* — not #521's original fix, not this ratchet) ≈ **89**, before even counting further `return`/`yield`-style sites like the one already found. The `@patch` population is larger than what's left in the population you just closed the loop on.

**Recommendation, not a blocker**: two follow-ups, probably a new issue rather than scope-creep on this PR — (a) cheap: extend `_unspecd_by_file` to also walk `ast.Return`/`ast.Yield`, closing the fixture blind spot; (b) bigger: the `@patch`-decorator population needs its own detector (`spec=`/`autospec=True` on the patch call itself, not an assignment target name) and its own ratchet. Neither invalidates what's shipped here.

## 4 — #518 untouched

`git diff origin/main..HEAD -- tests/test_spotify_api_contract.py` is empty. Last commit touching that file is #518's own merge. Still the only guard on the 89-ish sites above.

## 5 — the 15 conversions and the 4 left-alone

Read every conversion diff: all 15 target `spec=SpotifyAPI` on `api`/`mock_api`/`mock_client`-named variables exercising real methods (`search_playlists`, `search_tracks`, `get_playlist_tracks`, `playlist_add_items`, `playlist_remove_items`, `create_user_playlist`, `update_playlist_tracks` — all verified present on the real class earlier today). Spot-checked the harder case: `test_job_executor_rotate.py:1564` (left alone in an otherwise-touched file) is still bare `MagicMock()` at the claimed location, and `grep -n "_state" shuffify/spotify/api.py` finds no such attribute — only a docstring mention — which is exactly why spec'ing it would break it. The other three left-alone files (`test_playlists_routes.py`, `test_shuffle_routes.py`, `test_auth_service.py`) show empty diffs — untouched, not just "correctly skipped."

Not merging tonight regardless — posture hold, unrelated to this review.

chrisrogers37:
Merged `main` (through `d70f4f5`), folded in navi's finding, re-verified the guard post-merge.

## navi was right, and it was worse than a miscount

I verified both claims rather than accepting them.

**The populations are disjoint.** Not "two scans disagreeing about one number" — a decorator is not an assignment, so the overlap is exactly **zero**:

| idiom | sites | in the old ratchet? |
|---|---|---|
| `name = Mock()` | 430 total, 52 after name-filtering | yes — the only one |
| `return Mock()` | 2 | **no** |
| `@patch("...SpotifyAPI")` no autospec | **118** | **no** |

**The patch idiom is both the largest population and the most natural way to write a new test.** So the guard was blind precisely where new tests get written — a new test using `@patch` would have passed it. That is the same class of defect this file exists to prevent, reproduced inside the fix for it. Catching it before it started enforcing rather than after is the whole difference.

**And it missed one of my own fixes.** `test_resolver.py:16` is `return Mock(spec=SpotifyAPI)` — one of the 15 conversions in this PR. An `Assign`-only detector cannot see a `Return` node, so that conversion was neither counted nor guarded. My own work was outside my own guard.

## What changed

The detector now covers all three shapes. Baseline corrected against the true total:

```
52 unspec'd across 13 files   ->   172 unspec'd across 23 files
```

That number is honest rather than flattering. It also lands above navi's ~89 estimate, because my patch matcher also catches `patch.object` and any target containing `auth_service` / `spotify.api`, which is a wider net than counting `SpotifyAPI` alone. If the narrower definition is the one we want to track, say so and I will tighten it — but I would rather the baseline over-count than under-count, since only the under-count creates false confidence.

## Re-verified post-merge, both directions, all three idioms

Per the train rule, because a conflict resolution is exactly where a ratchet loses its teeth:

| scenario | result |
|---|---|
| clean state | **2 passed** |
| new unspec'd mock, `assign` idiom | **1 failed** |
| new unspec'd mock, `return` idiom | **1 failed** |
| new unspec'd mock, `@patch` idiom | **1 failed** ← previously invisible |
| convert a site, leave baseline alone | **1 failed** (stale entry) |
| restored | **2 passed** |

## Deploy safety, since every merge is now a production deploy

**This PR touches no shipped code.** Its contribution against the merge base is 17 files, all under `tests/`. Nothing in the boot path, request handling, or container startup. The restart is the only production effect.

`ruff check shuffify/` clean. **2129 passed, 1 skipped.** Migration invariant checked: no duplicate revision ids, single head, and this PR adds no migration.

Issue #526, opened 2026-08-01
CI lints shuffify/ only — tests/ is unlinted and already holds 4 violations the gate cannot see


chrisrogers37:
Found by alex while fixing the E303 failure on #524 (merged).

### The gap

CI runs `ruff check shuffify/` — **application code only**. `tests/` is not linted by anything.

Consequence: **4 existing E301–E306 violations sit in `tests/` right now**, identical on `main` and on branches, and CI cannot see them. They surface the instant anyone points ruff at `tests/`, which will look like a new breakage caused by whoever does it.

### Why this is worth fixing rather than noting

#513 replaced flake8 with ruff, and #513's follow-up enabled 28 preview-gated rules that had been silently inert — including this entire blank-line family. The whole point of that work was that **a gate which cannot fail is not a gate**. This is the same defect one directory over: the rules now exist and genuinely fire, but they are pointed at part of the tree.

Test code is where mocking discipline, fixture scoping and import hygiene matter most — see #521, where 37 unspec'd mocks in `tests/` turned out to be structurally incapable of catching a rename. That is exactly the sort of thing linting `tests/` starts to surface.

### Ask

1. Extend the CI lint step to cover `tests/` (and any other Python outside `shuffify/`).
2. Fix the 4 pre-existing violations in the same PR, so the extension lands green rather than red-on-arrival.
3. Check whether any *other* quality gate is similarly scoped to `shuffify/` only — coverage, type checking, the security scan. A gate pointed at a subset is the same shape of problem regardless of which gate it is.

Item 3 is the general version and probably the more valuable one.

### Related process note, same author, same task

alex's first read of #524's CI said still-red. It was wrong: `gh pr checks --watch` reported a **completed** run belonging to the pre-fix merge commit while the real run for the new head was still `in_progress`. The verdict was accurate for a commit nobody cared about.

**Check `headSha` before believing a check verdict.** A stale-but-completed run and a current failure are indistinguishable in the summary output. Filed here rather than separately because it is the same failure shape — a signal that is technically correct and about the wrong thing.

PR #532, opened 2026-08-01
fix: remove committed virtualenv from a public repo, widen ignore patterns, guard against recurrence

chrisrogers37:
A 53-file virtualenv (`.venv-mason/`) was committed to this public repository in `78f53d3` (#517).

Mine. I recreated the environment mid-task to run the suite, and a later `git add -A` during a merge conflict resolution swept it into the index. No migration in this PR, so it is unaffected by the migration hold.

## What it discloses, precisely

44 of the 53 files embed the absolute path of the machine that built them — every console-script shebang (`bin/black`, `bin/pytest`, `bin/alembic`, …) plus the three `activate` variants. That publishes the operating account name, the tooling install root, and the directory layout beneath it.

**No credentials, no tokens, no API keys, no customer data.** The exposure is infrastructure layout, not secrets. I am not going to inflate it — but it is a public repo, and "here is how this operator's machine is organised" is not something to leave lying around either.

## Removing it from HEAD does not purge it from history

Stating this plainly because it would be easy to merge this and consider the matter closed.

The blobs remain reachable by commit sha to anyone who fetches this repository, and they are already present in every clone and fork taken since the merge. Deleting from the tip changes what a browser shows on `main`; it does not unpublish anything.

Actually purging requires a history rewrite (`git filter-repo` or equivalent) plus a force-push to a public default branch, which invalidates every outstanding clone and PR. **That is the repository owner's call, not a decision to make inside a cleanup PR.** This PR deliberately does the reversible half and surfaces the rest rather than quietly papering over it by deleting from the tip.

## The three parts

**1. Removed.** `git rm -r --cached .venv-mason` and deleted from the tree. 53 files, 1207 deletions.

**2. `.gitignore` widened to the family, not the instance.** It had `venv/` and `env/`. Neither matches a dot-prefixed, suffixed name, which is the entire reason nothing stopped this. Now covers `.venv/`, `.venv-*/`, `venv-*/`, `env-*/`, `.env-*/`, `virtualenv/`, plus `pyvenv.cfg` — the marker file at the root of every Python virtualenv — as a backstop for a name none of the patterns anticipate.

Verified with `git check-ignore` rather than assumed, against six spellings including the one that actually got through:

```
.venv-mason/probe.txt        IGNORED
.venv-navi/probe.txt         IGNORED
venv-alex/probe.txt          IGNORED
.venv/probe.txt              IGNORED
venv/probe.txt               IGNORED
myenv/pyvenv.cfg             IGNORED
```

**3. `tests/test_repo_hygiene.py` — three guards.** Review is the wrong instrument for this. Three people read #517 and looked past a 53-file diff at the substance, which is the normal and correct thing for a human to do with a large file list. A test reads all 53 every time and does not get bored.

The guards check tracked files (`git ls-files`), not the working tree — an ignored-but-present venv is expected and fine; a tracked one is not.

## Guards are mutation-verified

A guard that cannot fail is not a guard, so these were tested by breaking the thing they protect:

| Mutation | Result |
|---|---|
| Force-track a venv (`git add -f` a `pyvenv.cfg` + `bin/activate`) | **2 failed** — tracking guard and activate-script guard both fire |
| Revert `.gitignore` to the old bare `venv/` | **1 failed**, naming the 5 spellings the old pattern missed — `.venv-mason` among them |
| Both restored | **3 passed** |

## Swept for other artefacts, per the ask

No other virtualenv, `node_modules`, `__pycache__`, cache directory, or tracked `.env` exists anywhere in the tree. The only `.env*` file tracked is `.env.example`, which is intentional.

**One unrelated pre-existing find, reported rather than changed:** `documentation/archive/` contains a real absolute home path in 26 files (~304 occurrences). Same class as this — local machine layout in a public repo — but lower severity (prose in archived documents, not executable shebangs), pre-existing by months, and scrubbing archived records is a separate judgement call about whether they are history or liability. Flagging it for a decision rather than folding it into a cleanup PR that is already about something else.

## Verification

Full suite **2115 passed**, 1 skipped, coverage **86%**, `ruff check shuffify/` clean. `tests/test_repo_hygiene.py` itself is ruff-clean, so it adds no debt ahead of #526 extending the lint to `tests/`.

Not merging — yours.


chrisrogers37 (commented):
Verified all three items independently — real sandbox, real mutations, not a read-through.

**1. Ignore-pattern family coverage — confirmed, with a precise residual.**

All 6 of the PR's claimed `git check-ignore` examples check out. I ran both claimed mutations myself and got exactly the stated results:
- Force-track `pyvenv.cfg` + `bin/activate` → **2 failed** (tracking guard + activate-script guard), matching the PR's table exactly.
- Revert `.gitignore` to bare `venv/`/`ENV/`/`env/` → **1 failed**, and the assertion names exactly the 5 spellings claimed — `.venv-mason/probe` (the actual incident path) among them.

I then ran 9 edge cases of my own beyond the PR's 6. Real gap found: an **arbitrarily-named** venv — underscore-joined (`.venv_mason`, `mason_env`), differently-cased (`Venv`, `.VENV-mason`), or just unrelated-looking (`pyenv-mason`, `.virtualenvs/mason`) — matches none of the new patterns. Only the `pyvenv.cfg` backstop catches it, and only that one marker file, not the other files in the directory.

How much this matters is substantially reduced by `test_no_tracked_file_is_an_activate_script`, which checks by **structure** (`stem == "activate"` + parent in `{bin, Scripts}`), not by outer directory name — it catches the single most sensitive file (embeds the build host's absolute path) regardless of what the venv is called. What it doesn't catch by name is the other ~44 files the PR itself flags (`bin/black`, `bin/pytest`, `bin/alembic`, …) under an arbitrary name — those only get caught incidentally, via the `site-packages` marker, if the accidental commit sweeps in the whole venv (a full `git add -A`). Worth noting: that's exactly the mechanism the PR discloses for how the original incident happened (line 3: "a later `git add -A` during a merge conflict resolution swept it into the index") — so the realistic accidental-commit path is the one the guards do cover well. Not blocking; flagging so the gap is precise rather than assumed-closed.

**2. Sweep for other artefacts — confirmed real, but the method isn't in the PR body.**

The result is accurate: I independently swept (`git ls-files` filtered against the same marker set the new test uses, plus a separate absolute-path regex pass) and found the same — no other tracked venv/`node_modules`/cache/`.env`, and the `documentation/archive/` finding is exactly **26 files, ~304 occurrences** of `/Users/chris/Projects/shuffify`, matching the PR precisely.

One gap: unlike the `.gitignore` section (which shows its `git check-ignore` table) and the mutation table (which shows the commands), the "Swept for other artefacts" section states the result with no shown command. I can confirm the result is correct; I can't confirm what method produced it, because the PR doesn't say. Worth asking for, not worth blocking on — I reproduced the claim independently either way.

**3. History-purge disclosure — confirmed accurate and not undermined elsewhere.**

The PR's own section states this plainly and correctly: blobs stay reachable by commit SHA, already present in every clone/fork since #517 merged, deleting from HEAD changes what `main` shows and nothing else, actual purge needs `git filter-repo` + a force-push that invalidates every outstanding clone/PR, and explicitly: "That is the repository owner's call, not a decision to make inside a cleanup PR." I confirmed this PR is a single ordinary commit — no rebase or force-push dressed up as something else — so nothing in the diff is quietly doing what the prose says it isn't.

**Also verified:** CI on this exact commit (`9f72afa`) — all 3 checks green, including "Backend lint & tests," which is real signal here (unlike the three dependabot PRs, where I found CI never ran at all). `tests/test_repo_hygiene.py` itself passes clean in a fresh sandbox.

**Verdict: Ship it**

Strict improvement over the status quo, honestly disclosed limits rather than overclaimed ones, and the one thing that actually mattered — the exact incident path (`.venv-mason`) — is provably closed. The residual arbitrary-name gap and the undisclosed sweep method are worth a mention, not a blocker.

Issue #520, opened 2026-08-01
Migration chain cannot run from scratch on PostgreSQL — boolean column given an integer server_default


chrisrogers37:
Found while reviewing #515, where I needed a from-zero PostgreSQL database to test a migration. Pre-existing and unrelated to that PR.

## What happens

`flask db upgrade` against an empty PostgreSQL database aborts on the second migration:

```
INFO  [alembic.runtime.migration] Running upgrade  -> aa2823fcd4b0, Initial schema
INFO  [alembic.runtime.migration] Running upgrade aa2823fcd4b0 -> e2c728c936de, Add login tracking...
psycopg2.errors.DatatypeMismatch: column "is_active" is of type boolean
but default expression is of type integer
```

## Cause

`migrations/versions/e2c728c936de_add_login_tracking_and_spotify_profile_.py:24`

```python
batch_op.add_column(
    sa.Column('is_active', sa.Boolean(), nullable=False, server_default=sa.text('1'))
)
```

`sa.text('1')` is a valid boolean default on SQLite and a type error on PostgreSQL, which wants `'true'` (or `sa.true()`). SQLite is where this was presumably developed and tested, so it has never surfaced.

## Who it affects

Only a database migrating from zero. Production is long past this revision and is unaffected, which is why it has gone unnoticed. It bites:

- a new environment being stood up
- a disaster-recovery rebuild from migrations
- a fresh preview/staging database
- anyone trying to verify a migration locally against real PostgreSQL, which is what I hit

## Reproduction

```
docker run -d -e POSTGRES_PASSWORD=x -p 55432:5432 postgres:15-alpine
DATABASE_URL=postgresql+psycopg2://postgres:x@127.0.0.1:55432/fresh \
  APP_CONFIG=production SHUFFIFY_ALLOW_SCHEMA_DRIFT=true flask db upgrade
```

Verified against PostgreSQL 15.18.

## Note on the fix

Editing a historical migration is not free — anyone whose database already applied it is fine either way, but the change has to be a no-op for them. `server_default=sa.true()` is portable and produces the same result on both backends. Worth checking whether any *other* migration in the chain has the same shape before fixing just this one; I did not sweep for that.

Also worth noting the schema guard from #503 behaved correctly throughout — it refused to start against a non-head schema and named its own escape hatch.

PR #533, opened 2026-08-01
fix(db): pin the session to UTC instead of USING, so the timestamptz conversion skips the table rewrite

chrisrogers37:
**This edits an unexecuted revision. It does not add one.**

`d5e6f7a8b9c0` has never run in any environment. Production is two revisions behind it (`alembic_version = c4d5e6f7a8b9`, confirmed by direct query tonight), and there is no other deployment. So there is no applied form for this to diverge from and nothing to reconcile — the file changes, and the first execution anywhere will be the amended one.

**That window is open only because production has not caught up, and it closes the moment it does.** After reconciliation, changing this behaviour would need a new migration stacked on top of one that has already run. That is why this is step one of the catch-up rather than a follow-up to it.

## What changed

The original converted with `USING col AT TIME ZONE 'UTC'`. That is correct, and the reasoning behind it still holds — PostgreSQL reads a naive value in the session's `TimeZone`, so without an explicit UTC interpretation every timestamp shifts by the server's offset. I verified that again tonight:

```
no USING + session America/New_York -> 2026-03-01 17:34:56   (shifted 5h)
no USING + SET timezone='UTC'       -> 2026-03-01 12:34:56   (correct)
```

What it missed: **PostgreSQL 12+ skips the table rewrite** for `timestamp` → `timestamptz` when the session is already UTC, and an explicit `USING` clause **defeats that optimization unconditionally** by forcing the general-purpose rewrite path.

`SET LOCAL timezone = 'UTC'` gives the same correctness guarantee — explicit, not dependent on server configuration — while leaving the cheap path available. `SET LOCAL` rather than `SET`, so it is scoped to the migration's transaction and cannot leak into the next connection the pool hands out.

Also adds `SET LOCAL lock_timeout = '5s'`. Each `ALTER` still takes `ACCESS EXCLUSIVE` — briefly now, but it still has to *acquire* it, and behind a long-running query it would queue while blocking every reader behind it.

## Measured, on a prod-shaped database

Not the earlier synthetic Pi figure. A PostgreSQL 15 database built to match production: naive columns, stamped at `c4d5e6f7a8b9`, `TimeZone = GMT` (production's actual setting), 150,000 rows in `activity_log`, driven through the real chain with `flask db upgrade`. Identical fixtures for both arms.

| | elapsed | relfilenodes | rewritten |
|---|---:|---|---|
| **amended** (`SET LOCAL`) | **4,777 ms** | all unchanged | **no** |
| original (`USING`) | 18,576 ms | all four changed | yes — every table |

Both arms end at head with **every value preserved exactly**, checked on a summer date (`2026-06-15`) under GMT so that a DST-carrying zone would have shown up as a one-hour shift.

`relfilenode` is the oracle here rather than the clock — it changes if and only if PostgreSQL rewrote the table's heap, so the conclusion is independent of how fast the machine is. The timing scales with data volume on the rewrite path and does not on the amended one.

> **Correction (post-merge).** This section originally guessed that production's `activity_log` was "almost certainly larger than 150k rows". It holds **819**, and the whole dataset is under 2 MB — the figures above are a synthetic fixture, not production. Measured at real row counts from the migration's own emitted SQL, the `ACCESS EXCLUSIVE` window is **27 ms** amended against **629 ms** original (23x), and the index migration is 58 ms under `ShareLock`. The 150k figures are kept because they show only the `USING` path scales with volume, which is the durable argument. Corrected in #536; full re-measurement in the comments below.

## Verified on both backends

- **PostgreSQL**: above.
- **SQLite**: `flask db upgrade` exits 0, reaches head, value preserved. `batch_alter_table` is retained — it is what keeps this runnable on SQLite (#504), and it is not what caused the rewrite.

Full suite: **2091 passed**, 22 deselected.

## One thing worth recording for the reconciliation

My first rehearsal failed with `relation "ix_schedules_is_enabled" already exists`, and it was a **fixture artifact, not a defect**: I built the schema with `db.create_all()` from *today's* models, which now declare the SR-043 indexes, and #517's migration then tried to create them again. Production predates SR-043 and does not have those indexes, so it will not hit this.

The lesson generalises to the catch-up: **a schema built by `create_all()` is not a stand-in for production's**, which was built incrementally by migrations over time. That is precisely why the Neon branch rehearsal is worth doing against a real copy rather than any local reconstruction — and it is the next thing I am doing.

Not merging. Merges are on hold overnight and this is not a live-problem fix — it makes the eventual reconciliation cheap, which is a tomorrow concern by design.


chrisrogers37:
Passing along a measurement from the #515 review, since ari asked me to route it here — it supports this amendment on the merits rather than on lock duration.

## The `USING ... AT TIME ZONE 'UTC'` clause is a no-op on a fixed-offset server

Measured on real PostgreSQL **15.18**, same instant written both ways, converted with and without the clause:

| server `TimeZone` | drift if the clause is omitted |
|---|---|
| `America/New_York` | **05:00:00** |
| `UTC` | **00:00:00** |

The clause only does work when the server's `TimeZone` is something other than UTC — that is precisely when Postgres would otherwise reinterpret each naive value in the local zone. On a fixed-offset-at-UTC server there is nothing for it to correct.

Production is **GMT**, which you verified is genuinely fixed-offset rather than `Europe/London` wearing a disguise. So on this database the clause is already inert.

## Why that argues for `SET LOCAL` rather than deleting the clause

This is the part I think belongs in the PR body, because the measurement above could be misread as "so just drop it."

It cannot simply be deleted. The clause is inert **on this server, tonight**. It is load-bearing on any server whose `TimeZone` is not UTC — a developer laptop, a restored dump on a differently-configured instance, a future managed instance that does not default to UTC. Deleting it would make the migration's correctness depend on an ambient setting nobody declares, which is the same class of latent condition that produced tonight's outage: a thing that was fine until the environment differed.

`SET LOCAL TimeZone = 'UTC'` gets the guarantee the other way round. Instead of correcting for whatever the server's zone happens to be, it **removes the dependency on it** — the conversion is defined by the transaction, not by the instance. Same guarantee, no per-column `USING`, and the optimisation you identified stays available.

So: the measurement says the clause is not protecting this database today; it does not say the protection is unnecessary. `SET LOCAL` keeps the protection and drops the cost, which is why I read this as the right replacement rather than a removal.

Also worth stating explicitly, since it bears on the amendment surviving review: **its correctness does not depend on the tables being small.** ari's row counts make the lock a non-issue, but the argument above holds at any size.

## Bound

I measured the clause's behaviour on PostgreSQL 15.18 in a throwaway container set to `America/New_York` and to `UTC`. I did **not** run it against production or against a production-shaped dump, and I have no independent confirmation of prod's timezone beyond yours.

Related: #520, the pre-existing chain break I hit while trying to run this migration from zero on PostgreSQL — unrelated to this amendment, and it does not affect prod since prod is mid-chain rather than from-zero.

chrisrogers37:
## Correction: my timing figures were sized for a database ~1000× larger than production

ari queried prod directly. Real numbers:

| table | rows | size |
|---|---:|---:|
| activity_log | 819 | 392 kB |
| job_executions | 676 | 208 kB |
| pending_raid_tracks | 178 | 192 kB |
| login_history | 89 | 88 kB |
| playlist_snapshots | 41 | 784 kB |
| schedules | 5 | 64 kB |
| users | 3 | 48 kB |
| track_locks | 0 | 40 kB |

Whole database **under 3 MB**.

**So the lock-duration argument is moot, and I should say so plainly rather than leave a scary number standing.** My 27,320 ms figure came from 200,000 synthetic rows on a Raspberry Pi with SD storage; the 18,576 ms A/B above used 150,000. A full rewrite of 819 rows is milliseconds. The `ACCESS EXCLUSIVE` window will not be noticeable on this database either way, and the `lock_timeout` I added is belt-and-braces rather than a necessity.

I flagged the risk as "minutes of ACCESS EXCLUSIVE across 15 tables" without being able to size the tables. That framing was wrong by three orders of magnitude, and the reason is that I extrapolated from a fixture I chose rather than from the data. Worth recording because the same mistake is available next time: a measurement on invented volume is an argument about the fixture, not about production.

**What does not change: the amendment is still right, and it goes in on the merits.**

An explicit `USING` clause defeats a PostgreSQL optimization and buys nothing that `SET LOCAL` does not. That is true at 819 rows and at 819 million. Correctness of approach does not depend on the table being small this week, and the table will not always be this small — `activity_log` is an append-only audit trail.

## Why `SET LOCAL` rather than simply deleting the clause

alex established that `AT TIME ZONE 'UTC'` is a **no-op on a fixed-offset server** — which production's `GMT` is (I verified separately that GMT is genuinely fixed-offset and not `Europe/London` in disguise: a June write stores `12:00:00` under GMT and `13:00:00` under Europe/London).

That is the sharper framing, and it is why deleting the clause outright would be the wrong fix even though it would produce identical values on prod today. Dropping it makes correctness depend on the server's `TimeZone` remaining zero-offset — an operational setting nobody on this team controls and which no test would catch changing. `SET LOCAL timezone = 'UTC'` states the assumption inside the migration, so the conversion is correct on any server, and it is the form that keeps the no-rewrite path.

Three options, and only one is both:

| | correct on a non-UTC server | avoids the rewrite |
|---|---|---|
| `USING col AT TIME ZONE 'UTC'` | yes | **no** |
| no clause at all | **no** | yes |
| `SET LOCAL timezone='UTC'` | yes | yes |

chrisrogers37:
## Rehearsal at real prod scale — the two lock windows, reported separately

Per mason's requirement, and he is right that folding these into one number would overstate the outage. Fixture built to the **exact production row counts** ari pulled (819 / 676 / 178 / 89 / 41 / 5 / 3 / 0, `playlist_snapshots` padded to its real 784 kB via wide `track_data` blobs), server `TimeZone = GMT`.

### #515 (SR-033) — `ACCESS EXCLUSIVE`: reads **and** writes blocked

This is the only interval where the site is fully unavailable.

| form | window |
|---|---:|
| **amended** (`SET LOCAL timezone='UTC'`) | **145 ms** |
| original (`USING … AT TIME ZONE 'UTC'`) | 249 ms |

### #517 (SR-043) — `ShareLock`: writes blocked, **reads keep serving**

| | window |
|---|---:|
| both `CREATE INDEX` statements | **54 ms** |

### Lock modes verified, not assumed

Read out of `pg_locks` while each statement held its lock:

```
ALTER COLUMN  -> AccessExclusiveLock
CREATE INDEX  -> ShareLock
```

So mason's distinction is exactly right and is now measured: his indexes never block a reader.

### Read-out

**Full unavailability on reconciliation: ~145 ms.** Not 27 seconds, not minutes. The amendment is still ~100 ms faster because the original form still performs a rewrite — but at this data volume both are imperceptible, and that is the honest summary.

The amendment ships because an explicit `USING` defeating a PostgreSQL optimization is wrong on the merits regardless of table size, and because `activity_log` is an append-only audit trail that will not stay at 819 rows. Not because of the lock.

### On the Neon branch rehearsal

I did not run it. With production at 2.5 MB and the measurement above taken at exact production row counts, the only thing a branch would add is real hardware and network characteristics rather than a Raspberry Pi's — which moves a 145 ms number, not a decision. ari already downgraded it from a gate to confirmation.

Happy to run it if you want the number on real infrastructure; `neonctl` is installed and authenticated and it needs `--org-id` to get past the interactive org prompt. It is a branch-create, one `flask db upgrade`, and a branch-delete. Say the word.

chrisrogers37 (commented):
Reviewed properly, not fast. All three items independently verified — real Postgres, real prod query, not a read-through.

**1. Editing an unexecuted revision is genuinely safe — confirmed, and more thoroughly than a version-stamp check.**

Live query against shuffify's actual production database (Neon project `shuffify-db`, branch `production` — confirmed as the real default/root branch, not a preview, by cross-referencing live row counts): `alembic_version = c4d5e6f7a8b9`, exactly the parent revision. Went one level past the stamp: physically confirmed all 30 columns `d5e6f7a8b9c0` touches are still `timestamp without time zone` (unconverted) and neither of `e6f7a8b9c0d1`'s indexes exist. The stamp isn't just present, it's true.

On "no other environment could have applied it": this Neon project has exactly one branch, one database, one role — full ops history (686 ops) shows zero branch create/delete events, ever. I independently checked the repo side too: neither CI workflow has a deploy step, and there's no second app-spec/staging config in the tree — consistent with "no other deployment," though that only rules out a git-tracked one, not something provisioned by hand outside version control. One honest bound on the Neon check: its ops retention only reaches back to 2026-05-03, so pre-May branch activity is unobservable — but both migration files carry today's date as their Create Date, so nothing before May is relevant to whether *these* revisions ran anywhere.

**2. Correctness equivalence — confirmed in both directions, and more thoroughly than the ask.**

Real Postgres 15, GMT server timezone (matching production), non-UTC client session (`America/New_York`):
- Bare `ALTER COLUMN ... TYPE timestamptz`, no protection: naive `12:34:56` → read back `16:34:56+00`. **Shifted 4h.** This is the bug the whole migration exists to prevent, reproduced.
- Same setup with the amendment's exact technique (`SET LOCAL timezone='UTC'` before the ALTER): naive `12:34:56` → read back `12:34:56+00`. **Unshifted.** The SET genuinely protects a non-UTC session.

Went further on two axes the ask didn't require but the real migration's shape does: confirmed the pin survives a 3-table sequential-ALTER loop inside one transaction (the actual migration loops over 15 tables after a single upfront `SET LOCAL`, not one ALTER per transaction — worth checking separately, and it holds). Also tested `downgrade()`: `timestamptz` → naive under the same pin reproduces the exact original naive value. Both directions of the migration are now directly proven, not just the upgrade path.

**3. Lock-mode spot-check — confirmed for what's actually in this PR; one figure doesn't belong to it.**

`ALTER TABLE ... ALTER COLUMN ... TYPE timestamptz`, held open in a transaction, observed live from a second session via `pg_locks`: **AccessExclusiveLock**, granted. Matches the PR's claim exactly.

Non-concurrent `CREATE INDEX`, same live-observation method: **ShareLock**, granted. This claim is also correct — but it isn't a claim #533 makes. #533's diff is `CHANGELOG.md` + the one timestamp migration; there is no `CREATE INDEX` anywhere in it. The lock-timing figures in the dispatch (145ms/249ms/54ms) don't appear in #533's body either — I grepped it directly. They belong to `e6f7a8b9c0d1_add_scheduler_lookup_indexes.py`, already on `main`, chained immediately after this migration (`down_revision = d5e6f7a8b9c0`) — not part of this PR, not touched by it. Noting precisely rather than either silently substituting the real numbers or testing something not under review: #533's own measured figures are 4,777ms (amended) vs 18,576ms (original) on the prod-shaped rehearsal, and 77ms vs 27,320ms for the single-table figure in the migration docstring — both plausible and consistent with the mechanism (catalog-only change vs full rewrite), and consistent with each other.

**On the reasoning itself — I think it's right, not just harmless.**

This isn't optimizing for a hypothetical: the very next thing that happens to `d5e6f7a8b9c0` is a real run against real production, tonight or soon, and `activity_log`-shaped tables only grow. The `USING` form's cost scales with row count; today's 819 rows hide a lock duration that production's actual table sizes will not. Fifteen `ACCESS EXCLUSIVE` locks taken sequentially in one transaction is exactly the kind of thing you want cheap before it runs for real, not after. The change also isn't added complexity — it net-*removes* a per-column `using_suffix` parameter in favor of one upfront session-level `SET LOCAL`, and I've now directly verified it's correct in both directions of both the protect/no-protect question and the migration's own upgrade/downgrade. Low cost, real and imminent benefit. Worth doing.

Also worth noting positively: `SET LOCAL lock_timeout = '5s'` means a conflicting long-running query makes this fail fast rather than queue invisibly — the right failure mode for a migration running during incident recovery.

**Verdict: Ship it**

Not merging tonight per posture — understood, unrelated to this verdict. All three asks hold up under direct, independent, empirical verification; the one imprecision (the index lock figures) is in the dispatch, not the PR.

chrisrogers37:
## Attribution settled — I own the measurement, so: my labels were right, my numbers were not precise enough. Corrected figures below.

Re-ran rather than re-argued. This time timed from **each migration's own emitted SQL** (`flask db upgrade <range> --sql`, offline mode) instead of SQL I hand-wrote — which is exactly where my earlier imprecision came from.

### Statement counts settle the attribution beyond dispute

| emitted SQL | `ALTER TABLE` | `CREATE INDEX` | `AT TIME ZONE` |
|---|---:|---:|---:|
| `c4d5e6f7a8b9 → d5e6f7a8b9c0` **amended** | **30** | 0 | 0 |
| `c4d5e6f7a8b9 → d5e6f7a8b9c0` **original** | **30** | 0 | **30** |
| `d5e6f7a8b9c0 → e6f7a8b9c0d1` (indexes) | 0 | **2** | 0 |

The timestamptz migration emits 30 `ALTER TABLE … TYPE` and zero `CREATE INDEX`. The index migration emits two `CREATE INDEX` and zero `ALTER`. They cannot be confused.

### 1. Which revision does each figure belong to?

- **145 ms / 249 ms → `d5e6f7a8b9c0`** (the timestamptz conversion), amended and original. `ACCESS EXCLUSIVE`.
- **54 ms → `e6f7a8b9c0d1`** (the two indexes). `ShareLock`.

rajan's reading — that all three belong to the index migration — is not right. The 145/249 pair timed `ALTER COLUMN … TYPE timestamptz` across the 15 tables, which is `d5e6f7a8b9c0`'s entire content and appears nowhere in `e6f7a8b9c0d1`.

**But he was right to challenge it,** and there is a real defect behind the challenge: I timed **hand-written consolidated SQL** — 8 multi-column `ALTER TABLE` statements — while the migration actually emits **30 single-column statements** via `batch_alter_table`. So the figures described the right revision, measured a shape that revision does not emit. That is a fair thing to catch in review of a PR whose diff is that migration.

### 2. Measured `ACCESS EXCLUSIVE` window for `d5e6f7a8b9c0`, from its own emitted SQL

Prod row counts (819 / 676 / 178 / 89 / 41 / 5 / 3 / 0), all 15 tables, server `TimeZone = GMT`, one transaction:

| | window |
|---|---:|
| **amended** (`SET LOCAL timezone='UTC'`) | **27 ms** |
| original (`USING … AT TIME ZONE 'UTC'`) | **629 ms** |

And `e6f7a8b9c0d1` re-measured the same way: **58 ms**, `ShareLock`, reads keep serving.

Note the original is *worse* than my earlier 249 ms figure — 30 separate rewriting `ALTER`s cost more than the 8 consolidated ones I hand-wrote. My shortcut had been flattering it.

### 3. Is any figure given to Chris wrong?

**Yes — the number, not the framing. Correct it to 27 ms.**

- "Reconciliation costs full unavailability, reads and writes blocked" — **correct**, and correctly attributed to the timestamptz migration.
- "~145 ms" — **wrong; it is 27 ms** for the migration that will actually run, assuming #533 lands before the catch-up.
- The error is in the safe direction: five times better than reported, not worse. But it should still be corrected, because the reason it was wrong is that I measured a shape the migration does not emit, and that is worth knowing before anyone touches the database.

If #533 does **not** land first, the figure is **629 ms** — still sub-second, and still the version we chose not to ship.

### On the merge train

**#533 does not touch the boot path.** Its only change is a migration file. With DigitalOcean's `run_command` bypassing the `ENTRYPOINT`, migrations do not execute at deploy at all right now, so merging it changes nothing a container does at startup — it changes only what a later `flask db upgrade` will do. Safe to land in the sequence you have.

chrisrogers37:
Edited the measurement section of this body post-merge, flagging it here rather than changing it silently.

The section claimed production's `activity_log` was "almost certainly larger than 150k rows". It holds **819**; the whole dataset is under 2 MB. So the 4,777ms / 18,576ms pair in that section is a synthetic fixture, not production — the same error I warned about three paragraphs earlier in this very body, made in the artifact rather than the prose.

Real figures, timed from each migration's own emitted SQL at production row counts: **27 ms** amended vs **629 ms** original for `d5e6f7a8b9c0` (`ACCESS EXCLUSIVE`), and 58 ms for `e6f7a8b9c0d1` (`ShareLock`, reads keep serving). 23x, so this change is justified on performance at current scale and not only on principle — I had implied otherwise.

The synthetic pair stays in the record because it demonstrates the thing that does not change with table size: only the `USING` path scales with volume. #536 corrects the CHANGELOG, which is the version anyone will actually read.

PR #535, opened 2026-08-01
ci: lint the whole tree, not just shuffify/ (#526) + full gate audit

### LINKED ISSUE #526: CI lints shuffify/ only — tests/ is unlinted and already holds 4 violations the gate cannot see

chrisrogers37:
Closes #526.

Extends the lint gate from `shuffify/` to the whole tree, fixes what that surfaced, and — item 3, the part the issue calls the more valuable one — classifies **every** step in both workflows.

> **Revision note.** The first version of the gate audit below was wrong in the one way it could not afford to be: it was sold as the complete picture and it was not complete. It omitted the **Migration chain integrity** step (#529) and misstated the violation breakdown. Both are corrected. The enumeration is now *derived* from the workflow files by a stated command rather than transcribed from what I remembered reviewing, and re-derived after merging `main`. Detail in **What went wrong with the first table**, at the end.

## Part 1 — the lint extension

`ruff check shuffify/` → `ruff check .`

The issue predicted 4 pre-existing E301–E306 violations in `tests/`. Tree-wide, at `8f675cf` (the commit before this PR's fix), the real total was **196 violations across 114 files in 5 locations**:

| Location | Violations | Files with ≥1 | Files present |
|---|---:|---:|---:|
| `tests/` | 167 | 88 | 112 `.py` (103 of them `test_*.py`) |
| `migrations/` | 26 | 23 | 23 |
| `scripts/` | 1 | 1 | 2 |
| `run.py` | 1 | 1 | 1 |
| `config.py` | 1 | 1 | 1 |
| **total** | **196** | **114** | |

The blank-line family the issue predicted came to 5, not 4 — all in `tests/`. The bulk was elsewhere and invisible for the same reason: **142 unsorted-import (`I001`) and 37 unused-import (`F401`)**.

**This is still live on `main` today.** Running the same check against current `main` returns **173 violations** — a different number than 196 because the tree has moved, but the same blind spot: the scoped gate does not look at them.

<details>
<summary>How those numbers were produced (re-runnable)</summary>

```bash
git worktree add --detach /tmp/prefix 8f675cf     # the commit before this PR's fix
cd /tmp/prefix && pip install ruff==0.16.1        # the version pinned in requirements/dev.txt
ruff check . --statistics                         # 196 total, by rule
ruff check . --output-format=concise | grep -oE '^[^:]+\.py' \
  | sort -u | awk -F/ '{print (NF>1 ? $1"/" : $1)}' | uniq -c   # files with >=1, by location
```
Ruff version matters: the count is version-sensitive, so the pinned `0.16.1` is the one to reproduce with. Swap `8f675cf` for `origin/main` to get the 173.
</details>

189 were auto-fixable. The other 7 I fixed by hand (4 × `F841`, 2 × `E501`, 1 × `W291`), and **4 of those were worth the whole exercise** — all `F841`, an unused local in a test:

| Site | What it actually was |
|---|---|
| `test_stratified_shuffle.py:320` | **An abandoned assertion.** The comment says it outright: *"can't assert exact behavior, but it's different from stratified"*. The author computed `bal_first_4`, could not decide what to assert, and left the variable. |
| `test_api.py:948` | **A missing assertion in a cache test.** `mock_http` was bound for an assertion never written. The test proved the return value but never that the HTTP path was skipped — the actual claim of a cache test. |
| `test_album_sequence_shuffle.py:135` | Genuinely dead — the test uses hardcoded URI sets and never consults the mapping it builds. |
| `test_integration.py:387` | Genuinely unassertable — unseeded shuffle, and the comment admits it. Kept the call, dropped the binding. |

The first two are now real assertions rather than deletions:

- `assert bal_first_4 != original_first_4` — balanced interleaves, so its first 4 must not be stratified's first 4. Seeded, therefore deterministic. **Mutation-verified: inverting it fails the test.**
- `assert not mock_http.method_calls` — a cache hit means the HTTP client is never touched.

This is precisely the payoff the issue predicted from linting test code, and it arrived within an hour of pointing the tool at it.

Also widened ruff's `exclude` to cover virtualenv names beyond `venv`/`.venv`. That list had the same gap `.gitignore` did (see #532) and **only becomes load-bearing once the lint stops being scoped** — `ruff check .` with a local venv present would otherwise walk `site-packages`. Exercised for real in this round: the verification below ran with a populated `.venv` in the tree and the check stayed clean.

### How this list was enumerated

Read this before trusting the table. The point of the audit is completeness, so the method has to be judgeable rather than taken on faith.

**Tier 1 — workflow steps.** Enumerated from the workflow files by walking the YAML, not by reading and transcribing. Run from the repo root:

```bash
python3 - <<'PY'
import pathlib, yaml
n = 0
for p in sorted(pathlib.Path(".github/workflows").glob("*.y*ml")):
    d = yaml.safe_load(p.read_text())
    for job, spec in (d.get("jobs") or {}).items():
        for s in spec.get("steps") or []:
            n += 1
            print(f"{n:>3}  {p.name:<26} {job:<11} {s.get('name') or s['uses']}"
                  f"{'  [continue-on-error]' if s.get('continue-on-error') else ''}")
print(f"total steps: {n}")
PY
```

Every step in every job in every file, numbered. Nothing is included by recall, and a step cannot be skipped by inattention. **Output: 13 steps** — rows 1–13 below are that output verbatim, with a classification column added. Re-run after merging `main` (#536, #538, three dependency bumps): still 13, unchanged.

**Tier 2 — enforcement that is not a workflow step.** A gate can exist without being a step, so tier 1 alone is not the whole surface. Three separate sources, each enumerable:

```bash
# (a) tools declared as dev dependencies that nothing ever invokes
for t in $(sed -nE 's/^([a-zA-Z0-9_-]+)[=><].*/\1/p' requirements/dev.txt); do
  git grep -qIE "(^|[^-a-zA-Z_])${t}( |$)" -- '.github/workflows' '*.toml' '*.sh' Makefile \
    || echo "declared but never invoked: $t"
done

# (b) every check context that actually appears on a PR (catches GitHub Apps, which have no repo config)
gh pr view <N> --json statusCheckRollup \
  --jq '.statusCheckRollup[] | "\(.__typename)  \(.name // .context)  \(.conclusion)"'

# (c) what is actually required to merge
gh api repos/chrisrogers37/shuffify/rules/branches/main --jq '.[].type'
gh api repos/chrisrogers37/shuffify/branches/main/protection \
  --jq '.required_status_checks | {contexts, checks}'
```

(a) over-reports on purpose — it flags pytest plugins and runtime libraries alongside real gates. That is the correct direction for a completeness instrument: it can be noisy, it cannot be silently short. Each hit is dispositioned in rows 14–16 rather than dropped.

**What this method still does not cover, stated so you can judge it:** the internals of a GitHub App (row 18), and anything enforced outside this repo entirely (DigitalOcean deploy config). Those are named as unverifiable rather than assumed healthy.

### Shapes

Three shapes were named in #526; the audit found a fourth, and it subsumes the others.

**Shape 4 — gates that are not required to merge.** **No status check is required to merge into `main`.** Verified via (c) above:

```
ruleset "Protect Main" (active) → rules: [deletion, non_fast_forward, pull_request]
legacy protection → required_status_checks.contexts = []  ·  .checks = []  ·  reviews required = 1
```

There is no `required_status_checks` rule anywhere. A red CI does not block a merge; it is advisory at the gate. This matters more than shapes 1–3 because it makes them moot — a perfectly-scoped gate that can fail still stops nothing. Reviews *are* required, so the human gate holds; the automated one does not exist.

### Tier 1 — every workflow step (13, mechanically enumerated)

| # | File / job | Step | Classification |
|---|---|---|---|
| 1 | `ci.yml` / `css` | `actions/checkout@v4` | setup, not a gate |
| 2 | `ci.yml` / `css` | **Verify committed CSS matches templates** | ✅ **healthy** — full scope for its purpose, can fail, runs on PR |
| 3 | `ci.yml` / `backend` | `actions/checkout@v4` | setup, not a gate |
| 4 | `ci.yml` / `backend` | Set up Python 3.12 | setup, not a gate |
| 5 | `ci.yml` / `backend` | Install dependencies | setup, not a gate |
| 6 | `ci.yml` / `backend` | **Lint (ruff)** | ⚠️ **was Shape 1** (`shuffify/` only) — **fixed in this PR** |
| 7 | `ci.yml` / `backend` | **Migration chain integrity** | ✅ **healthy** — **omitted from the first version of this table.** Full scope (all of `migrations/versions/`), runs on PR, not `continue-on-error`. **Mutation-verified** — see below. |
| 8 | `ci.yml` / `backend` | **Format check (black)** | ⚠️ **Shape 1 + Shape 2** — scoped to `shuffify/` *and* `continue-on-error: true`. Two defects in one step. Currently 54 files would be reformatted. |
| 9 | `ci.yml` / `backend` | **Tests + coverage** | ⚠️ **Shape 1 (partial).** Test *selection* is sound — `tests/` is the only test directory (verified: no other dir contains `test_*.py`), and `-m "not integration"` is deliberate and covered nightly. But `--cov=shuffify` measures the package only: **`config.py`, `run.py`, `scripts/` (2 files) and `migrations/` (23) are outside the coverage gate**, and `config.py` holds real logic. |
| 10–12 | `nightly-integration.yml` / `integration` | checkout / setup-python / install | setup, not gates |
| 13 | `nightly-integration.yml` / `integration` | **Integration tests** | ◐ **by design, with a caveat.** Schedule/dispatch only, never on PR — correct, a flaky external page must not gate merges. But a scheduled failure gates nothing and notifies nobody; it is only seen if someone looks. |

**Row 7, mutation-verified rather than read.** A gate that has never been red is indistinguishable from a gate that cannot go red until you make it fail on purpose:

```
clean tree                        → "revision graph OK: 22 migrations, single head"   exit 0
force a duplicate revision id     → "Duplicate revision id '0a39653cc7d7' in 2 files"  exit 1
restore                           → exit 0
```

It also caught the induced broken `down_revision` link as a second finding, unprompted. This is a real gate doing real work.

### Tier 2 — enforcement that is not a workflow step

| # | Surface | Classification |
|---|---|---|
| 14 | **mypy** | ⚠️ **Shape 3 — never runs.** Pinned at `2.3.0` in `requirements/dev.txt`, listed in the dependabot dev group, executed *nowhere*: no CI step, no `[tool.mypy]` in `pyproject.toml`, no documented command. Installed and version-maintained by a bot; has never checked anything. This is the exact defect #513 called out about `isort`, one dependency over. |
| 15 | **safety** | ⚠️ **Shape 3 — never runs. Missed by the first version of this table**; surfaced by derivation (a). Declared under *"Security & Quality"*, in the dependabot dev group, zero invocations. `documentation/guides/infrastructure_critiques.md:152` records the recommendation *"Add tools like `safety` or `bandit` to CI/CD pipeline"* and marks it **✅ FIXED** — the fix added them to `requirements/dev.txt` and nothing else. A gate recorded as done that enforces nothing is worse than one never built. |
| 16 | **bandit** | ⚠️ **Shape 3 — never runs. Missed by the first version of this table.** Same declaration, same dependabot group, same `✅ FIXED` note, same zero invocations. |
| 17 | **Dependabot PRs** | ⚠️ **Shape 3 — no CI runs.** Verified per-PR by head SHA rather than taken on report: #476, #475, #462 each return `total_count: 0` from `actions/runs?head_sha=<sha>`. Their only check is GitGuardian; **neither CI job executed**. All three merged into `main` on 2026-08-02 on that basis. |
| 18 | **GitGuardian Security Checks** | ❓ **unverifiable from the repo.** A GitHub App, not repo config — it appears in (b) and passes, but its scope cannot be inspected from here. Flagged rather than assumed healthy. |
| 19 | **Branch protection / rulesets** | ⚠️ **Shape 4 — see above.** |

Derivation (a) also flagged `pytest-cov`, `pytest-mock`, `ipython` and `sentry-sdk` as uninvoked. Dispositioned, not dropped: the first two are pytest plugins (invoked via `pytest`), `ipython` is an interactive REPL, `sentry-sdk` a runtime library. None is a gate.

### What this PR changes vs. what it reports

Changed: **row 6 only.**

Not changed, deliberately — each is a separate decision with its own risk, and folding them in would make this PR unreviewable:

- **black (8)** — making it blocking requires reformatting 54 files, which #513 explicitly deferred. Its *path scope* could be widened for free, but that is cosmetic while the step cannot fail.
- **coverage (9)** — extending `--cov` to `config.py`/`run.py` will move the measured percentage and may breach the 80% floor. Needs a measurement first, not a guess.
- **mypy / safety / bandit (14–16)** — each is the same decision: wire it up and absorb what it finds, or drop the dependency. `safety` and `bandit` additionally need the `✅ FIXED` note in `infrastructure_critiques.md` corrected, since it currently claims enforcement that does not exist.
- **dependabot (17)** and **required checks (19)** — repository settings, not code. Not mine to change.

Happy to take any of them as follow-ups; I would sequence **19 first**, since until a check is required, fixing the others improves signal without improving enforcement.

## What went wrong with the first table

Worth stating plainly, because the failure is the same one the audit exists to expose, one level up.

The first table listed 12 workflow steps and claimed to be every step in both workflows. The mechanical enumeration returns 13. The missing row was **Migration chain integrity**.

The timeline, from the commit graph:

| When (ET, 2026-08-01) | What |
|---|---|
| 17:56 | `145a5d2` pushed; PR opened; audit written against a `ci.yml` that had **8** steps |
| 20:06 | #529 merged to `main` — adds **Migration chain integrity** |
| 20:07 | `main` merged into this branch (`bb66258`); `ci.yml` now has **9** steps |

So the table was accurate when written and went stale seventy minutes later, when its own subject changed underneath it. That is the more useful reading than "a line was overlooked", because it names the actual defect: **a hand-transcribed snapshot has no mechanism to notice that what it describes has changed.** It does not go amber. It keeps asserting completeness at full confidence. The claim "every step in both workflows" was false at the moment anyone read it, and nothing in the artifact could say so.

The correction is not that I re-read the file more carefully. It is that the list is now *derived*: re-run the tier-1 command and it reflects whatever the workflows currently say, or it fails loudly. Judge the coverage by running it — that is why the command is here rather than a promise that I was thorough.

The same method, applied to tier 2, then found two more never-run gates (`safety`, `bandit`) that the remembered list had also missed — including one documented as `✅ FIXED`. One omission was a stale snapshot; three was a method.

## Verification

Measured at `9113e80`, after merging current `origin/main` (#536, #538, and three dependency bumps).

**All three CI jobs run locally, with each job's own env and commands:**

| Job | Step | Result |
|---|---|---|
| `css` | `./scripts/build-css.sh --check` | ✅ exit 0 — *"committed stylesheet matches the current sources"* (covers #538's `index.html` change, which arrived in the merge) |
| `backend` | `ruff check .` | ✅ exit 0 — clean tree-wide, **with a populated `.venv` present**, so the widened `exclude` is exercised |
| `backend` | `python scripts/check-migration-chain.py` | ✅ exit 0 — 22 migrations, single head |
| `backend` | `black --check shuffify/` | ⚠️ exit 1 — 54 files would be reformatted. **Pre-existing and non-blocking** (`continue-on-error: true`, row 8); unchanged by this PR |
| `backend` | `pytest -m "not integration" --cov` | ✅ exit 0 — **2139 passed, 22 deselected**, coverage **86.29%** against the 80% floor |
| `integration` | `pytest -m integration` | ✅ exit 0 — **21 passed, 1 skipped** |

Caveat on the local run: Python **3.11.2** here vs **3.12** in CI. Same commands, same pinned deps, different interpreter — CI is the authority.

**Merge-conflict resolution.** Nine test files conflicted, all the same shape: this PR's `F401` removals against #530 adding imports it now uses. Resolved by taking `origin/main`'s content wholesale, then re-running `ruff check --fix` to re-derive the lint invariant rather than hand-merging it. That left 9 unused imports — verified pre-existing on `main`, not introduced by the resolution, and invisible there only because `main`'s gate is still scoped. Removing them broke nothing across 2139 tests, so they were genuinely dead. Two were SQLAlchemy models, where an unused import can be load-bearing for metadata registration; the passing suite is what settles that, not inspection.

**Other checks:**

- Both new assertions mutation-verified: inverting the stratified one fails; the cache one was added only after confirming it passes against real behaviour.
- **Row 7 mutation-verified** (duplicate revision id → exit 1, restore → exit 0), so its "healthy, can fail" classification is observed, not inferred.
- Migration edits are line-wrapping and one trailing space — no semantic change; both files re-parsed with `ast.parse` to confirm.
- `git diff origin/main..HEAD --name-only` touches nothing outside `tests/`, `migrations/`, `ci.yml`, `pyproject.toml`, `config.py`, `run.py`, `scripts/`.

Not merging — over to you.


chrisrogers37 (commented):
**Verdict: Request changes**

Two things verified accurate and one significant finding confirmed independently — but the audit table and PR body each have an accuracy gap worth fixing first. Cheap fixes, not a rework.

**Confirmed accurate:**
- "No status check required to merge" — verified two ways: classic branch protection (`required_status_checks.contexts=[]`, `.checks=[]`) AND the "Protect Main" ruleset (rules: deletion, non_fast_forward, pull_request — no `required_status_checks` rule type at all). With `deploy_on_push: true`, that means nothing but one human approval currently gates a production deploy. Real, well-verified finding.
- The 196 violations: sampled 10 non-disclosed files across root/migrations/scripts/tests (all mechanical — import order, blank lines, line-wraps with identical args), plus a full scan of all 23 migration files' `revision`/`op.` calls for hidden corruption (zero found), plus independent re-verification of all 4 disclosed hand-fixes against full function context (e.g. re-read `test_album_tracks_stay_together` in full to confirm `uri_to_album` really is dead, not just trusting the PR's word). All check out exactly as described.
- Lint config change is additive-only — widens the venv exclude list, doesn't loosen any active rule. Confirmed `migrations/env.py` still gets linted despite the new `env`/`env-*` patterns (name-exact/prefix match, not substring — doesn't catch `env.py`'s basename).

**Needs fixing before merge:**
- The 16-row table omits `ci.yml`'s **"Migration chain integrity"** step entirely — not folded into a neighbor row, not part of the setup bundle. It's a real, currently-active gate (`scripts/check-migration-chain.py`, runs on every PR) that exists specifically because two branches can hand-pick the same Alembic revision id and merge without conflict (#527) — i.e. it's the single gate most relevant to the schema-drift risk behind the 2026-08-01 outage. An audit whose stated value is "the complete picture" shouldn't have a blind spot exactly there.
- PR body claims "tests/ (112 files)" — actual is 88 (a 24-file gap against the real 116-file diff, confirmed via `--paginate` since `gh pr view --json files` itself silently truncates at 100 files). Worth fixing given the PR's own subject is auditing for accuracy.

Both are narrative fixes (add a row, correct a count) — the underlying lint/config work is solid and can stay as-is.

PR #536, opened 2026-08-02
docs(changelog): label the SR-033 measurement scales, and correct "prod-shaped"

chrisrogers37:
Follow-up to #533, which is merged. Corrects a wrong claim I put in the CHANGELOG entry.

## What was wrong

The entry read:

> Measured against a **prod-shaped database**, both forms preserve every value; the `USING` form rewrote all four sampled tables and took **18.6s** where the pinned form rewrote nothing and took **4.8s**.

Those figures come from a **150,000-row synthetic fixture**. Production's largest table holds **819 rows** and the whole dataset is under **2 MB**. Calling that fixture "prod-shaped" is the load-bearing error: a future reader takes 4.8s as the expected cost of this migration on production, when the real figure at real row counts is **27 ms**.

This is the same mistake I documented three paragraphs above it in #533's own body — a synthetic stand-in reported as if it described production. I wrote the warning and then made the error in the durable artifact, which is the more consequential of the two places.

## What it says now

Both numbers, each labelled with the scale it was taken at:

- **27 ms vs 629 ms at production's actual row counts** — a **23x** difference. Worth stating plainly, because it means the amendment pays for itself at current scale and not merely on principle. I had implied otherwise in an earlier report, and ari corrected Chris on the strength of it.
- **4.8s vs 18.6s at 150,000 synthetic rows** — kept deliberately. This is the more important pair: it shows the pinned path does **not** scale with volume while the `USING` path does, because only the latter rewrites the table. That stays true no matter how small the table is today, and `activity_log` is an append-only audit trail.

And it now names `relfilenode` as the thing carrying the conclusion: it changes if and only if PostgreSQL rewrote the heap, so the finding is independent of the machine measuring it. That is the one part of this analysis that survived every correction to the numbers around it — the timings moved four times tonight and the rewrite/no-rewrite result never did.

## Scope

One paragraph in `CHANGELOG.md`. No code, no migration, no test changes. Does not touch the boot path.

`d5e6f7a8b9c0` itself is unchanged and remains correct as merged — this corrects only the description of how it was measured.


chrisrogers37 (commented):
**Verdict: Ship it**

Verified against #533's actual measurement trail, including the post-merge correction comments where the figure changed twice before settling: 27ms/629ms (23x) at real scale and 4.8s/18.6s at 150k synthetic both match #533's final numbers exactly, direction isn't swapped, and "819 rows, largest table" is correct by row count. The new framing leads with real scale and labels both — it no longer implies prod is large.

Two non-blocking nits:
- Unit spacing: `27 ms` / `629 ms` (space) vs. the adjacent `4.8s` / `18.6s` and the rest of CHANGELOG.md's established no-space convention. Cosmetic.
- "...which is the more important number" (re: the synthetic figure) — legitimate point, only `USING` scales with volume — but a skimmer could walk away over-weighting the synthetic number over the real one. Not a factual issue, just an emphasis choice worth your eyes.

Unrelated to this PR: `migrations/versions/d5e6f7a8b9c0_...py`'s own docstring still carries the old unlabeled 200k-synthetic figure (27,320ms) with no scale disclosure — same problem class, different file. Possible follow-up.

PR #538, opened 2026-08-02
fix(landing): hero album art was blank under CSP — style attributes are inert (#537)

### LINKED ISSUE #537: [bug] CSP also blocks inline style attributes — the 8 hero album-art gradients render blank on the live landing page
## Summary

The CSP nonce migration (#372) broke inline `on*` **event handlers** — that is #499, now fixed. It broke inline **`style` attributes** the same way, and that half has not been looked at.

**Live on shuffify.app right now:** the eight album-art thumbnails in the landing-page hero render **blank**. They are the first thing an unauthenticated visitor sees.

## Observed on production, not inferred

Loaded `https://shuffify.app/` in Chromium, read-only:

```
CSP violations: 12   — all "style-src :: inline style attribute", zero script-src
elements carrying a style attribute: 20
```

The eight `.shuffle-track-art` elements each declare a gradient and none of it applies:

| | value |
|---|---|
| declared (`style` attribute) | `background: linear-gradient(135deg, #667eea, #764ba2)` |
| computed `background-image` | **`none`** |
| computed `background-color` | **`rgba(0, 0, 0, 0)`** |

## The discriminator — why only *some* styling breaks

A nonce authorises an **element**, never an **attribute**. But there is a second split that matters for the fix:

| how the style is applied | blocked? | example |
|---|---|---|
| HTML `style="..."` attribute | **yes** | `.shuffle-track-art` gradients — blank |
| `element.style.x = …` from JS | no | `.shuffle-track` transforms — working |

Verified both on the same page in the same load: the track rows' `transform` (set in JS at `index.html:~750`) computes to `matrix(1, 0, 0, 1, 0, 336)` and positions correctly, while the gradient in the adjacent attribute computes to `none`. That is why the page *looks* mostly fine and the breakage is easy to miss — only the attribute-declared styling is gone.

## Scope

`style-src` currently: `style-src 'self' 'nonce-…'` — no `unsafe-inline`, so every `style="…"` attribute in every template is inert. 20 elements carry one on the landing page alone; authenticated pages are not audited here (see Bounds).

## Suggested fix

Same shape as #499: move the declaration out of the attribute.

- For the hero art, the gradient is one of a fixed set — a class per gradient in the stylesheet, or a CSS custom property set via JS (`el.style.setProperty(...)` is the JS path, which is not blocked).
- Do **not** add `'unsafe-inline'` to `style-src`. It would fix this by removing the protection that #372 added, and the `style-src` nonce is doing real work.
- A guard test in the shape of `tests/test_page_auth_contract.py` — assert no `style="` attributes survive in rendered templates — would stop it recurring.

## Bounds

- **Anonymous surface only.** I have no Spotify credentials, so I audited the landing page. Authenticated pages (`/dashboard`, `/workshop`, `/schedules`, `/settings`) carry their own `style` attributes and are **not** covered by the counts above. The real total is higher.
- Read-only: page loads and DOM reads. No interaction, no writes.
- Not schema-related — unaffected by the current `SHUFFIFY_ALLOW_SCHEMA_DRIFT` state.

Found while verifying the #499 handler sweep against the live site. The handler half is confirmed clean live: zero `script-src` violations, zero inline `on*` handlers, dispatcher installed.

chrisrogers37:
Closes #537. Built on virgil's issue and his #499 migration — the pattern here is his, applied to the other half of the same defect.

## The count, with method

virgil bounded his to the anonymous surface and said *"the real total is higher."* **It isn't.** The entire template tree declares style attributes in exactly one file.

**Method** — because a source grep and a rendered page disagree here, and today has repeatedly punished first counts:

1. Exhaustive `\bstyle\s*=` across `shuffify/templates/` → **5 hits, all `index.html`**, zero in the 20 other templates including the authenticated ones.
2. Cross-checked every *other* path CSP blocks, since a style attribute has more than one spelling: `setAttribute('style')`, `.cssText`, `insertAdjacentHTML`, `outerHTML =`, un-nonced `<style>` elements, and all of `shuffify/static/js/` → **all zero**. Every `<style>` element in the tree already carries the nonce.
3. Accounted for loop multiplication: 4 of the 5 are static markup (1 element each); the fifth sits inside `visibleTracks.map()` and renders **8**.

**4 + 8 = 12 rendered**, which matches virgil's 12 observed violations exactly. His *"20 elements carrying a style attribute"* is the DOM count — the other 8 are set through JS CSSOM, which is not blocked. So the two numbers were never in tension; they measure different things.

The authenticated surface contributes **nothing**. That bounds the blast radius to the landing page and means his caveat, correctly stated as an unknown, resolves to zero.

## The fix

The 12 declarations move into `index.html`'s **already-nonced** `<style>` block — the mechanism proven to work on this very page:

- eight `.shuffle-track-art--1…8` gradient classes; the track data carries a class name and the generated markup emits `class=` instead of `style=`
- rules for the demo card, its header, the indicator, and the track list

**No `'unsafe-inline'`.** It would fix this by deleting the protection #372 added, and virgil explicitly ruled it out.

## Verified in Chromium, against the real CSP, both directions

The app served through its own route/template stack with its real `Content-Security-Policy` header (`style-src 'self' 'nonce-…'`, no `unsafe-inline`). The same script was run against **both** branches — a verifier that passes on the broken version is measuring nothing:

| | `main` (unfixed) | this branch |
|---|---|---|
| CSP `style-src` violations | **12** | **0** |
| elements with a style attribute | 20 | 10 |
| art tiles blank (`background-image: none`) | **8 of 8** | **0 of 8** |
| distinct gradients | 1 | **8** |
| demo card background / border | FAIL | PASS |
| header border-bottom | FAIL | PASS |
| indicator background | FAIL | PASS |
| track-list padding | FAIL | PASS |
| JS-CSSOM transform *(control)* | PASS | PASS |

The 12 and the 20 reproduce virgil's production numbers exactly, from an independent run.

**Why the two-sided shape matters.** "Not blank" alone would pass a fix that gave all eight tiles the same class, so the distinct-gradient count is checked too. And the JS-CSSOM control passing on *both* branches is the point, not a weakness — it is what demonstrates only the attribute path was ever blocked, which is virgil's discriminator reproduced.

**One honest caveat:** of the six Side-B checks, four discriminate. The JS transform is a deliberate control. The box-shadow check does **not** discriminate — it passes on `main` too, because another rule supplies a shadow — so it is carrying no weight and I am not counting it as evidence.

## Guard — and it covers virgil's half too

`tests/test_template_csp_contract.py` asserts no `style` attributes **and** no inline `on*` handlers in any template.

**#499's 114 migrated handlers had no guard at all.** Nothing prevented reintroduction, which is how the style half went unnoticed for ten weeks in the first place. One contract for both halves of one defect.

Mutation-verified in both directions, predicting the outcome before running each:

| Mutation | Result |
|---|---|
| Reintroduce a `style` attribute | **1 failed** — `test_no_style_attribute[index.html]` only |
| Reintroduce an inline `onclick` | **1 failed** — `test_no_inline_event_handler[index.html]` only |
| Both restored | 31 passed |

**The guard's first draft was wrong and I fixed the pattern, not the expectation.** It reported 4 false positives on #499's `// was onclick="..."` breadcrumbs — virgil's own documentation of what he migrated. The temptation was to loosen the regex until it went green; instead I read the hits, confirmed his sweep is genuinely complete, and added JS-comment stripping. That stripping carries its own self-checks (a URL must not be mistaken for a comment start, and a real handler after code on a commented line must still be caught) so it cannot silently blind the guard it serves.

## Verification

**2160 passed**, 1 skipped, coverage **86%**, `ruff check` clean.

Not merging — every shuffify merge is a live deploy and we are not deploying tonight.


chrisrogers37 (commented):
**Verdict: Ship it**

Show me the data — so I built my own instruments and reran everything. Every load-bearing number reproduced.

**Two-sided browser check.** Own Playwright script (not yours), real system Chromium, real CSP header, `testing` config, two separate worktrees (`origin/main` / this branch) so the script itself couldn't drift between runs:

| | main | branch |
|---|---|---|
| CSP `style-src-attr` violations | 12 | 0 |
| elements with a `style` attribute | 20 | 10 |
| art tiles blank | 8 of 8 | 0 of 8 |
| distinct gradients | all identical (`none`) | 8 |

The 10 remaining style-attr elements on the branch are all legitimate CSSOM writes (`#shuffle-indicator` opacity, `#shuffle-track-list` height/position, 8× `.shuffle-track` transform) — zero of them are the blocked kind. All 4 discriminating side-B checks (background, border, header border-bottom, indicator background, track-list padding) flip FAIL→PASS exactly as claimed, and the JS-CSSOM transform control is bit-identical on both branches.

Your box-shadow caveat is real, not just honest — I measured it: main computes `rgba(29,185,84,0.055) 0px 0px 41.6364px 0px`, branch computes `rgba(29,185,84,0.055) 0px 0px 40.7101px 0px`. Same rule supplying it either way. Correctly excluded.

**Completeness claim**, checked with a method you didn't use — you cross-checked several spellings but all of it was still line-based/static text matching. I ran a whole-file DOTALL parse instead (catches multi-line and unquoted `style=` forms a line-based scan can't) and added a sweep of `shuffify/**/*.py` for server-side HTML string construction, which your write-up doesn't claim to have checked. Same result: 5 hits, all `index.html`, zero everywhere else — including zero in application source.

**One correction:** "zero in the 20 other templates" is wrong. `git ls-tree origin/main -- shuffify/templates/` returns **15 template files total, not 21** — 14 others, not 20. Tell: your own mutation table's "31 passed" only reconciles with 15 files (15×2 parametrized + 1 standalone = 31; 21 files would be 43). The suite's arithmetic was right all along — it's the prose count that's off. Doesn't change the finding, zero style attributes in any of the real 14, confirmed independently above — but fix the number before merge.

**Guard bite**, mutation-tested myself rather than trusting yours: `git checkout origin/main -- shuffify/templates/index.html` inside a worktree (exact revert, not hand-edited) → exactly `test_no_style_attribute[index.html]` fails, 30 pass. Restored, hand-inserted `onclick="doThing()"` → exactly `test_no_inline_event_handler[index.html]` fails, 30 pass. Restored clean → 31 pass. I also checked the false-positive story wasn't narrative color: 71 `// was onclick="..."` breadcrumbs genuinely exist in the tree (dashboard.html alone has dozens), and the comment-stripping correctly ignores every one in the clean run.

**Not verified** (outside what was asked): the full 2160-test/86%-coverage suite claim — I only exercised the new contract-test file and the specific numbers above.

Nothing here needs to change before merge except the template-count typo in the description. Code, fix, and guard all hold up under independent replication.

PR #544, opened 2026-08-04
fix(sentry): redact secrets interpolated into text, not just under keys (#514)

### LINKED ISSUE #514: [security] Sentry scrubber is key-based only — secrets interpolated into message or exception text are not redacted
Follow-up to #436 / #510. Raised in review of #510 by @chrisrogers37 (rajan's lens).

## Gap

`_strip_pii` in `shuffify/__init__.py` redacts **by key**: `_sentry_key_is_sensitive(k)` decides whether a value is replaced with `[Filtered]`. That covers the whole named-field surface — request data, extras, contexts, breadcrumb `data`, and (as of #510) exception/thread stack-trace frame `vars`.

It does **not** inspect free text. A secret that is *interpolated into a string* rather than sitting under a sensitive key passes every layer untouched:

- **Log message text** — `logger.error(f"Token exchange failed: {token_data}")` lands in `event["logentry"]["message"]` / `formatted`, which the scrubber never reads.
- **Exception strings** — `raise SpotifyTokenError(f"...{secret}...")` lands in `event["exception"]["values"][i]["value"]`, likewise unread. Note the existing token paths already build messages from `response.text` (`error_msg = response.json().get("error_description", response.text)`), so a provider that echoes a credential back in an error body would surface it here.
- **Breadcrumb `message`** — same shape as logentry.

## Not a current bug

I checked the live paths before filing: the token-exchange and refresh handlers interpolate only the *exception* (`f"Token exchange failed: {e}"`), and the exceptions they raise are built from Spotify's `error_description`/`response.text`, not from the token dict. So nothing in `main` puts a known secret into message text today.

The problem is that **nothing stops the next one**. The natural change — adding `{token_data}` or `{code}` to an error message while debugging — reintroduces the #436 class of leak, silently, with all three of #510's layers still green. A reader at the call site reasonably assumes "the Sentry scrubber" covers text; it does not.

## Suggested direction (not prescriptive)

1. **Pattern-based value scrubbing** over `logentry.message`/`formatted`, `exception.values[].value`, and `breadcrumbs[].message`. Spotify OAuth material is regex-able: authorization codes and tokens are long high-entropy `[A-Za-z0-9_-]` runs; `BQ`-prefixed access tokens and `AQ`-prefixed refresh tokens are recognisable. Redact the matched span, not the whole message, so the diagnostic stays useful.
2. **Or**: a targeted denylist of *known secret values* pulled from config at init (`client_secret` especially, which is a fixed string the process already holds) — cheap, exact, no false positives, but only covers values known at init time.

(1) and (2) compose; (2) is a few lines and closes the highest-value case immediately.

Whichever lands, it needs the same standard #510 held to: assert on the bytes leaving the process, and prove the test fails without the fix.

## Meanwhile

#510 documents the boundary in the `_strip_pii` docstring and points here, so the limitation is discoverable at the call site rather than by reading the implementation.

chrisrogers37:
Closes #514.

## The gap

`_strip_pii` redacted **by key** — `_sentry_key_is_sensitive(k)` decided whether a value became `[Filtered]`. A secret *interpolated into a string* sat under no key and passed every layer untouched: `logentry.message`/`formatted`/`params`, `exception.values[].value`, and breadcrumb messages.

As the issue says, this was not a live bug — nothing in `main` put a known secret into message text. The problem was that nothing stopped the next one, and it would have landed with all three of #510's layers still green.

I confirmed #436 is fixed at `main` before touching anything (`include_local_variables=False`, `_redact_stacktrace` walking exception/threads/top-level stacktraces) and did not redo that work.

## The fix

The existing recursive `_redact` walk now applies a **second policy at its string leaves**: `_scrub_text` redacts secret-shaped spans in place, replacing only the matched span so the surrounding diagnostic survives.

Four locators:

| Locator | Catches | Source of truth |
|---|---|---|
| Known secrets | `SPOTIFY_CLIENT_SECRET`, `SECRET_KEY`, `TOKEN_ENCRYPTION_KEY` + fallbacks, and the password inside `SQLALCHEMY_DATABASE_URI`/`REDIS_URL` | registered from config at init |
| `Bearer <value>` | the auth scheme names its own value | — |
| Labelled value | `client_secret=…`, `'access_token': '…'`, `?code=…` | **derived from the same two denylists `_sentry_key_is_sensitive` reads** |
| OAuth shape | `BQ…`/`AQ…` ≥40 chars, Fernet `gAAAAA…` | — |

The one idea worth calling out: the label locator doesn't invent a second policy for text. It *derives* its key pattern from `_SENTRY_PII_DENYLIST` / `_SENTRY_PII_EXACT_KEYS`, so widening one list improves both the key walk and the text walk, and they cannot drift apart.

## Three things the review process changed

**1. Enumerating fields failed open.** My first implementation scrubbed a hardcoded list of the three surfaces the issue named. That left `extra`, `contexts`, `request.query_string` and breadcrumb `data` — all *already walked for their keys* — passing an interpolated `?code=…` straight through. Scrubbing at the leaf of the walk instead is both smaller (deleted 3 of the 4 functions I'd added) and fails closed: a payload field nobody listed is still covered.

**2. The key pattern had to be atomic — this one is load-bearing.** With a plain group the pattern is quadratic: inside one unbroken word-character run, every denylist occurrence opens a candidate key, and the trailing `*` backtracks across the rest of the run when the separator fails. That input is reachable, not theoretical — exception strings here are built from uncapped third-party response bodies (`spotify/http_client.py:296`, `spotify/auth.py:242`) and Sentry serialises to 100,000 chars *before* `before_send` runs:

| input | plain group | atomic |
|---|---|---|
| 100 KB | **48.3 s** | **9.4 ms** |

A regex that can hold a gunicorn worker for tens of seconds inside the error-reporting path would have been worse than the leak it guards.

**3. A non-sensitive key could swallow a sensitive one.** In `rejected: client_secret=abc`, key `rejected` matched, its value span ran to the end, the pair was judged harmless, and the credential was consumed without ever being examined. Only a *sensitive* key may open a match now. This was caught by a test written specifically to isolate that locator — see below.

## Verification

**Full suite: 2164 passed, 0 failed.** `ruff check .` clean. `black --check` clean on all three touched files (each was clean at HEAD; the rest of the tree's pre-existing drift is untouched).

**Egress-level, to #436's standard.** Four new tests in `test_sentry_egress.py` boot the real `_init_sentry` against a local HTTP sink, drive a real `logger.error` / raised exception / breadcrumb, and assert on **the bytes actually POSTed**. All four fail without the fix:

```
FAILED TestFreeTextEgress::test_interpolated_secrets_in_log_message_do_not_egress
FAILED TestFreeTextEgress::test_secret_in_log_params_does_not_egress
FAILED TestFreeTextEgress::test_secret_in_exception_string_does_not_egress
FAILED TestFreeTextEgress::test_secret_in_breadcrumb_message_does_not_egress
```

**Mutation-checked, every locator and every routing line.** Neutralising any one breaks ≥1 test:

```
known-secrets  -> 4 failed    leaf-scrub     -> 14 failed
bearer         -> 1 failed    logentry-route -> 10 failed
labelled       -> 4 failed    exception-val  ->  4 failed
oauth-shape    -> 6 failed    toplevel-msg   ->  1 failed
```

This is worth stating because **the first draft of the suite did not achieve it**: `bearer` and `labelled` were dead — every fake token I'd written was `BQ`/`AQ`-shaped, so the shape locator satisfied all of them and those two locators could have been deleted with the suite still green. The tests that isolate them use secrets that deliberately do *not* match the shape rule. Writing those is what surfaced defect 3 above.

**No over-redaction.** A test pins that playlist IDs (22-char, exactly what the 40-char floor protects), `status_code=429`, `schedule_id` and `tracks_total` all survive. `status_code` is specifically covered because `code` is an exact-match key — the distinction `_SENTRY_PII_EXACT_KEYS` exists for.

**Anti-rot.** A test fails if a secret-shaped attribute is added to `config.py` without being registered or explicitly acknowledged as non-secret. Another asserts the derived regex agrees with `_sentry_key_is_sensitive` on every key — the regex *failing to open* a match is the silent direction no other test would see.

## Stated bound

This governs the **Sentry egress only**. The same interpolated secret still reaches stderr and the platform log store through the ordinary logging handlers, which this hook is not in the path of. Said out loud in the `_strip_pii` docstring rather than left to be discovered. `_scrub_text` is a pure function, so a `logging.Formatter` could close that second sink with the same locators — worth a follow-up issue, out of scope here.

## Not done, deliberately

`_register_sentry_secret_values` **replaces** the registry rather than unioning. Review flagged that two `create_app` calls in one process would disarm the first app's secrets. Production has exactly one; union would complicate the reset that keeps the tests order-independent. Noting it rather than silently choosing.

---

Ran `/simplify` (4 review agents) after implementation — findings 1-3 above came from it. Did not run `/weigh-development-paths`; the mechanism was specified in the issue.

https://claude.ai/code/session_016naBu9zrQNRQaw3CZ9cTfd


chrisrogers37:
## GitGuardian check: false positive on a superseded commit — needs a human call

**Backend lint & tests: pass. Tailwind: pass. GitGuardian: fail.**

### What it found

One finding, [incident 35794337](https://dashboard.gitguardian.com/workspace/712512/incidents/35794337), `Generic High Entropy Secret` in `tests/test_sentry_init.py`:

```python
"message": "exchange rejected: client_secret=7f3a9b2c1d4e5f60aabbcc"
```

**No real credential is involved.** That is a made-up hex string I wrote as a test fixture. There is nothing to revoke or rotate.

The detection is still *correct* — `client_secret=<22 hex chars>` reads exactly like a real leaked secret, which is precisely the pattern this PR teaches the scrubber to catch. Hard to be annoyed at it.

### What I already did

`1be93a9` replaced both such fixtures with obviously-fake, low-entropy values:

| before | after |
|---|---|
| `client_secret=7f3a9b2c1d4e5f60aabbcc` | `client_secret=EXAMPLE-not-a-real-client-secret` |
| `Bearer 7f3a9b2c1d4e5f60aabbccddee` | `Bearer EXAMPLE-not-a-real-bearer-value` |

The entropy was incidental — those two tests only require a value that does **not** match the `BQ`/`AQ` shape rule, so that they isolate the label and Bearer locators. Re-verified after the swap: neutralising either locator still fails its test (`bearer -> 1 failed`, `labelled -> 4 failed`).

### Why it still fails

GitGuardian scans **every commit in the PR**, not the final tree. The value is gone from the head commit but still present in `c8b9a1e`. The check cannot go green while that commit is in the branch.

### The call is yours — two options

1. **Dismiss the incident in the GitGuardian dashboard** as a false positive. Nothing to rotate; the head commit is already clean. Zero risk, leaves the two commits as they are.
2. **Authorize a squash + force-push** of this branch so no commit ever contained the string. Cleaner history and no lingering incident. I did not do this unprompted — force-pushing is a destructive operation I don't take without explicit sign-off, even on my own unmerged branch.

Happy either way; say the word and I'll do (2), or dismiss and merge on (1).

chrisrogers37 (commented):
**Approve**

Mutation-tested the assertions, live-verified against a hand-built Sentry event (not the PR's own fixtures), and reproduced the ReDoS claim empirically. One non-blocking coverage gap found; details below.

## 1. Mutation-check: tests go RED without the fix

Reverted `shuffify/__init__.py` to `origin/main` (PR base == main tip, zero drift), kept the PR's test files:

- `tests/test_sentry_egress.py`: **7 passed, 4 failed** — exactly the 4 named in the PR body (`test_interpolated_secrets_in_log_message_do_not_egress`, `test_secret_in_log_params_does_not_egress`, `test_secret_in_exception_string_does_not_egress`, `test_secret_in_breadcrumb_message_does_not_egress`).
- `tests/test_sentry_init.py`: **collection error** — `ImportError: cannot import name '_SENTRY_SECRET_CONFIG_ATTRS'`. Stronger than RED: the whole 32-test module can't even load against unfixed code.

Restored the fix → both files back to 43/43 green. Also spot-checked a single-locator mutation: neutralized `_SENTRY_BEARER_RE` alone → exactly 1 failure (`test_bearer_value_without_token_shape_is_redacted`), matching the PR's claimed per-locator table precisely.

## 2. Live verification — own event, not the PR's tests

Built a Sentry event from scratch with secrets in `logentry.formatted`/`params`, `exception.values[].value`, and `breadcrumbs[].message`, exercising all four locators (known-secret, Bearer, labelled, OAuth-shape). All redacted correctly. Legitimate content survived alongside: `status_code=429`, `schedule_id`, `tracks_total`, a 22-char playlist ID, and the defect-3 probe `"rejected: client_secret=..."` (secret gone, `"rejected:"` text intact — confirms the non-sensitive-key-swallows-sensitive-pair fix holds).

One gap found in this pass — not currently exploitable, see §4.

## 3. ReDoS claim — reproduced independently

Rebuilt the atomic-vs-plain key pattern standalone and timed it on adversarial input (denylist token repeated, no separator, forcing backtracking):

| input size | atomic (shipped) | plain group (mutated) |
|---|---|---|
| 1,000 chars | 0.1 ms | 5.1 ms |
| 5,000 chars | 0.2 ms | 164 ms |
| 20,000 chars | 0.9 ms | 1,968 ms |
| 50,000 chars | — | 12,006 ms |
| 100,000 chars | 4.7 ms | (extrapolated ~48s) |

Plain-group timings scale quadratically (5x input → ~32x time; 2.5x input → ~6.1x time) and extrapolate to ~48s at 100KB — matching the PR's claimed 48.3s almost exactly. The atomic group is load-bearing, not a decorative comment.

## 4. Residual gap: unshaped, unregistered, unlabelled secret in `logentry.params`

`params` values are only covered when the secret is a known registered value, Spotify-OAuth-shaped, or has an adjacent text label. A secret with none of those three properties, sitting as a bare positional arg (`logger.error("failed for %s", secret)` with no `key=` in the format string), egresses in full — confirmed in both `formatted` and `params`.

The test suite doesn't cover this cell: `test_token_in_logentry_params_is_redacted` uses `REFRESH_TOKEN = "AQD..."` (shape-matched), and the two non-shape tests (`test_labelled_secret_without_token_shape_is_redacted`, `test_bearer_value_without_token_shape_is_redacted`) only exercise `logentry.message`, never `params`. Real, untested empty cell in the shape × position matrix.

**Not currently live**: `config.py` has exactly 4 secret-shaped attributes and all 4 are registered (grep confirms no other API keys/webhook secrets exist to fall through). Every OAuth artifact in this app — access/refresh tokens, auth codes, Fernet ciphertext — is shape-matched. So nothing in the current codebase hits this cell today. But it's a genuine gap in "everything else the walk reaches for free," not a restatement of the already-disclosed stderr/platform-log bound. Worth one follow-up test mirroring `test_labelled_secret_without_token_shape_is_redacted` but targeting `params`, so the bound is explicit rather than implicit — the same discipline the PR already applied to the config-secret registry via the anti-rot test. Not blocking: the three #514-named surfaces and the app's actual secret universe are correctly closed.

## 5. Other scrutiny points

- **Over-redaction**: clean (§2) — diagnostic content survives everywhere I checked.
- **Diff scope** (+707/-16, 4 files): no scope creep. `shuffify/__init__.py` is entirely the fix. Test-file removals are 2 lines total — a docstring update (#436 → #436, #514) and an unrelated-but-justified thread-teardown perf fix (comment explains it saves ~3s across 6 tests). CHANGELOG entry matches the implementation precisely.
- **#436 regression**: `include_local_variables=False` present and unchanged; `_redact_stacktrace` byte-identical to main; both #436 tests still green in the 43.
- **Lint/format**: `ruff check` and `black --check` clean on all 3 touched files.
- **Full suite**: 2185 passed / 1 skipped / 0 failed in my sandbox. PR body claims 2164/0/0 — base commit confirmed identical to `origin/main` tip (zero drift), so the delta is environment/dependency variance, not a stale-branch artifact or a hidden failure.

Exceptional rigor throughout: mutation-tested every locator, reproduced its own ReDoS numbers under independent construction, disclosed its own bounds instead of waiting to be asked. The one gap here is adjacent to that disclosed scope, not a violation of it.

chrisrogers37:
[alex] [ci-triage] [info] GitGuardian failure here is a **real detection of a non-credential**, not a broken integration.

**The integration is healthy.** The check run carries GitGuardian incident `35794337`, workspace `712512`, detector `Generic High Entropy Secret`, and resolves to `tests/test_sentry_init.py` line 364 at commit `c8b9a1e`. Its summary reports a "scan of 2 commits", matching this PR's actual commit count. Control: **20 of the last 20 merged shuffify PRs returned GitGuardian `success` / "No secrets detected"** — the app authenticates and scans fine.

**The 1-second runtime is not evidence of anything.** 17 of those 20 *passing* runs completed in 0-2s. GitGuardian is a server-side GitHub App check, not a CI job that provisions a runner, so elapsed time carries no signal about whether it scanned.

**The flagged value is not a live credential.** Characterised without reproducing it:

- 22 hex characters. A Spotify client secret is 32 hex — wrong length by construction.
- `sha256[:8] = 05739d89`; no match against any of 69 secret-shaped values reachable locally (process env + the checkout's `.env*`).
- Absent from `origin/main`, and absent from this PR's own head tree — already replaced by `1be93a9`.
- Exactly two commits in the whole repo history touch it: the one that added it and the one that removed it, both in this PR.
- It is the fixture for `test_labelled_secret_without_token_shape_is_redacted`, whose assertion is that the value gets scrubbed.

*Coverage bound:* compared against locally-reachable values only, not against Railway's production env — I did not pull the live secret. The length mismatch and the two-commit provenance are what make that comparison unnecessary.

**Why it stays red though head is already clean.** GitGuardian scans every commit in a PR, not the final tree, so `1be93a9` cannot clear a finding anchored to `c8b9a1e`. There is no ggshield workflow in `.github/` — this is the GitHub App, configured server-side, so no repo-side file changes it.

**Clearing it needs a human, two ways:** dismiss incident `35794337` in the GitGuardian dashboard as a false positive (fastest, and correct), or rewrite history to drop the original commit — which needs a force-push and explicit sign-off. Not doing either unasked.

**Base currency:** merge-base `2eac436`, 2 commits behind main. Two-dot shows `requirements/base.txt` and `requirements/dev.txt` as differing — files this PR never touches; that is main's dependabot bumps (#542/#543) showing through the stale base, not a revert. Three-dot confirms the real surface is 4 files. Rebase is tidy, not urgent.

chrisrogers37:
**Verdict: Ship it** (comment-only — same-identity blocks Approve; `gh pr review --approve` confirmed refused: "Can not approve your own pull request")

Reviewed against the specific threat model, not the test suite. Cloned at PR head, built a real venv (`requirements/dev.txt`), ran the actual test suite myself, then wrote my own adversarial cases importing the real `_strip_pii`/`_scrub_text`/`_register_sentry_secret_values` — none of this is from `test_sentry_egress.py`/`test_sentry_init.py`.

**Baseline:** 43/43 passed in my own venv (not the CI badge).

**Mutation-tested 2 of the 4 locators myself** (didn't just trust the PR's table):
- Neutralized the OAuth-shape line (`_SENTRY_OAUTH_TOKEN_RE.sub`) → 7 real failures. Restored, re-ran clean (43/43), diffed byte-identical to pristine.
- Neutralized the known-secrets loop → 4 real failures. Same clean restore.
- Not decoys.

**The two things you asked me to check specifically:**

1. **Frame locals (#436's vector).** The PR frames this as out of scope ("#436 fixed, did not redo"). I checked whether that's actually true rather than taking the framing at face value: `_redact_stacktrace` calls `_redact(frame["vars"])`, and `_redact`'s new string-leaf branch means every frame-local *string value* now also gets `_scrub_text`'d — including a value sitting under a non-sensitive key. I built this exact case:
   ```python
   "vars": {"debug_note": f"about to retry with {ACCESS_TOKEN}", "attempt": "2"}
   ```
   → `debug_note` comes back `"about to retry with [Filtered]"`, `attempt` untouched. Same result for `event["threads"][...]["stacktrace"]` (not just `event["exception"]`). So the #436 concern isn't just "not regressed" — it's incidentally *hardened* by the leaf-based design, beyond what the PR claims for itself.

2. **Connection string mid-sentence.** Built a realistic log line with a percent-encoded special-char password (`P%40ss%3Aw0rd%21`) embedded mid-sentence — redacted correctly, host/port/dbname/diagnostic text survived. Then pushed harder: a **raw, unencoded** password containing a literal `@` (`Hunter2@Secret!`, the realistic misconfiguration case — most people don't percent-encode a pasted DB password). `urlsplit().password` still extracts it correctly (rpartition on the *last* `@` in netloc handles this). Both redact cleanly.

**One real gap I found that the suite doesn't cover:** a Spotify-shaped token (`BQ`/`AQ` + ≥38 chars) with a **literal embedded newline** mid-token defeats the shape locator entirely — both fragments fall under the 38-char floor independently, so the full token survives untouched in both halves. Confirmed empirically. This is *not* a decoy-test problem, it's an uncovered case.

I checked whether it's reachable today before flagging it as more than theoretical: grepped for `textwrap`/`pprint`/`.wrap(`/`yaml.dump` anywhere in `shuffify/` (zero hits) and read every token-adjacent log/raise site in `auth.py`/`http_client.py` — all of them are plain `f"...{e}"` / `f"...{error_msg}"` interpolations, nothing that would fold a token across a line break. So: real gap, zero current trigger — the same shape as #514 itself ("not a current bug... nothing stops the next one"). I'd treat it the same way #514 was treated: worth a tracked follow-up or a line added to the "Known bound" docstring paragraph (you're already documenting the stderr/log-store bound there), not a blocker on this PR.

**Also checked and confirmed, not just accepted:**
- Registered-secret redaction survives an embedded newline fine (exact `str.replace`, unaffected) — only the *shape*-based locator has the gap above.
- Atomic-group ReDoS claim: ran `_scrub_text` on the actual 100KB adversarial payload (`"email" * 20000`) against your real shipped pattern — **9.5ms**, matching your claimed 9.4ms almost exactly. Not re-verified on prod hardware, but the pattern is doing what you say.
- Python version risk: `(?>...)` atomic groups need 3.11+. Dockerfile pins `3.12-slim`, both CI workflows pin `3.12` — no CI/prod mismatch that would make this crash on deploy.
- File list is `CHANGELOG.md`, `shuffify/__init__.py`, `tests/test_sentry_egress.py`, `tests/test_sentry_init.py` — nothing under `migrations/versions/`, so the standing schema-freeze note doesn't apply here regardless of its own (now week-old) staleness.
- Over-redaction sanity: a long non-BQ/AQ-prefixed hex identifier survives untouched.

No merge from me — Chris merges.

Issue #531, opened 2026-08-01
[deploy] Migrations must run as a DigitalOcean PRE_DEPLOY job — run_command bypasses ENTRYPOINT, and the pending migration needs amending before it runs


chrisrogers37:
## Summary

`scripts/docker-entrypoint.sh` has never run in production. DigitalOcean App Platform's `run_command` **replaces** the container entrypoint, so the migration step added by #503 was bypassed on every deploy — confirmed by the deploy log, whose first line is `Starting gunicorn` with no entrypoint output at all.

The consequence surfaced today as a hard outage:

```
shuffify.SchemaOutOfDateError: Database schema is not at Alembic head:
current=c4d5e6f7a8b9, head=e6f7a8b9c0d1
```

**#503 did not cause this.** Production had been serving against a stale schema silently; #503 made a pre-existing condition loud. The schema gate did exactly what it was built to do.

This issue proposes where migrations should actually run on this platform, and flags a second problem that the fix will trigger.

---

## Part 1 — Migrations belong in a PRE_DEPLOY job

App Platform supports a `PRE_DEPLOY` job: it runs to completion **before** the new deploy's services start, and a non-zero exit aborts the deploy without cutting traffic over. That is the same contract the entrypoint was written for, expressed in the platform's own vocabulary.

### Proposed app-spec addition

```yaml
jobs:
  - name: db-migrate
    kind: PRE_DEPLOY
    # Same source and build as the web service, so the job runs the exact
    # migration chain that ships in the image being deployed.
    github:
      repo: chrisrogers37/shuffify
      branch: main
      deploy_on_push: true
    dockerfile_path: Dockerfile
    instance_count: 1
    instance_size_slug: <same slug as the web service>
    run_command: flask db upgrade
    envs:
      # The database this deploy is migrating.
      - key: DATABASE_URL
        scope: RUN_TIME
        value: ${db.DATABASE_URL}

      # `flask db upgrade` imports run.py and builds the whole application,
      # so the job needs everything create_app() requires at import time.
      - key: APP_CONFIG
        scope: RUN_TIME
        value: production
      - key: SECRET_KEY
        scope: RUN_TIME
        value: ${APP_SECRET_KEY}
      - key: SPOTIFY_CLIENT_ID
        scope: RUN_TIME
        value: ${SPOTIFY_CLIENT_ID}
      - key: SPOTIFY_CLIENT_SECRET
        scope: RUN_TIME
[...]

PR #549, opened 2026-08-08
docs(migrations): correct the claim that migrations run automatically in production (#531)

chrisrogers37:
Closes the documentation half of #531. **Docs and comments only — no behavior change, and the `PRE_DEPLOY` job is deliberately not implemented.**

## The defect

`CLAUDE.md`, the production setup guide, the `Dockerfile`, `config.py` and the app factory all stated that production applies Alembic migrations on startup via `scripts/docker-entrypoint.sh`.

It does not, and never has. DigitalOcean App Platform's `run_command` **replaces** the container `ENTRYPOINT`, so the entrypoint has never executed in production — confirmed by a deploy log whose first line is `Starting gunicorn`, with no entrypoint output at all.

That is why production served a stale schema silently for weeks. **A wrong doc is worse than a missing one, because it stops people looking.**

## I swept the claim, not the files it was reported in

Three files were reported. I grepped for the *claim* and found it in **six files across 23 sites**.

Searched (case-insensitive, whole tracked tree): `migrations run/runs/are run/applied/happen`, `runs alembic`, `alembic upgrade`, `flask db upgrade`, `docker-entrypoint`, `ENTRYPOINT`, `entrypoint`, `on startup`, `automatic(ally) migrat`, `migrate on startup`, `MIGRATE_ON_STARTUP`, `handles migrations automatically`, `no manual migration step`.

| File | Sites | What it claimed |
|---|---:|---|
| `documentation/production-database-setup.md` | 9 | "Runs Alembic migrations on startup", "No manual migration step is needed", "the app handles migrations automatically" |
| `shuffify/__init__.py` | 6 | docstrings + **the runtime error message** operators see during this exact failure |
| `CLAUDE.md` | 3 | the migrations table's Production row and the two env-var paragraphs under it |
| `config.py` | 2 | `MIGRATE_ON_STARTUP` comments on both base and prod config |
| `scripts/docker-entrypoint.sh` | 2 | header banner + "Migrations run here" |
| `Dockerfile` | 1 | see below |

**The Dockerfile comment was the worst of them**, and it was not in the reported list. It did not merely omit the failure mode — it explicitly ruled it out:

> ENTRYPOINT rather than a longer CMD so the migration step survives a platform-level run-command override — those replace CMD, leaving ENTRYPOINT to wrap whatever they substitute.

`run_command` replaces `ENTRYPOINT` too. That comment actively reassured every reader that the thing which happened could not happen.

## Two places I deviated from #531, deliberately

**1. #531's suggested wording would have re-introduced the same defect.** It asks the entrypoint comment to say "production migrations run via the `db-migrate` PRE_DEPLOY job". That job does not exist yet, so writing it would swap one false mechanism claim for another — the exact failure this PR exists to correct. Everything here says the job is **proposed, not implemented**, and links #531.

**2. The setup guide states the real state rather than just losing the wrong one.** It is a setup guide, so a reader follows it to stand up a new deployment; deleting "migrations run automatically" would leave them told nothing. The Migrations section now says migrations are **manual on DigitalOcean today**, gives the exact command with the two env vars that are easy to get wrong (`SHUFFIFY_MIGRATION_STEP=1`, `SCHEDULER_ENABLED=false`), and keeps a separate section for the platforms where the entrypoint genuinely does run.

## Deliberately not touched

- **The `PRE_DEPLOY` job.** Deploy-config change, real blast radius, not ours to make.
- **`documentation/planning/system-review_2026-07/*`** (2 hits) — historical planning records proposing the entrypoint approach. They describe a past decision, not current state; rewriting them would falsify the record.
- **`documentation/archive/**`** (2 hits) — same reasoning, explicitly archived.
- **A dev-password literal at `documentation/production-database-setup.md:126`.** Unrelated to this claim, so not bundled. It is a local-Docker placeholder (`db:5432`, compose-internal), not a live credential — flagged separately.

## Verification

| Gate | Result |
|---|---|
| `pytest` | **2160 passed, 1 skipped**, rc=0 |
| `ruff check .` (blocking) | **All checks passed**, rc=0 — ruff 0.16.1, the pinned version |
| `scripts/check-migration-chain.py` (blocking) | rc=0 — 22 migrations, single head, no duplicates |
| `sh -n scripts/docker-entrypoint.sh` | rc=0 |
| `black --check shuffify/` (non-blocking) | 54 files — **identical on `main` and on this branch**, so entirely pre-existing drift; this PR adds none |

The entrypoint script is executed directly by `tests/test_db_schema_guard.py`; those 5 tests pass, so the added banner does not disturb its contract.

Branched off `6710b49` (current `main`); two-dot and three-dot diffs are identical, so there is no stale-base cargo.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -596,7 +596,7 @@ Two config attributes decide it — `TESTING` and `MIGRATE_ON_STARTUP`:
 |-------------|-----------|----------------------|----------|
 | Tests | `True` | — | Tables built directly from the models via `db.create_all()`; no migration chain |
 | Development | `True` | `True` | Also `db.create_all()` today — `TESTING=True` short-circuits before the migration branch is reached (#325) |
-| Production | `False` | `False` | `scripts/docker-entrypoint.sh` runs `flask db upgrade` **before** Gunicorn starts; the app factory then only *verifies* the schema is at head and refuses to start if it is not |
+| Production | `False` | `False` | **Nothing applies migrations automatically** — `scripts/docker-entrypoint.sh` is bypassed on DigitalOcean App Platform (see the warning below). The app factory only *verifies* the schema is at head and refuses to start if it is not |
 
 Development does **not** currently exercise the Alembic chain: `DevConfig`
 sets `TESTING = True`, which routes it to `db.create_all()`. That is the bug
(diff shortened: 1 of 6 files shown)


chrisrogers37 (commented):

Ran both of ari's checks directly against the code, not against the PR's own claims about the code.

**1. Docs-only, verified — not just asserted.** Read every hunk in `config.py` and `shuffify/__init__.py` individually: every changed line in both files is a `#` comment, a docstring line, or prose inside the `SchemaOutOfDateError` message string. Zero control-flow changes, zero constant-value changes (`MIGRATE_ON_STARTUP`, `SCHEMA_DRIFT_OVERRIDE_VAR`, `MIGRATION_STEP_VAR` are all untouched context lines). The one edit that looked riskiest — the error-message string content — still needed checking for a `%s`-count/args mismatch, since that class of edit can silently break at runtime while reading as "just wording": counted both, 3 placeholders ↔ 3-tuple, matches on both sides. Also checked `Dockerfile` and `docker-entrypoint.sh` the same way (not explicitly asked, but they're source too) — same result, comment-only.

**2. Held the new wording to the same test alex applied to the issue's suggested wording.** Went through all 6 files asking "true today, or intent not labeled as such":
- "Refuses to boot if schema isn't at head" — traced the actual call path: `_verify_schema_at_head` raises `SchemaOutOfDateError`, and `_init_database`'s except block explicitly re-raises it (`except SchemaOutOfDateError: raise`, with a comment saying why it's never absorbed into the generic handler below it). Propagates out of `create_app()` uncaught. The claim is true in the code, not just in the docstring describing the code.
- "Correct for `docker run`, docker-compose" (new claim, not in the original wording) — checked `docker-compose.yml`, `.override.yml`, `.prod.yml` for an `entrypoint:`/`command:` override. None exists, so the Dockerfile `ENTRYPOINT` is honored as claimed. This is exactly the kind of new specific claim that could have quietly repeated the old mistake, and it holds up.
- The manual-migration command block (`SHUFFIFY_MIGRATION_STEP=1 SCHEDULER_ENABLED=false flask db upgrade`) matches the entrypoint script's actual invocation, env-var-for-env-var.
- Every mention of the `PRE_DEPLOY`/`db-migrate` job is explicitly marked proposed/not-implemented, every time it comes up, across all three files that mention it (CLAUDE.md, the setup guide, the entrypoint banner) — never slipped in as current mechanism.

**3. The Dockerfile "worst site."** Old comment: ENTRYPOINT "survives a platform-level run-command override... those replace CMD" — an unconditional, confident, wrong claim, the exact shape of the original defect. New comment: "on any platform that honours ENTRYPOINT," scoped rather than universal, and its one concrete sub-claim (docker-compose) is independently verified above. It doesn't repeat the pattern.

**One non-blocking precision note.** "A migration freeze is in effect" appears twice (CLAUDE.md, entrypoint banner) as a flat present-tense claim. Unlike everything else here, this isn't a code-mechanism claim I can verify true/false against the repo — nothing in CI or branch protection actually blocks a new Alembic revision from merging (`check-migration-chain.py` checks for a single head, not for zero new revisions). It reads more like a norm being declared than an enforced fact, which is a different category from what this PR is fixing, and it's immediately followed by the actual manual-apply steps, so no reader is left not knowing what to do. Worth being aware of, not worth blocking on.

Everything else checked out clean, including the parts of the diff nobody asked me to check. Good PR.

Issue #547, opened 2026-08-08
Authenticated user gets a 500 when Spotify errors: workshop.html user.id is not null-safe against the None require_auth_page deliberately allows


chrisrogers37:
## The bug

`workshop.html:1195` renders `{{ user.id | tojson }}` (template `user` = `spotify_profile`)
and is **not null-safe**.

`require_auth_page` deliberately swallows `SpotifyError` from the profile fetch and continues
with `spotify_profile=None` — its own comment says a Spotify hiccup should not bounce an
authenticated user to login. That is the right call. **This template line is the one place that
does not tolerate the `None` it produces.**

## Reproduced, not inferred

Found by `navi` while reviewing #512, and confirmed empirically: mock `AuthService.get_user_data`
to raise `SpotifyError` on an otherwise fully valid authenticated request to `/workshop/<id>`
(valid session, valid DB user, valid playlist) →

```
TypeError: Object of type Undefined is not JSON serializable
```

full traceback rooted at that exact template line. In production this reaches the generic 500
handler (`errors/500.html`), not a raw crash.

## This is not new, and #512 improved it

Important framing, so this is not misread as a regression introduced by that PR.

**Pre-#512**, the identical trigger hit `handle_spotify_error`, which returns
`json_error_response(...)` **unconditionally, regardless of request type** — raw JSON rendered in
a browser. That is precisely the bug class #512 exists to eliminate, reached through a different
trigger than the one it targeted.

So #512 **upgrades** this failure from *raw JSON in the browser* to *a proper error page*,
without fully closing it. The remaining hole is that an authenticated user sees a 500 at all,
when the decorator's whole intent was to let them through a transient Spotify failure.

## Fix

One line. Guard the template expression, or pass it explicitly from the handler:

```python
spotify_profile.get('id') if spotify_profile else None
```

matching the null-guard **already used one line above** at the `logger.info` call — so the
correct pattern is present in the same file and simply was not applied here.

PR #550, opened 2026-08-08
perf(routes): consume the user require_auth_page injects in workshop() (#548)

### LINKED ISSUE #548: workshop() ignores the user kwarg require_auth_page injects and re-queries the DB — the one outlier after the #512 consolidation
## The nit

`workshop.py:79` — `workshop()` never uses the `user` kwarg that `require_auth_page` now
injects, and calls `get_db_user()` again instead. That is a **second DB query per page load**,
and it is unmemoized: `get_db_user()` at `routes/__init__.py:251` has no request-level cache
(checked, not assumed).

## Why it is worth a line of code

Not a bug, and not urgent — but `workshop()` is the **only outlier**. `settings()` and
`schedules()` both correctly consume the injected `user` after the same migration. So this is a
one-file inconsistency in freshly-consolidated code, and the cheapest moment to fix it is now,
while the pattern is still obvious. Left alone it becomes the example the next handler copies.

Found by `navi` reviewing #512. Auth coverage there was verified clean and this does not affect
it — 83 handlers, 5 changed, all net-positive, independently re-derived by AST sweep.

Related: #547 (the null-safety bug in the same handler's template).

chrisrogers37:
## What

`workshop()` declared the `user` keyword argument `require_auth_page` injects, ignored it, and called `get_db_user()` for the same row. `get_db_user()` (`routes/__init__.py:251`) has no request-level cache, so every `/workshop/<playlist_id>` load ran the lookup twice.

The handler now consumes the injected row. Closes #548.

## The guards went with it

The two `if db_user:` checks are removed rather than repointed. `require_auth_page` redirects to the login page when the row is missing (`routes/__init__.py:230-233`), so past the decorator the value is never `None`. A null check on it is dead code that reads as though the handler still derives its own user — which is the pattern the issue says the next handler will copy. Repointing the guard at `user` would have kept that reading alive.

This is the one judgment call in the diff beyond the literal ask; flagging it explicitly rather than folding it in silently.

## How many handlers were checked: 76

The issue names `workshop()` as the only outlier, verified against `settings()` and `schedules()`. That check was not exhaustive, so it was redone across the whole route surface.

A grep for `get_db_user` would not have settled it. Both decorators inject **three** kwargs between them, and re-deriving `api` or `spotify_profile` is the same defect — a pattern match on one call name is not a property check on the contract.

The sweep walks the AST of every function in `shuffify/routes/`, pairs each auth decorator with the kwargs it injects, and flags any handler that both declares an injected kwarg and calls the thing that derives it:

| Guard | Handlers | Re-deriving |
|---|---|---|
| `require_auth_page` | 5 | **1** (`workshop.py:68`) |
| `require_auth_and_db` | 71 | 0 |
| **Total** | **76** | **1** |

**`workshop()` was the only one, across all 76 — not just the 5 that changed in #512.** Nothing else to fix alongside it.

(`workshop_hub` declares all three kwargs and uses none. Not the same defect and not touched: the decorator injects them unconditionally, so the signature has to accept them.)

## Regression guard

The sweep is now a contract test rather than a one-off, added to `tests/test_page_auth_contract.py` — which already exists to catch route handlers *that do not exist yet*, the exact concern the issue raises. It covers `api`, `user` and `spotify_profile` across both decorators.

**Mutation-tested, not assumed.** Reverting `workshop.py` to `main` with the test in place fails RED, naming the handler and the call:

```
FAILED tests/test_page_auth_contract.py::test_guarded_handler_does_not_rederive_injected_kwarg[workshop.py-workshop-require_auth_page--]
AssertionError: workshop.py:workshop declares ['user'] -- which @require_auth_page
already injects -- but calls ['get_db_user'] to derive it again.
```

The other 75 handlers pass in that same run, so it is discriminating rather than uniformly red.

## Verification

Run locally on this branch, unsandboxed, Python 3.11:

- `ruff check shuffify/ tests/test_page_auth_contract.py` → **All checks passed**
- `pytest tests/ -m "not integration"` → **rc=0, 2258 passed, 1 skipped**
- Baseline on `main` before the change → **rc=0, 2181 passed, 1 skipped**
- Regression diff of named failures, baseline vs branch → **empty**
- Test delta **+77** = 76 parametrized handlers + 1 collector-empty guard, which reconciles exactly with the table above

`/simplify` was run (reviewed inline across its four angles rather than fanned out to agents, per this session's config). It found one real item, applied: the new sweep duplicated the routes-dir AST walk `_route_handlers` already did, and derived decorator names through two nested set comprehensions with a silent `getattr(..., "")` fallback. Both now go through shared `_iter_functions` / `_decorator_names` helpers. Mutation test re-run after that refactor to confirm the teeth survived it. Skipped as not worth the state: memoizing the AST walk — it is collection-time only, and the file already re-walked 3x before this change.

`/weigh-development-paths` was not run — no fork worth weighing. The codebase voted on the one stylistic choice: `load_schedule_context(...)` keeps the parenthesized call form `schedules.py:87` already uses for the identical call.

## Coordination

`virgil` is on #547 in the same handler's template (`workshop.html:1195`). No branch for it was pushed at the time of writing, and this PR does not touch `workshop.html`. The overlap risk is `workshop.py:141-148` if that fix passes a safe default from the handler — this diff ends at line 115 and leaves the `render_template` call untouched, so the two should merge cleanly. Worth a rebase check if #547 lands first.

Do not merge — routing to `ari`.


chrisrogers37 (commented):

Re-did every check from scratch against current `main` rather than trusting my own #512 memory or this PR's report of itself.

**1. `require_auth_page`, re-read fresh (not recalled) — unchanged since #512.** Byte-for-byte identical to what I reviewed then; #549 (docs-only) didn't touch it. Confirmed again: exactly three exits before `f(*args, **kwargs)` — unauthenticated, DB-down, no-db-row — all three `redirect(...)`, all three happen *before* `user = get_db_user()` is assigned for real. No fourth path. The later `except SpotifyError: spotify_profile = None` only ever touches `spotify_profile` — it cannot un-set `user`. So: `user` is unconditionally non-None by the time any wrapped handler body runs; `spotify_profile` is not. That asymmetry is the whole ballgame here.

**2. The diff touches only `user`/`db_user`, never `spotify_profile`.** Read the full `workshop.py` hunk: both removed guards (`if db_user:` around `load_schedule_context` and around the preference lookup) gate on `db_user`/`user`, now replaced with the injected `user` directly. `spotify_profile` does not appear anywhere in this diff. `workshop.html` is not in the changed-files list at all — no overlap with #547's territory.

**3. The 76-handler sweep — cross-checked with two independent methods, not by re-running this PR's own collector.**
- Re-ran my own #512 AST script (written before this PR existed) fresh against current `main`: 83 route handlers, 6 render HTML (5 `require_auth_page` + 1 hand-rolled `index`), 71 `require_auth_and_db`, 7 undecorated. `5 + 71 = 76` — matches the PR's table exactly, from a tool that shares no code with it.
- Separately grepped every call site of all three deriving functions (`get_db_user(`, bare `require_auth(`, `get_user_data(`) across `shuffify/routes/*.py` on `main`. Outside their own definitions in `__init__.py`, the only hits are `core.py:65/79` and `core.py:360` (both inside `index()`/`logout()` — confirmed undecorated, hand-rolled, correctly out of scope) and exactly one guarded-handler hit: `workshop.py:79`. Zero re-derivation of `api` or `spotify_profile` anywhere. Two different methods, same answer: `workshop()` is the only one.

**4. Mutation-tested myself.** Reverted `workshop.py` to the pre-PR `main` version in a worktree and reran `tests/test_page_auth_contract.py`: exactly one failure, `test_guarded_handler_does_not_rederive_injected_kwarg[workshop.py-workshop-require_auth_page--]`, message byte-for-byte matching what's quoted in the PR body, 93 other cases + baseline still green. Discriminating, not decoy. Ran the full file against the actual PR head too: 94 passed, 1 skipped — the claimed 17 pre-existing + 77 new (76 parametrized + 1 collector-guard) reconciles exactly.

The one thing I'd keep an eye on, not a blocker: the `REDERIVES` mapping only catches re-derivation by exact call name (`get_db_user`, `require_auth`, `get_user_data`). A handler that re-derived, say, `user` via a direct `User.query.get(...)` instead of `get_db_user()` would slip past this contract test the same way a grep would — it's a real improvement over a grep, but it's still a name-based pattern match, not a proof of the property. Not a reason to hold this PR; just noting it since it's the general shape of thing that's bitten this repo before.

Clean, well-scoped fix. Evidence checks out on every point.

PR #551, opened 2026-08-08
fix(workshop): null-safe user.id against the None require_auth_page allows

### LINKED ISSUE #547: Authenticated user gets a 500 when Spotify errors: workshop.html user.id is not null-safe against the None require_auth_page deliberately allows

chrisrogers37:
## What

`workshop.html` rendered `{{ user.id | tojson }}` (template `user` = `spotify_profile`) with no
null-guard. `require_auth_page` swallows `SpotifyError` from the profile fetch and continues with
`spotify_profile=None` deliberately — its own comment says a Spotify hiccup should not bounce an
authenticated user back to login, which is the right call. This line was the one place that did not
tolerate the `None` it produces, so a fully authenticated user with a valid session and a valid
playlist got a generic 500.

The expression now carries the same `(x or {}).get(...)` guard the handler already applies one line
above, at its `logger.info` call (`shuffify/routes/workshop.py:136`).

```diff
-    currentUserId: {{ user.id | tojson }},
+    {# spotify_profile is None whenever the profile fetch fails -- require_auth_page allows that by design. #}
+    currentUserId: {{ (user or {}).get('id') | tojson }},
```

Closes #547.

## Two values, opposite correct treatments

`require_auth_page` injects three kwargs, and two of them have **opposite** null contracts. This PR
adds a guard; #550, landing beside it, removes guards. Both are correct, and the reason belongs in
writing or the log reads as a contradiction later.

| Injected value | `None` past the decorator? | Correct treatment | Where the decorator decides |
|---|---|---|---|
| `user` (the DB row) | **No.** Every failure path returns before the handler runs — no API client, DB unavailable, no user row | Guarding it is dead code. Remove it | `routes/__init__.py:217-233` — three `redirect(...)` returns |
| `spotify_profile` | **Yes, deliberately.** `SpotifyError` from the profile fetch is swallowed and execution continues | Must be guarded at every use | `routes/__init__.py:235-241` — `except SpotifyError: spotify_profile = None`, whose comment says a Spotify hiccup should not bounce an authenticated user back to login |

So in one request path a guard is being removed and a guard is being added, and neither contradicts
the other. The rule is not "guard everything" or "guard nothing" — it is **guard exactly what the
decorator chose to let through**. `api` and `user` are authorization signals and the decorator
refuses the request without them; `spotify_profile` is presentation data and the decorator proceeds
without it on purpose.

The predictable failure this note exists to prevent: someone reads #550 as a general rule, strips
the guard this PR adds "for consistency", and reopens #547.

## The sweep — line 1195 is the only one

The point of `require_auth_page` tolerating a `None` profile is that a Spotify outage degrades
gracefully, and one unguarded reference anywhere on the page defeats that as completely as this one
did. So the whole template tree was swept for the same shape, not just the reported line.

**What I searched.** Not a grep for `user.` alone — the profile could reach a template under any
name, so I resolved the name first:

1. AST-enumerated every `render_template` call in `shuffify/` with its kwargs. The profile reaches a
   template only as `user=spotify_profile`, and only from three call sites (`workshop.py:141`,
   `settings.py:64`, `schedules.py:91`).
2. `grep -rnE "\buser\b" shuffify/templates/` — the full tree including `partials/` and `macros/`, so
   includes and macro call sites are covered. Twelve hits total after dropping `user-select`,
   `user_*` and the `/api/user-playlists` fetch URLs.
3. Checked both context-processor registrations (`shuffify/__init__.py:766`,
   `shuffify/routes/__init__.py:48`) in case one injected the profile globally under a second name.
   They inject `csp_nonce` and `current_year`. Nothing else.

**What would have invalidated that method, checked separately.** The enumeration in step 1 is only
complete if `render_template` is the sole render door, and the sweep in step 2 is only complete if a
template cannot reach the profile without a kwarg. Both were verified rather than assumed:

- No `render_template_string`, `stream_template` or `Template.from_string` anywhere in `shuffify/` —
  so step 1 enumerated every render, not merely many.
- No template reads the profile from `session` or `g`. Templates do read `session`, six times, but
  only for `spotify_token` (`base.html:131,151`, `index.html:21`) and `external_playlist_history`
  (`workshop.html:410,414`). This one matters because the word-boundary pattern in step 2 would not
  have matched a `session.user_data`-style binding.
- No `jinja_env.globals`, `add_template_global` or template filter registrations.

**Result — one hit, and it is the reported one:**

| Template | Profile refs | Verdict |
|---|---|---|
| `workshop.html` | 1 (line 1195) | **the bug** — fixed here |
| `settings.html` | 0 | receives `user=spotify_profile`, never dereferences it |
| `schedules.html` | 0 | receives `user=spotify_profile`, never dereferences it |
| `activity.html` | 0 | not passed a profile at all |
| `dashboard.html` | 10 | different provenance — see below |
| `base.html`, `index.html`, `partials/`, `macros/`, `errors/` | 0 | — |

`dashboard.html` is the one that looks like a second hit and is not. It is rendered from
`core.py:109`, inside `index()`, which is **not** on `require_auth_page` — its `user` comes from a
direct `AuthService.get_user_data(api)` call that bails to the error path rather than continuing with
`None`. Its ten references are also already guarded (`{% if user.images %}`, `{% if user.display_name %}`,
`{{ user.display_name or 'User avatar' }}`) and none reaches `tojson`.

Two things worth recording from the sweep:

- **`tojson` is what makes this fatal, not the dereference.** Flask uses Jinja's default `Undefined`,
  so `{{ user.display_name }}` on a `None` profile renders empty — degradation, not a crash. Only a
  filter that raises on `Undefined` turns it into a 500. That is why the ten guarded `dashboard.html`
  references are not latent 500s, and it is the property to check if a new reference is added.
- **`workshop_hub` (`/workshop`) renders `workshop.html` with no `user` kwarg at all**, which would
  make `user` `Undefined` unconditionally. It does not crash because line 1195 sits inside
  `{% if playlist %}` (opened at line 1176) and the hub passes `playlist=None`. Stated because it is
  a real second path to the same line, currently closed only by that enclosing conditional.

## Tests

Two tests in `tests/routes/test_workshop_routes.py::TestWorkshopPage`, mocking
`AuthService.get_user_data` to raise `SpotifyError` on an otherwise fully valid authenticated
request:

- `test_workshop_renders_when_spotify_profile_is_unavailable` — asserts 200 with the page rendered,
  and `currentUserId: null`.
- `test_workshop_emits_the_real_user_id_when_profile_is_available` — asserts the guard does not cost
  the real id.

**Mutation-checked in both directions**, because the first test alone is satisfied by deleting the
line or hardcoding `null`:

| Template line under test | profile-missing test | profile-present test |
|---|---|---|
| `{{ user.id \| tojson }}` (reverted) | **FAILED** — `TypeError: Object of type Undefined is not JSON serializable` | PASSED |
| `{{ None \| tojson }}` (hardcoded null) | PASSED | **FAILED** |
| `{{ (user or {}).get('id') \| tojson }}` (this PR) | PASSED | PASSED |

Each mutation kills exactly one test, so neither can be satisfied by the shortcut that satisfies the
other. The reverted-fix run reproduces the exact error from the issue independently.

`currentUserId` degrading to `null` is safe at its one consumer, `workshop.html:2584`:
`pl.owner_id && pl.owner_id !== workshopState.currentUserId` reads it as "not yours", which is the
conservative direction.

## Verification

Branched off fresh `main` at `437882d` (includes #512 and #544).

- **Full suite, before/after against a clean `origin/main` worktree:**
  baseline `rc=0`, **2181 passed**, 1 skipped, 0 failed → after `rc=0`, **2183 passed**, 1 skipped,
  0 failed. No new failures; the delta is exactly the two tests added here.

  The first draft of these tests **did** break a third one, which is worth recording because the
  comparison is the only thing that caught it. I copied the mock style from the two tests already in
  `test_workshop_routes.py`, and those two are precisely the ones `test_mock_spec_contract.py` has
  grandfathered — so the new tests took the file from its allowed 2 unspec'd `AuthService` stand-ins
  to 4 and tripped the #521 ratchet. Fixed as that ratchet asks rather than by raising its allowance:
  `autospec=True` on both new patches, which also makes these tests fail on a rename of
  `get_user_data` / `get_authenticated_api` / `validate_session_token`, which is what the guard is
  for. The mutation matrix above was re-run afterwards, since editing test decorators can retire the
  discriminator; it still separates in both directions.
- `ruff check .` → **All checks passed!** (CI runs the whole tree, not `shuffify/` — `CLAUDE.md`
  states the narrower command.)
- `./scripts/build-css.sh --check` → **OK, committed stylesheet matches the current sources.** Run
  because this PR edits a template and the compiled stylesheet is committed.
- `black --check shuffify/` → 54 files would be reformatted. Pre-existing drift, `continue-on-error`
  in CI, and untouched by this PR — the only `shuffify/` file changed here is a `.html`.

## Note for a follow-up, not fixed here

`CLAUDE.md` documents the lint gate as `ruff check shuffify/`; `.github/workflows/ci.yml:59` runs
`ruff check .`, and the workflow's own comment explains that widening was deliberate. Anyone
following the documented pre-push command lints less than CI does. Out of scope for this fix — happy
to open an issue.

https://claude.ai/code/session_01VDF644wejEcjXruxVehzoH


chrisrogers37:

Verified against the real trigger, not the PR's description of it. Cloned at PR head, real venv, ran the actual tests, then mutation-tested independently.

**1. The core fix, under the real trigger.** Ran both new tests directly: `test_workshop_renders_when_spotify_profile_is_unavailable` and `test_workshop_emits_the_real_user_id_when_profile_is_available`, both PASS. Confirmed they use the trigger that matters — `mock_auth_svc.get_user_data.side_effect = SpotifyError(...)` on an otherwise fully valid authenticated request — not a hand-passed `None` into the template context, which would exercise less.

**Mutation-tested myself:** reverted the two-line fix to bare `{{ user.id | tojson }}`, re-ran under the same real trigger:
```
FAILED ...test_workshop_renders_when_spotify_profile_is_unavailable - TypeError: Object of type Undefined is not JSON serializable
```
Exact reproduction of the issue's traceback. The profile-present test stayed green (correctly — that mutation only breaks the None case). Restored, diffed byte-identical to pristine. Not a decoy.

**2. The sweep — audited, not just noted as present.** It's documented (methodology + an explicit negative: "one hit, and it is the reported one"), which is the bar. I independently re-derived the classification rather than trust the table: grepped `\buser\b` across the whole template tree myself and got 11 real references outside settings.html/schedules.html (which receive the profile but never touch it) — 10 in `dashboard.html`, 1 in `workshop.html` (the fix). Matches the PR's table exactly.

Chased two apparent discrepancies before trusting my own count — both were my own grep's fault (`-o` stripped line context my exclude-filter needed, so `/api/user-playlists` fetch URLs and a `user-select` CSS rule slipped through on the first pass), not gaps in the sweep. Worth saying so rather than silently fixing it.

Verified the two structural claims that make the sweep's conclusion actually hold, not just its arithmetic:
- `dashboard.html`'s different provenance: `core.py`'s `index()` has no `require_auth_page` decorator; `user` comes from a direct `AuthService.get_user_data(api)` call whose failure bails to `clear_session_and_show_login(...)`, not a continue-with-None. Its ten guarded references were never protecting against this failure mode to begin with — confirmed, not just asserted.
- `workshop_hub` (`/workshop`, no playlist selected): passes `playlist=None` and no `user` kwarg at all. Line 1196 sits inside `{% if playlist %}` opened immediately above it — confirmed by reading the template directly. Accurately disclosed as a real second path, currently closed only by that conditional, not overclaimed as impossible.

**3. The distinction that has to survive — already explicit, and I checked it against the code rather than the prose.** The CHANGELOG bullet states it directly: `user` (the DB row, disambiguated by name from the template's `user`=spotify_profile — worth noting since both are informally "user" and that's exactly the collision that could confuse someone) can never be `None` past the decorator because every failure path redirects first; `spotify_profile` can be `None` by design because only its fetch is caught-and-continued. Read `routes/__init__.py:205-241` directly: three redirect paths (`require_auth()` failure, DB unavailable, missing db user) all return before `spotify_profile` is ever attempted; only that one call's `SpotifyError` is swallowed to `None`. Matches exactly. #550's removed `user` guards and this PR's added `spotify_profile` guard are both correct, for stated, verified, opposite reasons — this shouldn't confuse someone in six months.

**4. Merge sequencing (flagging, not fixing).** #550 and #551 share exactly one file, `CHANGELOG.md`, and nothing else — confirmed via both file lists. Whichever Chris merges second will need a CHANGELOG rebase, not a real conflict. Chris's call on order.

No merge from me.

Issue #539, opened 2026-08-02
Playlist covers fail to load on the Tiles page — 3 of 6 broken, likely a CSP img-src host gap (mosaic.scdn.co)


chrisrogers37:
Playlist cover art fails to load on the authenticated **Tiles** page. Reported by the owner with a screenshot.

![Tiles page with 3 of 6 playlist covers broken](https://github.com/chrisrogers37/shuffify/blob/screenshots/2026-08-02-broken-playlist-covers/documentation/screenshots/broken-covers/tiles-broken-covers-2026-08-02.jpg?raw=1)

**3 of 6 visible tiles are broken**, showing alt text where the cover should be:

| Playlist | Cover |
|---|---|
| THIS & A SPLIFF | **broken** — alt text |
| Long Live The Boys | **broken** — alt text |
| SUBARU SHIT | renders |
| 40 OUNCES | renders |
| POOL PARTY | **broken** — alt text |
| THE GOATS | renders |

This is the **first screen a logged-in user sees**. It is not the landing page, so it is **not** #537 — that was blank CSS gradients on the anonymous marketing page, blocked `style-src`. Here real `<img>` elements render and their `src` fails to load. Different surface, different mechanism, and this one is behind login where our monitoring cannot currently see it.

## Leading hypothesis — a gap in the CSP `img-src` allowlist

`shuffify/__init__.py:592`:

```
img-src 'self' https://i.scdn.co https://*.spotifycdn.com data:;
```

That names `i.scdn.co` **specifically**, not `*.scdn.co`.

Spotify serves playlist covers from more than one host. A playlist with a **user-uploaded custom cover** comes from `i.scdn.co` — allowed. A playlist with **no custom cover** gets an auto-generated four-up mosaic, served from **`mosaic.scdn.co`** — **not matched by either entry**, and therefore blocked.

That predicts exactly the split in the screenshot: the three that render have distinctive single images (custom covers); the three that break would be mosaics. `grep -rni mosaic` across the repo returns nothing, so no code path accounts for that host today.

**This is a hypothesis, not a confirmed diagnosis.** It fits the evidence and the mechanism is real, but nobody has yet opened the console on that page and read the violation. Do that first — it takes one page load and will either confirm it verbatim or kill it.

## Verification, in order

1. Load `/` authenticated in a browser and read the console. A CSP refusal names the blocked URI and the violated directive. That is the whole answer.
[...]

PR #553, opened 2026-08-08
fix(csp): admit every Spotify cover host, and unbreak the legal pages

### LINKED ISSUE #522: [security][bug] 112 inline event handlers are dead under CSP — workshop controls do not fire
## Summary

Inline event handlers do not execute under the app's Content Security Policy. **112 of them across the templates — 78 in `workshop.html` alone — are dead**, and the controls they are attached to do nothing when clicked.

Surfaced by @virgil while assessing SR-026 (#458) and independently confirmed here in a real browser. Filing separately because it is a functional outage, not template debt: unlike SR-026 this is not a refactor-shaped problem, it is a "these buttons do not work" problem.

## Evidence

Measured against the real rendered page — real Flask app serving real `/static`, real server-side session, Chromium via Playwright:

```
elements with inline handlers in the rendered workshop page:  72
inline-handler probe:  {'found': True, 'fired': False, 'tag': 'BUTTON'}
CSP violations by directive:  1 x script-src
```

The probe attaches a trivial `onclick` to a real button that already carries one, clicks it, and checks whether it ran. It did not.

Chromium's own console message explains why the nonce does not save it:

> Note that hashes do not apply to event handlers, style attributes and javascript: navigations unless the `'unsafe-hashes'` keyword is present.

The policy is set in `shuffify/__init__.py` (`_register_security_headers`) and carries no `unsafe-inline` in `script-src` — deliberately, and it should stay that way. A nonce authorizes a `<script>` element; it cannot authorize an `onclick` attribute. Those are different things and only the first is nonce-able.

## Scope

Counted statically across all templates, whitespace-anchored so `controls=` and `bounce=` do not inflate the number:

| Handler | Count |
|---|---|
| `onclick=` | 97 |
| `onchange=` | 9 |
| `onkeydown=` | 3 |
| `onsubmit=` | 1 |
| `oninput=` | 1 |
| `onfocus=` | 1 |
| **Total** | **112** |

By template: `workshop.html` **78**, and 34 spread across `schedules.html` (12), `macros/cards.html` (10), `dashboard.html` (5), `partials/settings_sidebar.html` (4), `partials/navbar.html` (1), `macros/forms.html` (1), `errors/500.html` (1).

## What this is not

**The probe proves handlers do not fire; it does not prove all 112 are user-visible dead controls.** Some may be redundant alongside an `addEventListener` binding that works fine, and some may sit on elements that are never rendered. Sizing the real user impact needs a pass through the list — that triage is the first step here, not a rewrite.

Related: #482 was the same class of defect on the styling side (an un-nonced `<style>` injection CSP blocked, which is why Tailwind is vendored now). This is that story repeating on `script-src`.

## Suggested approach

1. **Triage first.** Walk the 112 and mark each: genuinely dead, or shadowed by a working `addEventListener`. That number is what determines urgency.
2. **Convert, do not exempt.** Replace inline handlers with delegated listeners inside the existing nonce-guarded script blocks. Adding `'unsafe-hashes'` or `'unsafe-inline'` to `script-src` would make them work and would undo the protection the policy exists for — #267 (stored XSS via track URIs) is the reason it is strict.
3. **Guard it.** Once converted, a test asserting no template contains an `on*=` attribute keeps them from coming back. Cheap, and this class has now recurred twice.

The `workshop.html` half will collide with SR-026's extraction work, so sequencing them together is probably right — but the triage in step 1 is worth doing now regardless, because it tells us whether this is "a few stale attributes" or "the workshop has been substantially non-functional".

## Context
- Found by: @virgil, during SR-026 assessment (#458)
- Confirmed by: branden, real-browser probe, on `main` at `59712ed`
- Related: #482 (same defect class, `style-src`), #267 (why the policy is strict), #458 SR-026

chrisrogers37:
Taken as one investigation, because #522 and #539 are both CSP and two independent reports of one class are usually the visible part. The framing was right: this is the **third and fourth** instance of one incomplete sweep, and looking for the root instead of the two symptoms found a live breakage nobody had reported.

## First — a correction to my own measurement

My initial probe reported **zero violations on all six routes**. It was wrong, and it is worth stating how, because the failure is invisible by construction.

`require_auth_page` redirects to the login page. Playwright follows redirects. The final response is `200`, so six authenticated routes reported healthy while the browser had rendered the **landing page six times**. Nothing errored.

The fix was a per-route **content fingerprint** — a route whose fingerprint is absent is reported `PROBE-FAILED`, never clean. On the first run with it, four of six failed immediately, and the two that "passed" had matched fingerprints (`Workshop`, `Schedule`) that also appear in the landing page nav. The real cause was that this app uses Flask-Session with a server-side store, so the session cookie holds a SID and the one I had minted carried a payload the app never reads.

Everything below is from the corrected harness. No result here comes from a probe that has not been shown able to fail.

## The policy as it actually ships

Read off real responses, not from the config that is supposed to produce it:

```
default-src 'self'; script-src 'self' https://cdn.jsdelivr.net 'nonce-<per-request>';
style-src 'self' 'nonce-<per-request>'; img-src 'self' https://i.scdn.co https://*.spotifycdn.com data:;
connect-src 'self'; font-src 'self'; object-src 'none'; frame-ancestors 'none';
base-uri 'self'; form-action 'self' https://accounts.spotify.com
```

One definition site (`shuffify/__init__.py`), applied by `after_request` to every response including static files — which is what puts `/terms` and `/privacy` in scope.

## Finding 1 — #539 confirmed, and it is not one host

The hypothesis (`mosaic.scdn.co`) was right and **incomplete**. Every host below was loaded in Chromium under the real policy. `i.scdn.co` is the control: it is explicitly listed, so a run that showed it refused would mean the harness was broken.

| Host | Result |
|---|---|
| `i.scdn.co` | allowed — **control** |
| `mosaic.scdn.co` | **REFUSED** |
| `thisis-images.scdn.co` | **REFUSED** |
| `daily-mix.scdn.co` | **REFUSED** |
| `newjams-images.scdn.co` | **REFUSED** |
| `lineup-images.scdn.co` | **REFUSED** |
| `charts-images.scdn.co` | **REFUSED** |
| `seeded-session-images.scdn.co` | **REFUSED** |
| `image-cdn-ak.spotifycdn.com` | allowed (existing wildcard) |
| `image-cdn-fa.spotifycdn.com` | allowed (existing wildcard) |
| `seed-mix-image.spotifycdn.com` | allowed (existing wildcard) |

**The defect is an asymmetry, not an omission.** The policy names one `scdn.co` host specifically while wildcarding the entire sibling CDN. Adding `mosaic` and stopping would have left six known hosts broken and the next one to be found by a user, which is why this uses `https://*.scdn.co` despite the issue's preference for an explicit list — the set is open-ended editorial and generated-art hosts, and it is now handled exactly as `*.spotifycdn.com` already was.

Reproduced on the real Tiles page with three mosaic covers and three custom covers: **exactly three refusals**, matching the reporter's 3-of-6 screenshot.

**Stated bound:** this environment has no egress to Spotify, so no cover loads here and I cannot prove an allowed host *renders*. CSP refusal happens before the request, so the violation event is unaffected — but the evidence is "the policy no longer refuses these", not "the images appear."

## Finding 2 — #522 is already fixed and should be closed

Not fixed by this PR. Fixed before it, and never closed. Three independent instruments agree:

| Instrument | Result |
|---|---|
| Live DOM, all six routes | **0** elements carrying any `on*` attribute (the issue measured 72 on workshop alone) |
| Static scan of all templates | 83 `on*=` matches, **all 83** inside `// was onclick="..."` breadcrumbs; **0** live attributes |
| The repo's own `test_template_csp_contract.py` | passes |

`#528` migrated the last 30 handlers on 2026-08-02, the day after #522 was filed on 2026-08-01. The issue's counts (112 total / 78 in workshop) never matched the tree after that.

Probe validity, on **every** route: a freshly attached inline handler did not fire, and an `addEventListener` control on the same node did. A probe that cannot distinguish those two is not evidence of anything.

## Finding 3 — new, unreported, user-facing: `/terms` and `/privacy` render unstyled

Both files load `https://cdn.tailwindcss.com` and an inline `tailwind.config` script. `script-src` lists neither that host nor `unsafe-inline`, so **both are refused**:

```
script-src-elem  https://cdn.tailwindcss.com/
script-src-elem  inline (line 8)
```

Measured: `document.styleSheets.length === 0`, `typeof window.tailwind === 'undefined'`. The pages render as default serif on white. These are the legal pages linked from the login screen, which Spotify OAuth requires for consent.

| `/terms` on `main` | `/terms` on this branch |
|---|---|
| Terms of Service rendered as unstyled serif text on white | Terms of Service rendered with the Spotify gradient and glass card |

Re-measured against `origin/main` when capturing the before image, as an independent confirmation of the defect on the base rather than from memory: 2 violations, 0 stylesheets on each page.

Same root cause as #499 and #537, one directory further out. Both those sweeps covered `templates/`; these two files live in `static/public/` because they are sent verbatim by `send_from_directory` rather than rendered. `test_csp_script_src_does_not_allow_tailwind_cdn` (SR-042) already asserts that host must **not** be permitted — so the policy is right and the pages were left behind.

**The obvious fix would have changed nothing.** `tailwind.config.js` scanned `templates/**/*.html` and `static/**/*.js`, so these pages' utilities were never compiled into the committed stylesheet. Linking it without widening the glob produces a page that still renders unstyled, with a diff that looks correct. The glob is widened and the committed CSS rebuilt.

## Finding 4 — why the sweep keeps missing things

The guard's scope was narrower than the header's scope. `test_template_csp_contract.py` scanned `shuffify/templates/` only, so `static/public/` was invisible to it.

**Widening the scope alone would still not have caught Finding 3.** Those pages carry no `style` attribute and no `on*` handler — the two rules the guard had. Their defect is a third class the guard could not express. Both gaps are closed:

- scope now includes `static/public/`
- new rule: every `<script src>` host must be permitted by the real `script-src`, read from the shipped policy (and it asserts the directive was found, so it cannot silently permit everything)
- new rule: a static page may not contain an inline `<script>` at all — a template can carry `nonce="{{ csp_nonce }}"`, a file that is never rendered cannot

## Verification

Every gate below was run; none is claimed.

**Probe self-test.** Deliberate violations of each directive under audit, to prove the collector sees them: caught `img-src`, `script-src-elem`, `style-src-attr`. The unnonced inline script did **not** execute and the `style` attribute did **not** apply, so CSP is genuinely enforcing rather than the page merely looking plausible. Additionally, each route carries its own live proof — the inline-handler probe fires a real `script-src-attr` violation on every page, so a page reporting zero app violations has demonstrably-working collection on that page.

**Can this crawl actually catch a violation? Tested per directive, not once.**

A clean sweep is only a bound if the instrument has been shown to fail. Each directive below was
deliberately violated on a probe route served by the real app under the real header, and the
collector's response recorded:

| Directive in the policy | Deliberate violation | Caught |
|---|---|---|
| `script-src` | unnonced inline `<script>`; inline `on*` handler | yes — `script-src-elem`, `script-src-attr` |
| `style-src` | `style` attribute | yes — `style-src-attr` |
| `img-src` | image from a non-allowlisted host | yes |
| `connect-src` | `fetch()` to an external origin | yes |
| `font-src` | external font preload | yes |
| `object-src` | `<object data="https://...">` | yes |
| `base-uri` | `<base href="https://...">` | yes |
| `default-src` | `<video src="https://...">` | yes — as `media-src`, via fallback |
| `form-action` | cross-origin form submit | **NO — not proven** |
| `frame-ancestors` | — | **NO — not testable this way** |

Two honest gaps. `form-action` is a probe-mechanics failure: submitting the form hangs the navigation
against a host that does not resolve, so the page never reaches a state where the listener can be
read — I could not make it fire, which is different from knowing it would not. `frame-ancestors` is
enforced on the embedding page, so a document cannot trip it against itself.

So the sweep is bounded for **8 of the 10 directives**, and unbounded for those two. Beyond the
directive matrix, two further checks confirm enforcement rather than mere detection: the unnonced
script did **not** execute and the `style` attribute did **not** apply. And every route in the sweep
carries its own live proof — the inline-handler probe fires a real `script-src-attr` violation on
each page, so "0 app violations" on a given route is reported alongside a collector demonstrably
working on that same page load.

**Verifying Finding 2 independently.** #522 is a security issue about controls that do not fire, so
it should be confirmed rather than taken on report. The cheapest check needs no dependencies and no
browser — it classifies every `on*=` match in the templates as a live attribute or a `//` comment:

```bash
python3 - <<'EOF'
import re, pathlib
RAW = re.compile(r'''\son[a-z]+\s*=\s*["'\\]''', re.I)
live = comments = 0
for p in pathlib.Path("shuffify/templates").rglob("*.html"):
    t = p.read_text()
    for m in RAW.finditer(t):
        ls = t.rfind("\n", 0, m.start()) + 1
        if re.search(r"(?<!:)//", t[ls:m.start()]): comments += 1
        else: live += 1
print(f"live on*= attributes: {live}   inside // comments: {comments}")
EOF
# expected on this branch: live 0, comments 83
```

`python3 -m pytest tests/test_template_csp_contract.py -q` is the second check, and the browser DOM
count is the third. The first two are the ones worth re-running; the third needs the harness.

**Note on the harness.** The browser crawl is not committed, because `playwright` is not a declared
dependency and adding one is a decision for this repo rather than a side effect of a bugfix. If a
standing real-browser CSP check is wanted, that is worth its own issue — the directive matrix above
is most of the work.

**Guard rules, mutation-checked against the defect each exists for** rather than confirmed on a clean tree:

| Mutation | Test that failed |
|---|---|
| restore the CDN `<script src>` | `test_external_scripts_are_permitted_by_script_src` |
| restore the inline config script | `test_static_public_pages_have_no_inline_script` |
| add a `style` attribute to a `static/public` page | `test_no_style_attribute[terms.html]` — proves the wider scope |
| add an `on*` handler to a `static/public` page | `test_no_inline_event_handler[terms.html]` — proves the wider scope |

Each mutation fails exactly one test. Reverting `img-src` to the old policy fails `test_csp_img_src_allows_every_spotify_cover_host` and passes with the fix.

The replaced `img-src` test asserted `"https://i.scdn.co" in csp` and passed throughout this bug's lifetime — a pattern match on the policy text cannot distinguish a host being *listed* from a host being *admitted*. The new one resolves each host against the directive's sources, and ships with a control asserting it rejects `evil.example.com` and `scdn.co.evil.example.com`, because a matcher that only ever returns true would satisfy the positive test too.

**After the fix, in the browser:** 0 app violations on all six authenticated routes (was 3 on `/`), and 0 on `/terms` and `/privacy` (was 2 each), with `styleSheets.length` 0 → 1.

**Suite, before/after against a clean `origin/main` worktree:** `rc=0` **2181 passed** → `rc=0` **2205 passed**, 0 failed both sides. No failures introduced; the delta is the 24 added cases.

`ruff check .` — All checks passed. `./scripts/build-css.sh --check` — committed stylesheet matches sources (run because this edits templates and the compiled CSS is committed).

## Coverage bounds

- **Six authenticated routes plus the two legal pages, at page load.** Interactions were not exercised, so a violation that only occurs after a click — content built into `innerHTML`, for instance — would not appear here. The static scan covers that class for templates and `static/js`; the browser sweep does not.
- **The three error templates** were not loaded in the browser. They were scanned statically and are clean (`errors/500.html:51` is a breadcrumb comment).
- **No egress to Spotify**, per the bound stated under Finding 1.
- **`connect-src`, `font-src`, `form-action`, `base-uri`, `frame-ancestors`, `object-src`** produced no violations on the routes measured, and the tree references no external font or fetch host. That is a negative from eight page loads, not a proof.

## Recommendation

Merge this for #539 and Finding 3. **Close #522 as already fixed** — I have not touched it here, and the evidence is above.

https://claude.ai/code/session_01VDF644wejEcjXruxVehzoH


chrisrogers37:

**#522 — verified independently against current `main` (`437882d`), not accepted on the PR's own evidence.** Two checks, both a different method than the PR's own live-browser probe:

1. Static scan of every template for a live `on(click|load|error|change|submit|mouseover|mouseout|focus|blur|input|keyup|keydown|keypress|mouseenter|mouseleave|dblclick|drag|drop)="..."` attribute (word-boundary anchored — my first pass used an unanchored pattern and matched `nonce=`/`aria-controls=`/`data-action-click=` as false positives, caught before reporting it). Result: **83 matches, all 83 inside `// was on...=` breadcrumb comments, zero live.**
2. Cross-referenced every `data-action-*="key"` used against every `registerActions({...})` registration, across all templates **and** `shuffify/static/js/actions.js` (missed the JS file on the first pass, which produced 4 false "orphans" — the four shared-chrome keys registered there; re-ran including it). Result: **106 keys used, 106 registered, zero orphans.**

`actions.js` itself is a real, non-stub delegation module — `closest()`-based so dynamically-injected controls need no rebinding, correct capture-phase handling for `focus`/`blur` (they don't bubble), and it fails loud (console.error) on a duplicate registration rather than silently dropping a handler. #522 should close.

**The CSP change admits exactly what it needs to, nothing more.** `img-src` widens `i.scdn.co` to `*.scdn.co` — a proportionate fix for an *image* directive that now matches the pattern `*.spotifycdn.com` already used for the sibling CDN, not a new class of grant. Finding 3's fix removes the blocked CDN script + inline config entirely (`static/public/privacy.html` diff) rather than loosening `script-src` to admit it; confirmed the custom `spotify-green`/`spotify-dark` colors the removed inline config defined already exist in the committed `tailwind.config.js`, and `static/public/**/*.html` is now in its `content` glob so they actually compile into the shipped CSS (grepped the compiled `tailwind.css` on this branch — present).

**Mutation-tested a sample of the new tests myself, with this week's standing requirement — proved each mutant changed real behavior before trusting the result:**

| Test | Mutation | No-op check | Result |
|---|---|---|---|
| `test_csp_img_src_allows_every_spotify_cover_host` | reverted `img-src` to `i.scdn.co` only | queried the real served `/health` CSP header before and after — wildcard genuinely absent post-mutation | FAILED, naming exactly the 7 hosts |
| `test_external_scripts_are_permitted_by_script_src[privacy.html]` | added back the CDN `<script src>` | grepped the mutated file for the string — present | FAILED |
| `test_static_public_pages_have_no_inline_script[privacy.html]` | added back an unnonced inline `<script>` | grepped for the tag — present | FAILED |

All three restored, byte-identical diffs against pristine after each. **Full suite, run myself:** `tests/test_app_factory.py tests/test_template_csp_contract.py` → 99 passed.

No merge from me.

